PR #24, opened 2023-11-03
Add reconciler default timout, cluster and machine scopes, machine `Reconcile` function, and `InstanceStatus` enums

zliang-akamai:
Some basic components


cloud/scope/cluster.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// ClusterScope defines the basic context for an actuator to operate upon.
+type ClusterScope struct {
+	client      client.Client
+	patchHelper *patch.Helper
+
+	LinodeClients
+	Cluster    *clusterv1.Cluster
+	GCPCluster *infrav1.LinodeCluster
```
luthermonson, line 50: gcp cluster eh?

cloud/scope/clients.go
```diff
@@ -0,0 +1,25 @@
...
+limitations under the License.
+*/
+
+package scope
+
+import (
+	"github.com/linode/linodego"
+)
+
+type LinodeClients struct {
```
luthermonson, line 23: i might be missing context on other usages of the linodego client... is this plural on purpose? At present I don't see the benefit over doing it this way than just adding a `lindoeClient` next to the k8s client on the on the scope.

zliang-akamai: I stole the pattern from the achieved provider and other provider and but later realized we just have one client. Thanks for the comment!

zliang-akamai: After digging into the cluster API provider implementation, now I think the `LinodeClients` struct is still useful for centralizing Linode client lifecycle management logic, e.g. client creation. 

[This was found in the achieved version of the CAPI provider.](https://github.com/linode/cluster-api-provider-linode-archive/blob/master/cloud/scope/sessions.go#L30-L45) But yea, it is not that necessary and we should be fine for now with the a private function to create the client in cluster scope package. We can always move to this pattern later when we find it's more useful/necessary (unless you think we should do it now, lol).

luthermonson (approved):
great!

Issue #25, opened 2023-11-06
Implement business logic to create Linode instances


mhmxs:
Soon or later we would need some business logic to create Linode instances. Our plan is to use https://github.com/linode/linodego SDK under the hood.

Acceptance criteria:
- Helper function for instance creation
- Wait for ready solution
- Unit tests

mhmxs:
This issue solves the other end of  https://github.com/linode/cluster-api-provider-linode/issues/11

mhmxs:
WIP

PR #31, opened 2023-11-15
Implement helper for Linode instance creation

mhmxs:
This changes introduces Linode Go SDK helper functions.

Fixes: https://github.com/linode/cluster-api-provider-linode/issues/25


mhmxs:
@zliang-akamai FYI

zliang-akamai:
I think [we intentionally removed `internal` directory before](https://github.com/linode/cluster-api-provider-linode/pull/15). Would you integrate your implementation into the code outside of `internal`?

mhmxs:
> I think [we intentionally removed `internal` directory before](https://github.com/linode/cluster-api-provider-linode/pull/15). Would you integrate your implementation into the code outside of `internal`?

Done @zliang-akamai

cmd/main.go
```diff
@@ -64,7 +71,16 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
-	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
+	zapLogger := zap.New(zap.UseFlagOptions(&opts))
+
+	ctrl.SetLogger(zapLogger)
+	klog.SetLogger(zapLogger)
+
+	linodeToken, ok := os.LookupEnv("LINODE_API_TOKEN")
```
lgarber-akamai, line 79: optional: The standard environment variable we use for DX-owned integrations is `LINODE_TOKEN`, so it might make sense to also use that here. It might also make sense to move the `LINODE_API_TOKEN` string into a const.

cmd/main.go
```diff
@@ -53,8 +58,10 @@ func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
+	var instanceCreateTimeout time.Duration
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.DurationVar(&instanceCreateTimeout, "instance-create-timeout", time.Minute, "The timeout of Linode instance creation.")
```
avestuk, line 64:
```suggestion
	flag.DurationVar(&instanceCreateTimeout, "instance-create-timeout", time.Minute, "The timeout for Linode instance creation.")
```

util/gosdk/instance.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type instanceHandler struct {
+	client        *Client
+	createTimeout int
+	log           logr.Logger
+}
+
+// CreateInstance creates an instance and wits for booting.
+func (ih *instanceHandler) CreateInstance(ctx context.Context, insanceOptions linodego.InstanceCreateOptions) (*linodego.Instance, error) {
```
avestuk, line 33:
```suggestion
func (ih *instanceHandler) CreateInstance(ctx context.Context, instanceOptions linodego.InstanceCreateOptions) (*linodego.Instance, error) {
```

util/gosdk/instance.go
```diff
@@ -0,0 +1,49 @@
...
+	instance, err := ih.client.CreateInstance(ctx, insanceOptions)
```
avestuk, line 34:
```suggestion
	instance, err := ih.client.CreateInstance(ctx, instanceOptions)
```

avestuk (commented):
Caught a couple typos. Also hello again @mhmxs :D

cmd/main.go
luthermonson: +1 for switching this to `LINODE_TOKEN`

avestuk (approved):
LGTM

zliang-akamai:
I think we can cleanup this older client creation function and replace it with the newer one.
https://github.com/linode/cluster-api-provider-linode/blob/main/cloud/scope/cluster.go#L51-L65

util/gosdk/instance.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	ih.log.V(3).Info("instance created", "ID", instance.ID)
+
+	instance, err = ih.client.WaitForInstanceStatus(ctx, instance.ID, linodego.InstanceRunning, ih.createTimeout)
+	if err != nil {
+		return nil, logging.LogAndWrapError(ih.log, "wait for instance creation failed", err, instanceOptions)
+	}
+
+	ih.log.V(3).Info("instance running", "ID", instance.ID)
```
zliang-akamai, line 46: Should we not to block the overall process by waiting for the status of an instance?

mhmxs: @zliang-akamai The reason for the block is machine controller has to know when an instance is ready to deploy Kubernetes. So i think at some point we have to know the machine status to update custom resource status. May i ask what is in your head about status updates?

luthermonson: i dont think we should be blocking anything in a reconcile loop, should be setting it to create and requeue the machine to run again in 5s and check the status until it's ready. No point creating our own blocking methods when controller framework was designed to take care of that for us

zliang-akamai:
By the way, I think we might be able to leave all logic that makes calls to `linodego` package within `cloud/scope`? And we had a previous discussion in [another PR](https://github.com/linode/cluster-api-provider-linode/pull/24#discussion_r1382521318) that a simpler `linodego` client in the cluster scope might be enough for we have now.. What do you think?

mhmxs:
Change requests would be targeted in a separate PR.

util/gosdk/instance.go
mhmxs: @luthermonson My plan was also to not block reconciliation loop, I just experienced higher CPU load during re-queuing items, than retrying or waiting things in a separated goroutine. Of course we pay with memory on the other hand. We have fixed number of reconciliation loops, so if we have lot's of re-queues, we should increase reaction time. On the other hand we can create millions of sleeping goroutines. WDYT?

Issue #11, opened 2023-09-29
Define API for LinodeMachine


luthermonson:
We need to pull from other CAPI Linode providers and define our new LinodeMachine

* What spec and status do we feel will give us what we want?

zliang-akamai:
May I take this?

dennisme:
yep @zliang-akamai go for it!

PR #36, opened 2023-12-01
Create and delete machines via controller

mhmxs:
I open this PR for transparency. I'm still testing, but i think it is ready to discuss :).

This change doing reconciliation loop for Linode Machines. Only create and delete has been implemented. You can find some buxfixes and code cleanups too, i tried to push the scope as low as possible.

Fixes: https://github.com/linode/cluster-api-provider-linode/issues/25
Fixes: https://github.com/linode/cluster-api-provider-linode/issues/11


controller/linodemachine_controller.go
```diff
@@ -117,17 +137,151 @@ func (r *LinodeMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	// Delete
+	if !linodeMachine.ObjectMeta.DeletionTimestamp.IsZero() {
+		logger.Info("deleting machine")
+
+		if machineScope.LinodeMachine.GetInstanceID() != 0 {
+			if err = machineScope.LinodeClient.DeleteInstance(ctx, machineScope.LinodeMachine.GetInstanceID()); err != nil {
+				logger.Info("Failed to delete Linode machine instance", "error", err.Error())
+
+				// Not found is not an error
+				if apiErr, ok := err.(*linodego.Error); ok && apiErr.Code != http.StatusNotFound {
```
mhmxs, line 175: I'm not sure this is the best way to validate linode api errors

controller/linodemachine_controller.go
```diff
@@ -56,28 +66,33 @@ func (r *LinodeMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 	}
 
 	machine, err := util.GetOwnerMachine(ctx, r.Client, linodeMachine.ObjectMeta)
 	if err != nil {
-		return ctrl.Result{}, err
-	}
-	if machine == nil {
+		log.Info("Failed to fetch owner machine", "error", err.Error())
+
+		return ctrl.Result{}, client.IgnoreNotFound(err)
```
mhmxs, line 82: @luthermonson Does it make sense to requeu the object if owner machine is missing? Do we have a chance some other controller creates it in the background? Sorry I'm not sure.

controller/linodemachine_controller.go
```diff
@@ -56,28 +66,33 @@ func (r *LinodeMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	} else if machine == nil {
 		log.Info("Machine Controller has not yet set OwnerRef")
+
 		return ctrl.Result{}, nil
 	}
 
 	log = log.WithValues("Linode machine: ", machine.Name)
+
 	cluster, err := util.GetClusterFromMetadata(ctx, r.Client, machine.ObjectMeta)
```
mhmxs, line 91: @luthermonson @zliang-akamai It seems we trust in metadata information, but there is no guarantee to find the right cluster. I pretty understand it is an edge case, but not impossible. WDYT? Do we need some extra validation here?

api/v1alpha1/linodemachine_types.go
```diff
@@ -25,17 +26,59 @@ import (
 
 // LinodeMachineSpec defines the desired state of LinodeMachine
 type LinodeMachineSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	Region          string                                 `json:"region"`
+	Type            string                                 `json:"type"`
+	Label           string                                 `json:"label,omitempty"`
+	Group           string                                 `json:"group,omitempty"`
```
avestuk, line 32: What would this "Group" property mean for a Linode?

mhmxs: I don't know @avestuk , this part of the code is a copy of https://github.com/linode/linodego/blob/1c6ad41c9ec16f2d1e583da338dc00d407629b22/instances.go#L113

avestuk: Ah ok - The docs say it's deprecated so maybe drop it? 

https://www.linode.com/docs/api/linode-instances/#linode-create

mhmxs: @avestuk we should drop it, but we have a dynamic converter here: https://github.com/linode/cluster-api-provider-linode/pull/36/files#diff-8daab1045cecf480419b0a31c2f8558e51197d31a7be910386ef4e420db0f7a4R314 which in my best understanding is able to handle this problem, I just don't want to ensure :) Currently we have a simple test to verify integrity between the two objects: https://github.com/linode/cluster-api-provider-linode/pull/36/files#diff-a78247066969ae5bb22df9e10e6a41c5435d78a1ac24264c31a69df60ec8ff90R13 WDYT?

avestuk: Ah right its for the lindego client to create the machine. Yeah fair enough then leave it as is :D

controller/linodemachine_controller.go
luthermonson: if something updates the owner refs later this should eventually get tripped so probably okay to ignore for now right?

controller/linodemachine_controller.go
luthermonson: i agree with you that it's a super edge case... maybe skip extra validation for now

controller/linodemachine_controller.go
luthermonson: this is pretty typical for what ive seen

luthermonson:
> @luthermonson @zliang-akamai I'm a bit afraid of re-queuing every instance we wait for, maybe it makes sense to retry only X times and dropping the event instead of filling workqueue and waste reconciliation threads WDYT?:
> 
> ```
> if lastKnownState != linodego.InstanceRunning {
> 			logger.Info("Instance is not ready, wait for running...", "state", lastKnownState)
> 
> 			res = ctrl.Result{RequeueAfter: reconciler.DefaultMachineControllerWaitForRunningDelay}
> 		}
> ```

want to extend the back off as it gets older? if these dont end up running we could be in a way worse case for the data center so eventually stopping at 10 or 15min might be plent?

controller/linodemachine_controller.go
```diff
@@ -117,17 +160,148 @@ func (r *LinodeMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+
+			err = errors.Join(err, patchErr)
+		}
+	}()
+
+	// Delete
+	if !linodeMachine.ObjectMeta.DeletionTimestamp.IsZero() {
+		logger.Info("deleting machine")
+
+		if machineScope.LinodeMachine.GetInstanceID() != 0 {
```
luthermonson, line 193:
```suggestion
		if machineScope.LinodeMachine.GetInstanceID() >= 0 {
```

luthermonson: maybe just for safety if we get a negative number some how, if you agree to this change then swap out the others you do the same check on

controller/linodemachine_controller.go
```diff
@@ -117,17 +160,148 @@ func (r *LinodeMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	// Create
+	tags := []string{string(machineScope.LinodeCluster.UID), string(machineScope.LinodeMachine.UID)}
+	filter := map[string]string{
+		"tags": strings.Join(tags, ","),
+	}
+
+	rawFilter, err := json.Marshal(filter)
+	if err != nil {
+		// This should never happen
+		panic(err.Error() + " Oh, snap... Earth has over, we can't parse map[string]string to JSON! I'm going to die ...")
```
luthermonson, line 252: 🤣

api/v1alpha1/linodemachine_types.go
```diff
@@ -25,17 +26,59 @@ import (
...
+	Primary     bool                            `json:"primary,omitempty"`
+	SubnetID    *int                            `json:"subnetId,omitempty"`
+	IPv4        *VPCIPv4                        `json:"ipv4,omitempty"`
+	IPRanges    []string                        `json:"ipRanges,omitempty"`
+}
 
-	// Foo is an example field of LinodeMachine. Edit linodemachine_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+// VPCIPv4 defines VPC IPV4 settings
+type VPCIPv4 struct {
```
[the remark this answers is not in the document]
mhmxs: @luthermonson the cfields of the object are coming from: https://github.com/linode/linodego/blob/f93d6ba73bc9ff6a79d398ee51f585a45a96b350/instance_config_interfaces.go#L34 We support what linodego supports, so adding support is out of the scope of the provider.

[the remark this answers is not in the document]
luthermonson: yup, totally agree

luthermonson (approved):

mhmxs:
@luthermonson It is ready to merge. I couldn't test the change, because of the missing other parts of the project, but verified the behavior based on other well known provider.

controller/linodemachine_controller.go
```diff
@@ -117,22 +181,220 @@ func (r *LinodeMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		if patchErr := machineScope.PatchHelper.Patch(ctx, machineScope.LinodeMachine); patchErr != nil &&
+			client.IgnoreNotFound(patchErr) != nil {
+
+			logger.Error(patchErr, "failed to patch LinodeMachine")
+
+			err = errors.Join(err, patchErr)
+		}
+	}()
+
+	// Delete
```
avestuk, line 223: What do you think about breaking the delete/update/create parts of this function out into their own functions? I don't mind if we do it later in another PR (I'm happy to take a crack at it for instance) but I think the function would be more easily understood if we did so.

cloud/scope/cluster.go
```diff
@@ -34,25 +33,20 @@ import (
...
 }
 
 func validateClusterScopeParams(params ClusterScopeParams) error {
 	if params.Cluster == nil {
-		return fmt.Errorf("Cluster is required when creating a ClusterScope")
+		return errors.New("cluster is required when creating a ClusterScope")
 	}
 	if params.LinodeCluster == nil {
-		return fmt.Errorf("LinodeCluster is required when creating a ClusterScope")
+		return fmt.Errorf("linodeCluster is required when creating a ClusterScope")
```
avestuk, line 44:
```suggestion
		return errors.New("linodeCluster is required when creating a ClusterScope")
```

avestuk (approved):
LGTM as is. Comments are nits/prompts for futher discussion we can have after merge if that's desirable.

controller/linodemachine_controller.go
```diff
@@ -43,41 +85,64 @@ type LinodeMachineReconciler struct {
...
+	case machine == nil:
+		log.Info("Machine Controller has not yet set OwnerRef, skipping reconciliation")
+
+		return ctrl.Result{}, nil
+	case skippedMachinePhases[machine.Status.Phase]:
+		log.Info("Machine phase is not the one we are looking for, skipping reconciliation", "phase", machine.Status.Phase)
+
 		return ctrl.Result{}, nil
+	default:
+		match := false
```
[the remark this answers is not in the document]
mhmxs: @avestuk in my personal opinion is `slices` package breaks one of the principles of Golang: simplicity. I would be the last who suggest adopting it :D But of course the question is what we think about his teamwise.

luthermonson (approved):

luthermonson:
good to see some progress! well done @mhmxs and thanks for the help @avestuk

Issue #44, opened 2023-12-18
Static code analysis during CI build


mhmxs:
Would be nice to introduce static code analysis into CI pipeline.

PR #50, opened 2024-01-02
Introduce static code analysis

mhmxs:
Golangci-lint, Gosec, Govulncheck and Nilaway has been introduced. CodeQL Github also has been added.

Fixes: https://github.com/linode/cluster-api-provider-linode/issues/44

### DIFF
--- a/.github/workflows/build_test_ci.yml
+++ b/.github/workflows/build_test_ci.yml
@@ -6,11 +6,31 @@ on:
   pull_request:
     branches: [ "main" ]
   
-jobs:
+permissions:
+  contents: read
+  pull-requests: read
+  actions: read
 
+concurrency:
+  group: build-test-ci-${{ github.ref }}-1
+  cancel-in-progress: true
+
+jobs:
   go-build-test:
     runs-on: ubuntu-latest
     steps:
+    - name: Harden Runner
+      uses: step-security/harden-runner@v2
+      with:
+        disable-sudo: true
+        egress-policy: block
+        allowed-endpoints: >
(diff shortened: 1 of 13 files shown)


mhmxs:
@luthermonson @avestuk @zliang-akamai Could you please take a look

avestuk (commented):
I'm fine with the SCA change. I'm glad to see the updates to break out reconcile into smaller pieces!

My one question is why you've chosen to use named returns in all of the the new reconcile functions such as reconcileUpdate. Personally I'm just not used to seeing naked returns in Go and I find it a bit jarring.

mhmxs:
> why you've chosen to use named returns in all of the the new reconcile functions

@avestuk the answer is simply because the original `Reconcilie` function is full of defer statements, and i didn't change the code style during refactoring.

avestuk:
> @avestuk the answer is simply because the original Reconcilie function is full of defer statements, and i didn't change the code style during refactoring.

Eminently reasonable! What do you think about only doing named returns in `Reconcile/reconcile` functions where we have got the defer statements? I'm thinking that it'd be preferable to keep the named return pieces limited to places where we need them in order to modify the return values. That way it's clear why it's being done and we don't have it across the whole code base?

mhmxs:
> That way it's clear why it's being done and we don't have it across the whole code base?

@avestuk I did what i could, but reconcile update has to be an exception too.

avestuk:
That works for me!

luthermonson (approved):

shanduur (approved):

PR #53, opened 2024-01-11
Support linux for kind clusters s/x86_64/amd64

srust:


luthermonson (approved):

Makefile
```diff
@@ -203,7 +203,11 @@ $(TILT): $(LOCALBIN)
 kind: $(KIND) ## Download kind locally if necessary. If wrong version is installed, it will be overwritten.
 $(KIND): $(LOCALBIN)
 	test -s $(KIND) && $(KIND) version | grep -q $(KIND_VERSION) || \
-	(cd $(LOCALBIN); curl -Lso ./kind https://kind.sigs.k8s.io/dl/v$(KIND_VERSION)/kind-$(OS)-$(ARCH) && chmod +x kind)
+	(KIND_ARCH=$(ARCH); \
```
luthermonson, line 206: maybe better to change ARCH to GOARCH?

srust: you mean do something like:

```
KIND_ARCH=$(shell go env GOARCH)
```
?

luthermonson: ya, that

eljohnson92:
This is now handled as part of #65

Issue #10, opened 2023-09-29
Define API for LinodeCluster


luthermonson:
We need to pull from other CAPI Linode providers and define our new LinodeCluster

* What spec and status do we feel will give us what we want?

dennisme:
Ill take this one!

PR #57, opened 2024-01-18
add spec for LinodeCluster

eljohnson92:
Add a minimal spec for the LinodeCluster resource, does not include any VPC related config as that is still being discussed in a separate issue.
fixes: #10


dennisme (approved):
Looks awesome!

luthermonson (approved):

Issue #4, opened 2023-09-28
VPC Support


luthermonson:
As we get closer to VPCs becoming beta we should work with how we could implement them.

mhmxs:
@luthermonson I did a short research, and we are able to manage vpc-s and subnets. And on the UI there is an option to add instance to the vpc, so we just change machine creation (or golangsdk if it is missing) to set VPC. The big question is i think do we want to support both vpc and classic instances? Most of the capi providers are shipping vpcs by default, without any option to change. WDYT?

luthermonson:
We should support both if for some reason VPC will not be available in all data centers. Does it make sense that we should default to VPC and error or fall back to classic if they're not available in the datacenter?

mhmxs:
> We should support both if for some reason VPC will not be available in all data centers. Does it make sense that we should default to VPC and error or fall back to classic if they're not available in the datacenter?

@luthermonson we should. I'm not a big fan of automatic fall back to potentially lower security level. But we can create a validation webhook, to validate VPC availability. WDYT?

mhmxs:
@luthermonson I think we have to think about how to bootstrap Kubernetes cluster on an isolated environment.
- The most common way to solve this problem is a bastion node (per datacenter or per cluster, private ip or vlan).
- The second option is to opening a backdoor for cluster management. We should achieve high security (2 way ssl, inclusive firewall, etc), but i don't recommend it.
- The third option is to use stackscripts and a small agent on hosts for any kind of code execution. Capi provider creates a stackscript, which the agent polling and executing.
- Forth option came into my mind is totally immutable infrastructure, so Metadata service is the source of truth and we replace nodes on any case of change.

What is your vision?

luthermonson:
@mhmxs i like the 4th option, the metadata should be the source of truth. ive seen option bastion hosts too but only if its option in case whomever is using this rolls their own. i dont like opening backdoors or using a small agent. do you prefer 4?

mhmxs:
> do you prefer 4?

:+1:

PR #62, opened 2024-01-24
Introduce VPC controller

mhmxs:
This change introduces a new controller to manage VPCs on Linode. The Cluster object has an optional `vpcRef`erence field. The customer should bring them own VPC by setting the `label` field to the custom resource to pointing to an existing VPC. The machine controller also has been updated to attach new instances to the VPC.

Fixes: https://github.com/linode/cluster-api-provider-linode/issues/4


controller/linodevpc_controller.go
```diff
@@ -0,0 +1,226 @@
...
+	vpcScope.LinodeVPC.Status.Ready = true
+
+	return nil
+}
+
+func (r *LinodeVPCReconciler) reconcileDelete(ctx context.Context, logger logr.Logger, vpcScope *scope.VPCScope) error {
+	logger.Info("deleting VPC")
+
+	if vpcScope.LinodeVPC.Spec.VPCID != nil {
+		if err := vpcScope.LinodeClient.DeleteVPC(ctx, *vpcScope.LinodeVPC.Spec.VPCID); err != nil {
```
luthermonson, line 192: think we need to do any validation the vpc is empty or anything before the delete?

util/helpers.go
```diff
@@ -1,6 +1,40 @@
...
+	}
+
+	if len(tags) != 0 {
+		filter["tags"] = strings.Join(tags, ",")
+	}
+
+	rawFilter, err := json.Marshal(filter)
+	if err != nil {
+		// This should never happen
+		panic(err.Error() + " Oh, snap... Earth has over, we can't parse map[string]string to JSON! I'm going to die ...")
```
luthermonson, line 36: i get this won't happen but a panic in a util scares me...  can we just log and return an empty filter?

mhmxs: An empty filter would break correctness of list operations of the API. So we have to return the error (which never happens) and deal with it everywhere. WDYT?

luthermonson: if you call `CreateLinodeAPIFilter("", nil))` you'll get `{}`, we can't return that as a base case without making things worse in the subsequent calls?

mhmxs: > if you call `CreateLinodeAPIFilter("", nil))` you'll get `{}`, we can't return that as a base case without making things worse in the subsequent calls?

@luthermonson I'm worries about this list call for example. https://github.com/linode/cluster-api-provider-linode/pull/62/files#diff-8daab1045cecf480419b0a31c2f8558e51197d31a7be910386ef4e420db0f7a4R253 If `CreateLinodeAPIFilter` returns an empty filter, this call would list all of the instances, and we and in this error: https://github.com/linode/cluster-api-provider-linode/pull/62/files#diff-8daab1045cecf480419b0a31c2f8558e51197d31a7be910386ef4e420db0f7a4R303. 

This call https://github.com/linode/cluster-api-provider-linode/pull/62/files would end in https://github.com/linode/cluster-api-provider-linode/pull/62/files#diff-371045646e070a553a5d00d8927bc43b4c0d67da0b77e054a6ddcf5123b838f4R52. We select the first item. 

So i'm glad to change the code, but we have to count with some refactoring and in the future we have to remember to this behavior all the time. Otherwise, we should iterate on the wrong set of API objects.

Anyway this is a common pattern in go, maybe i can rename the function to `CreateLinodeAPIFilterOrDie` or `MustCreateLinodeAPIFilter` to turn it super goish.

api/v1alpha1/linodecluster_types.go
```diff
@@ -35,6 +36,10 @@ type LinodeClusterSpec struct {
 	// NetworkSpec encapsulates all things related to Linode network.
 	// +optional
 	Network NetworkSpec `json:"network"`
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// +optional
+	VPCRef *corev1.ObjectReference `json:"vpcRef,omitempty"`
```
eljohnson92, line 42: Do we plan to support CAPI creating VPCs for your cluster if not provided? if so we might want to maybe add a VPC section or something like that to support some options that might be used other than an external VPC ref

mhmxs: @eljohnson92 I can imagine a VPCTemplate similar to Machine Template. But i think we should address it in a new PR. At the moment VPC API is not available, so the only option is bring your own vpc at the moment.

eljohnson92: It looks like there are client packages for VPCs [here](https://github.com/linode/linodego/blob/main/vpc.go) do we need to just use the beta API with the client in order to use it?

mhmxs: @eljohnson92 I'm not pretty sure. I couldn't create VPC via CLI or API on v4. Web UI uses v4beta as version, but this version is not available with my personal token. Documentation also mentions this story.

eljohnson92: @mhmxs in the linodego [client](https://github.com/linode/linodego/blob/main/client.go#L57) we can set apiVersion to `v4beta` to access the beta api. I would be fine tackling this in a later PR but I just want to make sure we account for those changes in this interface

mhmxs: @eljohnson92 Your points are valid, just should we continue discussion at the ticket? https://github.com/linode/cluster-api-provider-linode/issues/4

controller/linodevpc_controller.go
luthermonson: followup on this, i tested creating a vpc and deleting it with a node in it and it will error out telling you to delete the linodes... maybe this is handled and will error and reconcile gracefully?

mhmxs: @luthermonson First we check vpc: https://github.com/linode/cluster-api-provider-linode/pull/62/files#diff-7fca4d96225fe1de8efb3111cc74cd194997da3af44dbde0eedf6db7cc6e257cR202 than we check subnets and if any has node(s) we re-queu the object https://github.com/linode/cluster-api-provider-linode/pull/62/files#diff-7fca4d96225fe1de8efb3111cc74cd194997da3af44dbde0eedf6db7cc6e257cR220 . After all subnets are empty we try to delete VPC. If deletion fails, we re-queue the object https://github.com/linode/cluster-api-provider-linode/pull/62/files#diff-7fca4d96225fe1de8efb3111cc74cd194997da3af44dbde0eedf6db7cc6e257cR231. I think this covers the case you have mentioned, but please feel free to fixme.

avestuk (commented):
Sorry I had prepared comments but never submitted them!

controller/linodevpc_controller.go
```diff
@@ -0,0 +1,226 @@
...
+		r.Recorder.Event(vpcScope.LinodeVPC, corev1.EventTypeWarning, string(infrav1.CreateVPCError), err.Error())
+
+		return err
+	}
+	vpcScope.LinodeVPC.Status.Ready = true
+
+	return nil
+}
+
+func (r *LinodeVPCReconciler) reconcileUpdate(ctx context.Context, logger logr.Logger, vpcScope *scope.VPCScope) error {
```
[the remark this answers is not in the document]
mhmxs: @avestuk currently LinodeVPC.Spec is ~immutable so we don't have different reconciliation flow. At the moment! Later we should implement custom logic for update.

controller/linodemachine_controller_helpers.go
```diff
@@ -132,7 +136,76 @@ func (r *LinodeMachineReconciler) requestsForCluster(ctx context.Context, namesp
...
+	}
+
+	hasPrimary := false
+	for i := range existingIfaces {
+		if existingIfaces[i].Primary {
+			hasPrimary = true
+
+			break
+		}
+	}
```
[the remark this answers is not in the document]
avestuk: Ok but you better put those 8 bytes to good use somewhere else then.

controller/linodevpc_controller.go
luthermonson: oh i apologize, i didnt see that code come into the PR. you are correct and im good on this

eljohnson92:
I have tested this and confirmed a VPC is created and `LinodeMachines` are attached to that VPC correctly. once other code review comments are resolved this should be good to merge

eljohnson92 (approved):

luthermonson (approved):

controller/linodevpc_controller.go
```diff
@@ -0,0 +1,261 @@
...
+
+		conditions.MarkFalse(vpcScope.LinodeVPC, clusterv1.ReadyCondition, string(infrav1.CreateVPCError), clusterv1.ConditionSeverityError, err.Error())
+
+		r.Recorder.Event(vpcScope.LinodeVPC, corev1.EventTypeWarning, string(infrav1.CreateVPCError), err.Error())
+
+		return err
+	}
+	vpcScope.LinodeVPC.Status.Ready = true
+
+	return nil
```
ChihweiLHBird, line 173: Should we actually create a VPC here?

mhmxs:
@eljohnson92 @avestuk I addressedd all the review items, so could you please do the final review (one more commit) and it should be ready to merge.

avestuk (approved):

eljohnson92 (approved):

PR #64, opened 2024-01-24
add initial LinodeCluster controller logic

### LINKED ISSUE #52: Implement cluster controller
Reconciliation loop of the cluster controller need to be implemented.

Depends on: https://github.com/linode/cluster-api-provider-linode/issues/10
Depends on: https://github.com/linode/cluster-api-provider-linode/issues/4

eljohnson92:
initial logic for the linodeCluster controller. creates a nodebalancer for the API endpoint and automatically sets the nodebalancer IP as the ControlPlaneEndpoint.

I've tested this locally and validated nodebalancers get created and deleted with `LinodeCluster` resource creation/deletion and it gets associated with the owning `Cluster` resource correctly

Fixed #52


cloud/services/loadbalancers.go
```diff
@@ -0,0 +1,141 @@
...
+	"strconv"
+
+	"github.com/go-logr/logr"
+	"github.com/linode/cluster-api-provider-linode/cloud/scope"
+	"github.com/linode/cluster-api-provider-linode/util"
+	"github.com/linode/linodego"
+)
+
+var (
+	defaultLBPort = 6443
```
luthermonson, line 19:
```suggestion
	defaultNBPort = 6443
```

luthermonson: thoughts on NB everywhere vs. going back and forth? Maybe LB everywhere is right since NodeBalancer is the implementation?

cloud/services/loadbalancers.go
```diff
@@ -0,0 +1,141 @@
...
+		return nil, err
+	}
+	logger.Info("Creating NodeBalancer")
+	if linodeNBs, err = clusterScope.LinodeClient.ListNodeBalancers(ctx, linodego.NewListOptions(1, string(rawFilter))); err != nil {
+		logger.Info("Failed to list NodeBalancers", "error", err.Error())
+
+		return nil, err
+	}
+
+	switch len(linodeNBs) {
```
luthermonson, line 44: this switch and the bottom of this function is hard to follow, e.g. things like reusing `err`, checking `linodeNB != nil` a couple times, default is too many nodebalancers? any chance you'd have motivation to rewrite this with conditionals and returns?
```
if len(linodeNBs) > 1 {
   return nil, errors.New("too many")
}
len(linodeNBs) == 1 {
  // already exists logic
  return &linodeNBs[0], nil  
}

// create logic 

return linodeNB, nil

```

cloud/services/loadbalancers.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+var (
+	defaultLBPort = 6443
+)
+
+// CreateNodeBalancer creates a new NodeBalancer if one doesn't exist
+func CreateNodeBalancer(ctx context.Context, clusterScope *scope.ClusterScope, logger logr.Logger) (*linodego.NodeBalancer, error) {
+	var linodeNBs []linodego.NodeBalancer
+	var linodeNB *linodego.NodeBalancer
```
luthermonson, line 25: can we do no pointer on both of these (i think you need for ListNodeBalancers) and return with & so we don't switch between val/ref a couple times in this function?

cloud/services/loadbalancers.go
```diff
@@ -0,0 +1,141 @@
...
+		logger.Error(err, "Panic! Failed to create NodeBalancer")
+
+		return nil, err
+	}
+
+	return linodeNB, nil
+}
+
+// CreateNodeBalancerConfig creates NodeBalancer config if it does not exist
+func CreateNodeBalancerConfig(ctx context.Context, clusterScope *scope.ClusterScope, logger logr.Logger) (*linodego.NodeBalancerConfig, error) {
```
luthermonson, line 99: similar feedback for this func as `CreateNodeBalancer`, i guess whatever we decide there do the same thing in this one too

cloud/services/loadbalancers.go
eljohnson92: the consistency I was thinking of here is if a setting could be applied to any load balancer type supported by this provider then use `LB` and if it was specific to setting up nodebalancers themselves and wouldn't apply to other future loadbalancer types then use `NB` happy to standardize on just NB if we want to kick future use cases down the road

cloud/services/loadbalancers.go
eljohnson92: This can be simplified even more now that I switched it to find based on `label` since there can only ever be at max one. let me rewrite this logic to be more simple.

api/v1alpha1/linodecluster_types.go
```diff
@@ -47,13 +47,13 @@ type LinodeClusterStatus struct {
 	// reconciling the Machine and will contain a succinct value suitable
```
AshleyDumaine, line 47: s/Machine/Cluster ? Same on line 53.

controller/linodecluster_controller.go
```diff
@@ -18,45 +18,216 @@ package controller
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 
-	infrastructurev1alpha1 "github.com/linode/cluster-api-provider-linode/api/v1alpha1"
```
AshleyDumaine, line 27: Could we maybe do `infrav1alpha1` since this isn't actually v1 yet?

controller/linodecluster_controller.go
```diff
@@ -18,45 +18,216 @@ package controller
...
-		Complete(r)
+	controller, err := ctrl.NewControllerManagedBy(mgr).
+		For(&infrav1.LinodeCluster{}).
+		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(mgr.GetLogger(), r.WatchFilterValue)).
+		Build(r)
+	if err != nil {
+		return fmt.Errorf("failed to build controller: %w", err)
+	}
+
+	return controller.Watch(
```
AshleyDumaine, line 228: I think similar to above we'll want to do something like below to add context around the error instead of just returning it as-is:
```suggestion
	if err = controller.Watch(
		...
	); err != nil {
		return fmt.Errorf("failed adding a watch for ready clusters": %w", err)
	}
	
	return nil
```

controller/linodecluster_controller.go
eljohnson92: good catch, updated

api/v1alpha1/linodecluster_types.go
eljohnson92: fixed

controller/linodecluster_controller.go
```diff
@@ -18,45 +18,216 @@ package controller
...
+	}
+
+	if err := clusterScope.LinodeClient.DeleteNodeBalancer(ctx, clusterScope.LinodeCluster.Spec.Network.NodeBalancerID); err != nil {
+		logger.Info("Failed to delete Linode NodeBalancer", "error", err.Error())
+
+		// Not found is not an error
+		apiErr := linodego.Error{}
+		if errors.As(err, &apiErr) && apiErr.Code != http.StatusNotFound {
+			setFailureReason(clusterScope, cerrs.DeleteClusterError, err, r)
+
```
AshleyDumaine, line 205: We should probably also record a warning event:
```
r.Recorder.Eventf(clusterScope.LinodeCluster, corev1.EventTypeWarning, "ClusterReconcilerDeleteFailed", err.Error())
```

controller/linodecluster_controller.go
```diff
@@ -18,45 +18,216 @@ package controller
...
+			return res, err
+		}
+	}
+
+	clusterScope.LinodeCluster.Status.Ready = true
+	conditions.MarkTrue(clusterScope.LinodeCluster, clusterv1.ReadyCondition)
+
+	r.Recorder.Event(clusterScope.LinodeCluster, corev1.EventTypeNormal, string(clusterv1.ReadyCondition), "Load balancer is ready")
+
+	return
```
AshleyDumaine, line 152:
```suggestion
	return ctrl.Result{}, nil
```

controller/linodecluster_controller.go
```diff
@@ -18,45 +18,216 @@ package controller
...
+
+	// Handle deleted clusters
+	if !clusterScope.LinodeCluster.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, r.reconcileDelete(ctx, logger, clusterScope)
+	}
+
+	controllerutil.AddFinalizer(clusterScope.LinodeCluster, infrav1alpha1.GroupVersion.String())
+	// Create
+	if clusterScope.LinodeCluster.Spec.ControlPlaneEndpoint.Host == "" {
+		if err := r.reconcileCreate(ctx, clusterScope, logger); err != nil {
```
AshleyDumaine, line 142:
```
return ctrl.Result{}, r.reconcileCreate(ctx, clusterScope, logger)
```
for consistency with above `reconcileDelete` (also swap the argument order for consistency between `reconcileCreate` and `reconcileDelete`)

controller/linodecluster_controller.go
```diff
@@ -18,45 +18,216 @@ package controller
...
 )
 
 // LinodeClusterReconciler reconciles a LinodeCluster object
 type LinodeClusterReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Recorder         record.EventRecorder
+	LinodeApiKey     string
+	WatchFilterValue string
+	Scheme           *runtime.Scheme
```
AshleyDumaine, line 60: I think we can remove the Scheme? I don't see it used anywhere

controller/linodecluster_controller.go
```diff
@@ -18,45 +18,216 @@ package controller
...
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.16.0/pkg/reconcile
+
 func (r *LinodeClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	ctx, cancel := context.WithTimeout(ctx, reconciler.DefaultedLoopTimeout(r.ReconcileTimeout))
+	defer cancel()
+
+	logger := ctrl.LoggerFrom(ctx).WithName("LinodeClusterReconciler").WithValues("name", req.NamespacedName.String())
+	LinodeCluster := &infrav1alpha1.LinodeCluster{}
```
AshleyDumaine, line 76:
```suggestion
	linodeCluster := &infrav1alpha1.LinodeCluster{}
```

cloud/services/loadbalancers.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+
+		return &linodeNBs[0], nil
+	}
+
+	logger.Info(fmt.Sprintf("Creating NodeBalancer %s-api-server", clusterScope.LinodeCluster.Name))
+	createConfig := linodego.NodeBalancerCreateOptions{
+		Label:              util.Pointer(fmt.Sprintf("%s-api-server", clusterScope.LinodeCluster.Name)),
+		Region:             clusterScope.LinodeCluster.Spec.Region,
+		ClientConnThrottle: nil,
```
AshleyDumaine, line 58: I think we should be able to leave this off if we're not setting it or do `util.Pointer(0)` to be explicit

cloud/services/loadbalancers.go
```diff
@@ -0,0 +1,115 @@
...
+	if linodeNB, err = clusterScope.LinodeClient.CreateNodeBalancer(ctx, createConfig); err != nil {
+		logger.Info("Failed to create Linode NodeBalancer", "error", err.Error())
+
+		// Already exists is not an error
+		apiErr := linodego.Error{}
+		if errors.As(err, &apiErr) && apiErr.Code != http.StatusFound {
+			return nil, err
+		}
+
+		err = nil
```
AshleyDumaine, line 71:
```suggestion
```

controller/linodecluster_controller.go
eljohnson92: in this case we don't want to return if there is an error so we can set the status and conditions I originally kept those outside the scope of the create function so we can add update functions later but I can combine them if we think that would be better

controller/linodecluster_controller.go
```diff
@@ -18,45 +18,220 @@ package controller
...
+			reterr = err
+		}
+	}()
+
+	// Handle deleted clusters
+	if !clusterScope.LinodeCluster.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, r.reconcileDelete(ctx, logger, clusterScope)
+	}
+
+	controllerutil.AddFinalizer(clusterScope.LinodeCluster, infrav1alpha1.GroupVersion.String())
```
luthermonson, line 138: i see some capi providers will call a patch immediately after adding the finalizer to make sure it's there before doing anything else e.g. https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awscluster_controller.go#L275-L280 

maybe worth doing too?

cloud/services/loadbalancers.go
eljohnson92: These return values are set by the linodego library so we are going to punt on trying to standardize these better in our code for now

controller/linodecluster_controller.go
eljohnson92: moved finalizer logic into a function that always does a patch immediately after

luthermonson (approved):

PR #65, opened 2024-01-25
Fix dependency downloads by architecture in Makefile

mhmxs:
Originally `amd64` was hard coded to the download path of some dependencies, this fix make it architecture "independent".


eljohnson92:
does this supersede #53?

luthermonson:
> does this supersede #53?

i prefer this method, @eljohnson92 if you do to give it an approval and close #53 and let's move on

luthermonson (approved):

eljohnson92:
I also like this method, LGTM

eljohnson92 (approved):

Issue #7, opened 2023-09-28
Testing


luthermonson:
- e2e with a linode token... find out how we can get a capi testing account?
- Mocks? Look into how CAPA was mocking kube api and aws api calls
- ginkgo?

mhmxs:
@luthermonson may i suggest to use https://kuttl.dev/ to provision Kubernetes and execute e2e (or other tests which fits into) tests? We have good experiences with it, and is is simple as possible.

luthermonson:
if you think itll help us get something that looks like e2e in both local and a gh action then let's give it a shot

mhmxs:
@luthermonson here is a very basic test case in Kuttl (plus environment) https://github.com/linode/cluster-api-provider-linode/pull/62/commits/b0ce89799e56c743991cbced3c1f80352076b5b0

PR #67, opened 2024-01-26
Introduce E2E framework

mhmxs:
This change introduces Kuttl as e2e framework and CI integration. Enjoy!

Good begining of: https://github.com/linode/cluster-api-provider-linode/issues/7


Makefile
```diff
@@ -140,9 +147,11 @@ install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~
 uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 	$(KUSTOMIZE) build config/crd | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
 
-.PHONY: deploy
-deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+_setImage: kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+
+.PHONY: deploy
+deploy: manifests kustomize _setImage ## Deploy controller to the K8s cluster specified in ~/.kube/config.
```
eljohnson92, line 154: could we use `tilt ci` to do this instead of manually deploying things here or is it worth testing this outside of tilt as well?

mhmxs: @eljohnson92 We have two different environments. Tilt for development and Kuttl for testing. It doesn't make much sense to introduce cross project complexity to avoid a make target. And this method has been generated by Kubebuilder so ~all Kubebuilder projects have this function. Anyway I'm open to other solutions, my focus was KIS. It is possible to call Tilt from Kuttl to deploy the system before executing the tests. Or it is also possible to start the environment via Tilt and use Kuttl only to execute the tests themselves (we can start Kuttl from Tilt). So there are options. WDYT?

.github/workflows/build_test_ci.yml
```diff
@@ -114,7 +164,7 @@ jobs:
     - name: Docker cache
       uses: ScribeMD/docker-cache@0.3.7
       with:
-        key: docker-${{ runner.os }}-${{ hashFiles('Dockerfile') }}
+        key: docker-${{ runner.os }}-${{ hashFiles('Makefile') }}
```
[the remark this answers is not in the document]
mhmxs: @eljohnson92 Dockerfile changes very rarely, that's why i selected a more frequent file. The nice solution would be to drop cache only when code or dependencies have been changed. I just didn't want to spend much time on a sophisticated solution in this early stage of the project.

Makefile
eljohnson92: given this is for test I don't want to block review on this, my preference would be we use Tilt to call `tilt ci` instead of `make deploy` so we can just use the existing deployment interface. if you have a strong opinion that we should leave it as it is for now I would be ok with leaving this as it and giving it a thumbs up. let me know what you think

eljohnson92 (approved):

luthermonson (approved):

mhmxs:
Reopen PR on base repo: https://github.com/linode/cluster-api-provider-linode/pull/77

PR #71, opened 2024-01-30
E2E test machine controller

mhmxs:
Basic machine create and delete test of machine controller.


util/reconciler/defaults.go
```diff
@@ -30,6 +30,13 @@ const (
 	DefaultMachineControllerWaitForRunningDelay = 5 * time.Second
 	// DefaultMachineControllerWaitForRunningTimeout is the default timeout if instance is not running.
 	DefaultMachineControllerWaitForRunningTimeout = 20 * time.Minute
+	// DefaultMachineControllerLinodeImage default image.
+	DefaultMachineControllerLinodeImage = "linode/ubuntu23.10"
```
eljohnson92, line 34: are there any concerns around us having to maintain updates on this mapping? if we are ok with this default perhaps we should use a LTS image so we have longer support?

.github/workflows/build_test_ci.yml
```diff
@@ -89,6 +89,59 @@ jobs:
...
+
+    - name: Docker cache
+      uses: ScribeMD/docker-cache@0.3.7
+      with:
+        key: docker-${{ runner.os }}-${{ hashFiles('go.sum') }}}
+
+    - name: E2E test
+      run: make e2etest
+      env:
+        LINODE_TOKEN: ${{ secrets.LINODE_TOKEN }}
```
luthermonson, line 137: this has been created now and you can test, let me know if you run into any problems with the token i can get them resolved for you

Issue #3, opened 2023-09-28
Upstream Bootstrap Provider for Kubeadm


luthermonson:
Get kubeadm working with metadata service.

- should take a LINODE_TOKEN
- create...
  - ***node balancer*** for HA control plane?
  - machine pools/deployments? ***linodes*** ?
  - investigate ccm/csi? avoid using ccm for the control plane?

Issue #1, opened 2023-09-28
Investigate Metadata network-config


luthermonson:
There is some limited functionality to how MDS uses network config and we should investigate which features we may be missing from capi/cni/etc. if we solely rely on the MDS.

- Do we need to disable network helpers?

Possible good reading: https://www.linode.com/docs/products/compute/compute-instances/guides/metadata/?tabs=linode-api

eljohnson92:
the default cluster template is now using metadata correctly, closing this out

PR #74, opened 2024-01-31
[feat]: Set metadata for LinodeMachines, disable swap

AshleyDumaine:
This modifies the `LinodeMachine` controller to wait for bootstrap data to be ready before creating Linode instances since the cloud-config needs to be stored in the metadata field to configure the Linode properly. Additionally sets swap to be disabled since kubeadm fails with swap enabled.

Related issues:
- https://github.com/linode/cluster-api-provider-linode/issues/1
- https://github.com/linode/cluster-api-provider-linode/issues/3

### DIFF
--- a/cloud/scope/machine.go
+++ b/cloud/scope/machine.go
@@ -2,26 +2,26 @@ package scope
 
 import (
 	"context"
-	b64 "encoding/base64"
 	"errors"
 	"fmt"
 
+	"github.com/linode/linodego"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
-
-	infrav1 "github.com/linode/cluster-api-provider-linode/api/v1alpha1"
-	"github.com/linode/linodego"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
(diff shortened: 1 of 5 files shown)


cloud/scope/machine.go
```diff
@@ -69,3 +74,35 @@ func NewMachineScope(apiKey string, params MachineScopeParams) (*MachineScope, e
...
+	value, ok := secret.Data["value"]
+	if !ok {
+		return "", fmt.Errorf(
+			"bootstrap data secret value key is missing for LinodeMachine %s/%s",
+			m.LinodeMachine.Namespace,
+			m.LinodeMachine.Name,
+		)
+	}
+
+	return b64.StdEncoding.EncodeToString(value), nil
```
luthermonson, line 107: should the b64 encoding happen by the callee incase anyone is looking for the string? this feels like a generic function and how it's used shouldn't be determined by this func

controller/linodemachine_controller.go
```diff
@@ -209,6 +209,9 @@ func (r *LinodeMachineReconciler) reconcile(
 		}
 	}()
 
+	// Add the finalizer if not already there
+	controllerutil.AddFinalizer(machineScope.LinodeMachine, infrav1.GroupVersion.String())
```
luthermonson, line 213: since youre playing with this line could you add a patch like @eljohnson92 PR earlier to cluster https://github.com/linode/cluster-api-provider-linode/pull/64/files#diff-d6148d3b8d79d76ad02a5e26b74a8bbff6541058ae664f3c10e5dd97accffd76R99-R113

AshleyDumaine: Good call, I'll get that added 👍

luthermonson (approved):

PR #75, opened 2024-02-01
Add IgnoreKubeNotFound helper function

mhmxs:
Kubeclient Patch function returns an aggregate error, so we need a helper function to identify not found errors.

### DIFF
--- a/util/helpers.go
+++ b/util/helpers.go
@@ -1,6 +1,28 @@
 package util
 
+import (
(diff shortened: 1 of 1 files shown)


eljohnson92:
@mhmxs I think we can use filterOut like we used in the [cluster controller](https://github.com/linode/cluster-api-provider-linode/blob/main/controller/linodecluster_controller.go#L127) for this for this use case possibly?

PR #77, opened 2024-02-05
E2E framework

mhmxs:
At the time of this PR repo secrets are not available on forked repos, so this PR is a replication of https://github.com/linode/cluster-api-provider-linode/pull/67


eljohnson92:
Should we also close out #71 in favor of just this PR or are there additional features in that PR we want as well?

mhmxs:
> Should we also close out #71 in favor of just this PR or are there additional features in that PR we want as well?

Exactly this PR is an empty framework, and the mentioned PR adds machine test and some fixes to go green with machine related tests.

avestuk (approved):

PR #81, opened 2024-02-06
[feat]: Add private IP on Linodes and add control plane nodes to NodeBalancer backend nodes

AshleyDumaine:
We need to enable private IPs on created Linodes in order to add control plane nodes to the NodeBalancer. 
This change will add any new control plane nodes to the NB backend automatically which does eventually go to "up" on the status check in Cloud Manager.

### DIFF
--- a/cloud/services/loadbalancers.go
+++ b/cloud/services/loadbalancers.go
@@ -208,15 +208,13 @@ func DeleteNodeFromNB(
 		return nil
 	}
 
+	if machineScope.LinodeMachine.Spec.InstanceID == nil {
+		return errors.New("no InstanceID")
+	}
+
 	linodeNBConfig, err := GetNodeBalancerConfig(ctx, clusterScope, logger)
 	if util.IgnoreLinodeAPIError(err, http.StatusNotFound) != nil || linodeNBConfig == nil {
 		logger.Error(err, "Failed to get Node Balancer config")
-
-		return err
-	}
-
-	if machineScope.LinodeMachine.Spec.InstanceID == nil {
-		err = errors.New("no InstanceID")
 
 		return err
 	}
--- a/cloud/services/loadbalancers.go
+++ b/cloud/services/loadbalancers.go
@@ -13,11 +13,14 @@ import (
 	"github.com/linode/cluster-api-provider-linode/cloud/scope"
 	"github.com/linode/cluster-api-provider-linode/util"
 	"github.com/linode/linodego"
+	kutil "sigs.k8s.io/cluster-api/util"
 )
 
 var (
 	defaultLBPort = 6443
 )
+
+const expectedIPV4Count = 2
 
 // CreateNodeBalancer creates a new NodeBalancer if one doesn't exist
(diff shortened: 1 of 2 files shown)


cloud/services/loadbalancers.go
```diff
@@ -102,11 +109,125 @@ func CreateNodeBalancerConfig(ctx context.Context, clusterScope *scope.ClusterSc
...
+	linodeInstance *linodego.Instance,
+) error {
+	// Update the NB backend with the new instance if it's a control plane node
+	if !kutil.IsControlPlaneMachine(machineScope.Machine) {
+		return nil
+	}
+	// Get the private IP that was assigned
+	ips := linodeInstance.IPv4
+	if len(ips) < expectedIPV4Count {
+		err := errors.New("no private IP address")
```
avestuk, line 165: Are we able to assert that there's no private IP address if we have fewer than 2 addresses? Is it also possible that the only address in the machine spec is a private IP?

AshleyDumaine: I don't think that should be possible to have an instance created without a public IP, but to be safe I can add another [API call](https://pkg.go.dev/github.com/linode/linodego#Client.GetInstanceIPAddresses) to actually return the IPs explicitly for public vs private

avestuk: I'm pretty sure you can do it if you provision a Linode inside a VPC. 

The *net.IP struct has an IsPrivate() method that you can use to determine public vs private. That should work for the VPC address ranges too.

AshleyDumaine: Ah gotcha. I switched to just getting the private address(es) for the instance instead of checking each IP in the range of assigned addresses  for the instance.

avestuk (approved):

avestuk (approved):

cloud/services/loadbalancers.go
```diff
@@ -102,11 +107,129 @@ func CreateNodeBalancerConfig(ctx context.Context, clusterScope *scope.ClusterSc
...
+		return err
+	}
+	if len(addresses.IPv4.Private) == 0 {
+		err := errors.New("no private IP address")
+		logger.Error(err, "no private IPV4 addresses set for LinodeInstance")
+
+		return err
+	}
+
+	linodeNBConfig, err := GetNodeBalancerConfig(ctx, clusterScope, logger)
```
eljohnson92, line 174: would it be helpful to store the `NodeBalancerConfigID` as a status on the cluster so we can save ourselves a API call for these operations?

AshleyDumaine: Yeah, sounds good to me 👍

eljohnson92 (approved):

cloud/services/loadbalancers.go
AshleyDumaine: Just to follow-up here, I think more refactoring work is needed to make this change (how we store NB IDs and default the port, potentially needing a defaulting admission webhook for the network config). I've put those changes into a different branch to make a follow-up PR addressing those later on.

Issue #76, opened 2024-02-01
os.Exit() call drops last log message


mhmxs:
It is hard to identify errors in `main`, because the system doesn't have enough time to flush logs.

https://github.com/linode/cluster-api-provider-linode/blob/1bfa6dc02aebd64984cae98d48e8ebee943f0c99/cmd/main.go#L61

cbzzz:
I think the issue was with that one specific error message gets called before the `setupLogger` is setup [here](https://github.com/linode/cluster-api-provider-linode/blob/1bfa6dc02aebd64984cae98d48e8ebee943f0c99/cmd/main.go#L82). Please let me know if https://github.com/linode/cluster-api-provider-linode/pull/82 fixes your issue.

PR #80, opened 2024-02-05
Basic E2E test for machine controller

mhmxs:
At the time of this PR repo secrets are not available on forked repos, so this PR is a replication of https://github.com/linode/cluster-api-provider-linode/pull/71

> Some small bugfixes came around the test execution itself, so business logic didn't change compare to the original PR.

PR #82, opened 2024-02-06
fixup! Missing LINODE_TOKEN error message

cbzzz:
# Description
Addresses: https://github.com/linode/cluster-api-provider-linode/issues/76#issue-2111484116

# Notes
I'm not sure we should be using a global `LINODE_TOKEN` for the operator (and all it's managed workload clusters) here. We should allow workload clusters to specify the Linode account in which to provision its resources.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -55,17 +55,16 @@ func init() {
 }
 
 func main() {
-	linodeToken := os.Getenv("LINODE_TOKEN")
-	if linodeToken == "" {
-		setupLog.Error(errors.New("failed to get LINODE_TOKEN environment variable"), "unable to start operator")
-		os.Exit(1)
-	}
+	var (
+		// Environment variables
+		linodeToken string = os.Getenv("LINODE_TOKEN")
 
-	var machineWatchFilter string
-	var clusterWatchFilter string
-	var metricsAddr string
-	var enableLeaderElection bool
-	var probeAddr string
+		machineWatchFilter   string
+		clusterWatchFilter   string
+		metricsAddr          string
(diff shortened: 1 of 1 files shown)


mhmxs (approved):
It makes sense, lots of other project uses `setupLog` in the same way.

eljohnson92:
@mhmxs is it expected that the e2e-test failure just because the test is coming from a fork?

cbzzz:
@mhmxs It seems like the E2E test doesn't have the LINODE_TOKEN environment variable set. Is this expected? Thanks!

https://github.com/linode/cluster-api-provider-linode/actions/runs/7805739210/job/21292435053#step:7:346

AshleyDumaine:
@cbzzz I think you would need to set the LINODE_TOKEN secret on your fork for GHA to pass: https://github.com/linode/cluster-api-provider-linode/blob/main/.github/workflows/build_test_ci.yml#L139C35-L139C47

mhmxs:
> Is this expected?

@cbzzz I our current setup: Yes

mhmxs:
> set the LINODE_TOKEN

@cbzzz @AshleyDumaine keep in mind, that test execution should cost money. At the time of this we don't have any tests, but this should change soon: https://github.com/linode/cluster-api-provider-linode/pull/80

eljohnson92:
I think there is a larger discussion about when we want to run these e2e tests but I don't think that should be blocking on this PR. I'm good to merge as is if there are no other comments

eljohnson92 (approved):

AshleyDumaine (approved):

PR #83, opened 2024-02-06
Machine e2e test with bring your own VPC

mhmxs:
This test suit creates a VPC via raw APi call, and uses the created VPC as a target of Linode machine creation.

Related to: https://github.com/linode/cluster-api-provider-linode/issues/7


e2e/Makefile
```diff
@@ -1,5 +1,20 @@
-runTestCase:
-	@T="$$(KUBECONFIG="$$ROOT_DIR/kubeconfig" kuttl test --timeout 300 --skip-delete --namespace "$$NAMESPACE" "$$TC" 2>&1)" ;\
+ROOT_DIR ?= $(shell git rev-parse --show-toplevel)
+TARGET_API ?= api.linode.com
+TARGET_API_VERSION ?= v4beta
+
```
luthermonson, line 4: i think you need `$(KUTTL)` and `$(ENVSUBST)` which points to the local ./bin dir of downloaded binaries, this looks like you're assuming those are already in your path

mhmxs: @luthermonson you are right, e2e test related makefile tasks expect a well configured developement environment. Anyway This is not related to the PR, but do you prefer any reproducable environment like [Nix](https://nixos.org/)?

eljohnson92: We have been using devbox in some other projects which generally works very nicely. if we want to consider that for this as well I can create an issue and propose an update to support  it

mhmxs:
@eljohnson92 @avestuk i rebased this change to main. Please take a look.

avestuk (approved):

PR #84, opened 2024-02-06
[tweak]: refactoring CRDs for spec vs status

AshleyDumaine:
This shuffles some fields around since `*ID` fields should belong in the Status instead of the Spec for `LinodeMachines` and `LinodeClusters`.
This additionally stores the created NB Config ID on the `LinodeCluster` to save us an API call (follow-up from #81).


api/v1alpha1/linodemachine_types.go
```diff
@@ -28,13 +28,6 @@ import (
 
 // LinodeMachineSpec defines the desired state of LinodeMachine
 type LinodeMachineSpec struct {
-	// ProviderID is the unique identifier as specified by the cloud provider.
-	// +optional
-	ProviderID *string `json:"providerID,omitempty"`
```
eljohnson92, line 33: https://cluster-api.sigs.k8s.io/developer/providers/machine-infrastructure I think at least ProviderID at least needs to stay in the spec

mhmxs: It was weird for me too, I also wanted to put `ProviderID` and `InstanceID` to status, but it seems spec is the official way :(

PR #90, opened 2024-02-08
fix: reconciliation spam for controllers 

cbzzz:
fix: Stop controllers from constantly reconciling their objects due to overly-aggressive patching in `cloud/scope.AddFinalizer` methods

### DIFF
--- a/cloud/scope/cluster.go
+++ b/cloud/scope/cluster.go
@@ -91,9 +91,12 @@ func (s *ClusterScope) Close(ctx context.Context) error {
 	return s.PatchObject(ctx)
 }
 
-// AddFinalizer adds a finalizer and immediately patches the object to avoid any race conditions
+// AddFinalizer adds a finalizer if not present and immediately patches the
+// object to avoid any race conditions.
 func (s *ClusterScope) AddFinalizer(ctx context.Context) error {
-	controllerutil.AddFinalizer(s.LinodeCluster, infrav1alpha1.GroupVersion.String())
+	if controllerutil.AddFinalizer(s.LinodeCluster, infrav1alpha1.GroupVersion.String()) {
+		return s.Close(ctx)
+	}
 
-	return s.Close(ctx)
+	return nil
 }
--- a/cloud/scope/machine.go
+++ b/cloud/scope/machine.go
@@ -85,11 +85,14 @@ func (s *MachineScope) Close(ctx context.Context) error {
 	return s.PatchObject(ctx)
 }
 
-// AddFinalizer adds a finalizer and immediately patches the object to avoid any race conditions
+// AddFinalizer adds a finalizer if not present and immediately patches the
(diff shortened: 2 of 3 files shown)


api/v1alpha1/linodevpc_types.go
```diff
@@ -34,8 +34,6 @@ type LinodeVPCSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	// +optional
 	Label string `json:"label,omitempty"`
-	// +optional
-	Description string `json:"description,omitempty"`
```
mhmxs, line 38: The VPC object has `description` fireld: https://github.com/linode/linodego/blob/0bd0c941538822724dbd5cd23c990c9ac7717b17/vpc.go#L25 why we need to delete it?

cloud/scope/cluster.go
```diff
@@ -91,9 +91,11 @@ func (s *ClusterScope) Close(ctx context.Context) error {
...
 }
 
-// AddFinalizer adds a finalizer and immediately patches the object to avoid any race conditions
+// AddFinalizer adds a finalizer if not present and immediately patches the
+// object to avoid any race conditions.
 func (s *ClusterScope) AddFinalizer(ctx context.Context) error {
-	controllerutil.AddFinalizer(s.LinodeCluster, infrav1alpha1.GroupVersion.String())
-
-	return s.Close(ctx)
+	if controllerutil.AddFinalizer(s.LinodeCluster, infrav1alpha1.GroupVersion.String()) {
```
AshleyDumaine, line 97: Can we add a comment here and on the other files explaining we only do Close if the finalizer was actually updated?

api/v1alpha1/linodevpc_types.go
cbzzz: Ah, thanks! I must have missed that. I thought the field was completely unused.

AshleyDumaine:
I've confirmed there is _much_ less spam in the CAPI-linode-controller logs with this change, namely, reconciles on updates are not being spammed for LinodeMachines when creating a Cluster. I'm not 100% sure we need the predicates for the VPC, but the changes around `Close` only if the finalizer changed seems to have helped. 🎉

cbzzz:
> I've confirmed there is _much_ less spam in the CAPI-linode-controller logs with this change, namely, reconciles on updates are not being spammed for LinodeMachines when creating a Cluster. I'm not 100% sure we need the predicates for the VPC, but the changes around `Close` only if the finalizer changed seems to have helped. 🎉

I'm fine with removing the Updates predicates until we know we actually need them.

eljohnson92:
can we limit this scope to just the patching parts since it looks like that is enough to fix the spamming of reconciling

eljohnson92 (approved):

AshleyDumaine (approved):

PR #93, opened 2024-02-08
fixup! e2e: Support echo on darwin

bcm820:
See https://github.com/linode/cluster-api-provider-linode/pull/80#discussion_r1483123483

By default, kuttl uses `sh` native to the OS, and Mac's builtin echo does not support passing `-n`.


e2e/Makefile
```diff
@@ -31,7 +31,7 @@ renderTestCase:
 	@D="$$(mktemp -d)" ;\
 	mkdir -p "$$D/case" ;\
 	$(BIN_DIR)/envsubst -i "$$TPL" -o "$$D/case/00-case.yaml" ;\
-	echo -n "$$D"
+	echo "$$D\c"
```
AshleyDumaine, line 34: this works on my mac but I see a literal \c on my linux VM. Could we instead do `printf "$$D\n"`? That seems consistent on both

AshleyDumaine (approved):
LGTM, works on both Linux and Mac for me 👍

PR #95, opened 2024-02-08
wait until after delete step to check that linodeCluster still exists

eljohnson92:
This PR fixes an issue where checking for a cluster existing happens while stuff is in the process of deleting and some resources don't get deleted. just move the check until after we validate if we are doing a delete operation. this also changes the tagging in cloud manager to be based on cluster `name` instead of `uuid` for easier readability.

### DIFF
--- a/cloud/services/loadbalancers.go
+++ b/cloud/services/loadbalancers.go
@@ -177,6 +177,12 @@ func DeleteNodeFromNB(
 		return errors.New("no InstanceID")
 	}
 
+	if machineScope.LinodeCluster.Spec.ControlPlaneEndpoint.Host == "" {
+		logger.Info("NodeBalancer already deleted, no NodeBalancer backend Node to remove")
+
+		return nil
+	}
+
 	err := machineScope.LinodeClient.DeleteNodeBalancerNode(
 		ctx,
 		machineScope.LinodeCluster.Spec.Network.NodeBalancerID,
--- a/controller/linodemachine_controller.go
+++ b/controller/linodemachine_controller.go
@@ -163,18 +163,6 @@ func (r *LinodeMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 
 	linodeCluster := &infrav1alpha1.LinodeCluster{}
-	linodeClusterKey := client.ObjectKey{
-		Namespace: linodeMachine.Namespace,
-		Name:      cluster.Spec.InfrastructureRef.Name,
-	}
-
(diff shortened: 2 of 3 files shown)


cloud/services/loadbalancers.go
```diff
@@ -177,6 +177,12 @@ func DeleteNodeFromNB(
 		return errors.New("no InstanceID")
 	}
 
+	if machineScope.LinodeCluster.Spec.ControlPlaneEndpoint.Host == "" {
+		logger.Info("NodeBalancer already deleted, no NodeBalancer config to delete")
```
AshleyDumaine, line 181:
```suggestion
		logger.Info("NodeBalancer already deleted, no NodeBalancer backend Node to remove")
```

AshleyDumaine (approved):
Tested locally and works for me 👍

controller/linodemachine_controller.go
```diff
@@ -258,6 +246,19 @@ func (r *LinodeMachineReconciler) reconcile(
...
+		Namespace: machineScope.LinodeMachine.Namespace,
+		Name:      machineScope.Cluster.Spec.InfrastructureRef.Name,
+	}
+
+	if err = r.Client.Get(ctx, linodeClusterKey, machineScope.LinodeCluster); err != nil {
+		if err = client.IgnoreNotFound(err); err != nil {
+			logger.Error(err, "Failed to fetch Linode cluster")
+		}
+
+		return ctrl.Result{}, err
```
avestuk, line 259: I hate being put in this position where I'm advocating for naked returns... but we're using naked returns in the majority of places in this function. @AshleyDumaine snuck one past me in Create but I think we should be consistent. 

I think that we need the named returns in order to handle the event creation on errors, so we should keep these returns in their birthday suits.

mhmxs: Programmatically correct, because `if err = r.Client.Get` uses the original error, so we do not defer one error and return another.

avestuk: That is true, this is purely a style question. I can also see that it's not really particularly important I just think that using named returns is somewhat jarring anyway so I'd prefer that we keep it very consistent inside functions where we are using named returns. 

I would add that this is not a hill I wish to die on.

mhmxs: > keep it very consistent

Totally agree. Empty returns are markers/hints that the function does something different. In our case, it enforces update of the status.

eljohnson92: changed the return to an empty return

avestuk (approved):

PR #96, opened 2024-02-08
[feat]: add cluster template, use CAAPH to install cilium for the CNI

### LINKED ISSUE #3: Upstream Bootstrap Provider for Kubeadm

AshleyDumaine:
This adds CAAPH (https://github.com/kubernetes-sigs/cluster-api-addon-provider-helm) into the KIND management created by Tilt to install Cilium via the helm controller. We are doing this instead of using `postKubeadmCommands` because we need to be able to retry in the case of transient failures with reaching the control plane.

Using this template, it is possible to set env vars and use `envstubst` to create a cluster in a one-liner:
```
envsubst "`printf '${%s} ' $(sh -c "env|cut -d'=' -f1")`" < templates/cluster-template.yaml | kubectl apply -f -
```

Result:
```
(⎈|kind-tilt:N/A) ~  clusterctl describe cluster adumaine-1
NAME                                                           READY  SEVERITY  REASON  SINCE  MESSAGE
Cluster/adumaine-1                                             True                     5m28s
├─ClusterInfrastructure - LinodeCluster/adumaine-1             True                     9m55s
├─ControlPlane - KubeadmControlPlane/adumaine-1-control-plane  True                     5m28s
│ └─Machine/adumaine-1-control-plane-27qv8                     True                     8m50s
└─Workers
  └─MachineDeployment/adumaine-1-md-0                          True                     5s
    └─Machine/adumaine-1-md-0-q6nsd-p8fqd                      True                     4m35s

```

closes #3

### DIFF
--- a/templates/addons/cilium-helm.yaml
+++ b/templates/addons/cilium-helm.yaml
@@ -8,7 +8,7 @@ spec:
       cni: cilium
   repoURL: https://helm.cilium.io/
   chartName: cilium
-  version: 0.15.0
+  version: 1.15.0
   options:
     waitForJobs: true
     wait: true
--- a/templates/addons/cilium-helm.yaml
+++ b/templates/addons/cilium-helm.yaml
@@ -1,0 +1,21 @@
+apiVersion: addons.cluster.x-k8s.io/v1alpha1
+kind: HelmChartProxy
+metadata:
+  name: cilium
+spec:
+  clusterSelector:
+    matchLabels:
+      cni: cilium
+  repoURL: https://helm.cilium.io/
+  chartName: cilium
+  options:
+    waitForJobs: true
+    wait: true
+    timeout: 5m
+  valuesTemplate: |
+    hubble:
+      relay:
+        enabled: true
+      ui:
+        enabled: true
+---
--- a/templates/addons/cilium-helm.yaml
+++ b/templates/addons/cilium-helm.yaml
@@ -8,6 +8,7 @@ spec:
       cni: cilium
   repoURL: https://helm.cilium.io/
   chartName: cilium
+  version: 0.15.0
   options:
     waitForJobs: true
     wait: true
--- a/templates/cluster-template.yaml
+++ b/templates/cluster-template.yaml
@@ -204,3 +204,5 @@ stringData:
     modprobe overlay
     modprobe br_netfilter
     sysctl --system
+    sed -i '/swap/d' /etc/fstab
+    swapoff -a
--- a/templates/cluster-template.yaml
+++ b/templates/cluster-template.yaml
@@ -22,7 +22,6 @@ apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
 kind: LinodeCluster
 metadata:
   name: ${CLUSTER_NAME}
-  namespace: default
 spec:
   region: ${LINODE_REGION}
 ---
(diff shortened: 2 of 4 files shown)


templates/cluster-template.yaml
```diff
@@ -0,0 +1,222 @@
...
+      region: ${LINODE_REGION}
+      authorizedKeys:
+        - ${LINODE_SSH_KEY}
+---
+apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+kind: KubeadmConfigTemplate
+metadata:
+  name: ${CLUSTER_NAME}-md-0
+spec:
+  template:
```
eljohnson92, line 176: Is there a way we can have this template apply to both the controlplane and worker nodes?

AshleyDumaine: `KubeadmControlPlaneTemplate` (control plane nodes only) and `KubeadmConfigTemplate` (worker nodes only) don’t seem like they can inherit a common base template AFAICT? I can do some more poking around to see if we can't simplify this.

AshleyDumaine: Looks like we can put the common files into a k8s Secret and then reference that in `KubeadmConfigTemplate.spec.template.spec.files.contentFrom` and `KubeadmControlPlaneTemplate.spec.template.spec.kubeadmConfigSpec.files.con
tentFrom`. We'd have to use a Secret since it doesn't seem like there's ConfigMap support. I can make that tweak on this template if we're okay with that approach to deduplicate the data.

templates/addons/cilium-helm.yaml
avestuk, line 1: Do we have a mechanism to pin the chart version? I gather this is just for KIND right now so I can see that this might not be desirable.

templates/cluster-template.yaml
```diff
@@ -0,0 +1,206 @@
...
+          contentFrom:
+            secret:
+              name: common-init-files
+              key: sysctl-k8s.conf
+        - path: /kubeadm-pre-init.sh
+          contentFrom:
+            secret:
+              name: common-init-files
+              key: kubeadm-pre-init.sh
+          permissions: "0777"
```
avestuk, line 161: the nittiest of nits: do these perms need to be so wide?

AshleyDumaine: updated to `0500`

templates/addons/cilium-helm.yaml
AshleyDumaine: I updated this to set the chart version to the latest stable cilium release (0.15.0)

templates/cluster-template.yaml
```diff
@@ -0,0 +1,208 @@
...
+  infrastructureRef:
+    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
+    kind: LinodeCluster
+    name: ${CLUSTER_NAME}
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
+kind: LinodeCluster
+metadata:
+  name: ${CLUSTER_NAME}
+  namespace: default
```
cbzzz, line 25: Stray namespace?

cbzzz (approved):

eljohnson92 (approved):

PR #102, opened 2024-02-09
e2e: Use full infrastructureRef to delegate owner assignment to CAPI

bcm820:
This PR updates the minimal LinodeCluster controller test to let the CAPI controller assign ownership of the sample LinodeCluster (rather than assigning it manually).


AshleyDumaine:
LGTM, does it make sense to do something similar for the [LinodeMachine controller](https://github.com/linode/cluster-api-provider-linode/blob/main/e2e/linodemachine-controller/minimal/01-create-cluster.yaml#L18) in another PR?

eljohnson92 (approved):

PR #104, opened 2024-02-12
add support for devbox

### LINKED ISSUE #87: investigate using devbox for tool management
[devbox](https://www.jetpack.io/devbox/docs/quickstart/) is a great tool for managing local, directory-specific nix environments for projects. we should investigate using this instead of our Makefile cache directory

eljohnson92:
This PR adds support for [devbox](https://github.com/jetpack-io/devbox), it does not force any tool to use devbox but just supports the devbox version of the tool if it is present. if the Make target is not run within a devbox shell it reverts to the existing Makefile tool caching logic. Additionally, this PR fixes the tool Make targets so they are only run if the tool doesn't exist locally already(fixing a bug where some targets ran every time)

closes #87

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ test: manifests generate fmt vet envtest ## Run tests.
 e2etest:
 	make --no-print-directory _e2etest # Workaround to force the flag on Github Action
 
-_e2etest-infra: kind ctlptl tilt kuttl kustomize clusterctl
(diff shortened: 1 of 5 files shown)


AshleyDumaine:
Can we also update the README to call out the addition of devbox support?

AshleyDumaine (approved):

luthermonson (approved):
🚀

Issue #101, opened 2024-02-09
Add-on for linode-cloud-controller-manager


AshleyDumaine:
We should make it easy for clusters to opt in to having the [Linode CCM](https://github.com/linode/linode-cloud-controller-manager) installed via a label on the Cluster resource so Linode NBs are automatically created for K8s Services of type `LoadBalancer`.

This should be doable via the Helm chart plus CAAPH introduced in https://github.com/linode/cluster-api-provider-linode/pull/96

PR #109, opened 2024-02-12
[fix] providerID must conform to linode://{instanceID} for node external/internal IP setting

AshleyDumaine:
In order for the [linode-ccm](https://github.com/linode/linode-cloud-controller-manager) to correctly set the external and internal IPs for nodes in a cluster, the provider ID needs to be `linode://$(instanceID}`. See:
- https://github.com/linode/linode-cloud-controller-manager/blob/main/cloud/linode/common.go#L9
- https://github.com/linode/linode-cloud-controller-manager/blob/main/cloud/linode/instances.go#L111

### DIFF
--- a/templates/cluster-template.yaml
+++ b/templates/cluster-template.yaml
@@ -70,13 +70,13 @@ spec:
       nodeRegistration:
         kubeletExtraArgs:
           cloud-provider: external
-          provider-id: 'linode:///{{ ds.meta_data.region }}/{{ ds.meta_data.id }}'
+          provider-id: 'linode://{{ ds.meta_data.id }}'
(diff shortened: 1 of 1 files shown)


templates/cluster-template.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: ${CLUSTER_NAME}
   labels:
     cni: cilium
+    ccm: linode
```
AshleyDumaine, line 7: This is for once the CCM is installed via CAAPH (#101) which is needed for k8s 1.29 so the internal/external IP can be set correctly.

eljohnson92 (approved):

PR #112, opened 2024-02-13
install cilium-helm addon using tilt

rahulait:
We added support to use CAAPH to install cilium via helm controller in PR https://github.com/linode/cluster-api-provider-linode/pull/96. By default, the addon is not installed when the management cluster is created from scratch. This PR installs the required addon when the management cluster is provisioned using tilt.
Screenshot 2024-02-13 at 12 27 36 AM


AshleyDumaine (approved):
LGTM, thanks!

Issue #92, opened 2024-02-08
Cache Linode API calls


mhmxs:
Would be nice to introduce a cache layer in front of the Linode API client, to avoid unnecessary requests.

mhmxs:
We don't need this at the moment, because we fixed a bug which caused high load on the Linode API.

PR #113, opened 2024-02-13
Cache Linode API GET calls

mhmxs:
This change introduces a caching layer in front of the Linode client.

Fixes: https://github.com/linode/cluster-api-provider-linode/issues/92


cloud/scope/common.go
```diff
@@ -1,21 +1,94 @@
...
 import (
+	"fmt"
 	"net/http"
+	"sync"
 
+	"github.com/bxcodec/gotcha"
+	"github.com/bxcodec/gotcha/cache"
+	"github.com/bxcodec/httpcache"
+	"github.com/bxcodec/httpcache/cache/inmem"
+	"github.com/linode/cluster-api-provider-linode/util"
```
AshleyDumaine, line 12: Should we have the defaults in the util package if they are only being used here?

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.21
 toolchain go1.21.5
 
 require (
+	github.com/bxcodec/httpcache v1.0.0-beta.3
```
eljohnson92, line 8: I'm not sure if this is a concern or not, but just noticed this package(and assuming the gotcha package as well) have been on the same beta version for over 3 years at this point with no updates, are we ok with adopting these dependencies? if we want to accept them I could buy into that, just want to make sure we are aware of them

srust: Can we use ETag caching here?

mhmxs: @eljohnson92 Other libraries (what i found) are archived, so i found this one a better option than writing our own.
I'm open to use something else, do you have any preference?

mhmxs: @srust Maybe i'm wrong, but Linode API doesn't send ETag header, at least the for web ui. Could you please clarify your suggestion?

eljohnson92: @mhmxs as an alternative to this, could we perhaps use a more general purpose go caching library to store cache instead of the purpose-built http caching library? I would be open to using this if we thing it would be the best option, just want to explore some alternatives as well

mhmxs: @eljohnson92 This was the simplest to implement. The alternatives i found:
- stores `string` or `[]byte`, so we have to serialize data
- stores `any` or `interface{}`, so we have to type enforce type after every get
- needs an extra layer between business logic and Llinode client
- already archived
- not designed for high concurrency
- doesn't support TTL

So i'm glad to use another library, if you have better option. This library was the best result of my researching :)

cloud/scope/common.go
```diff
@@ -1,21 +1,109 @@
...
+	// DefaultLinodeAPICacheExpiration is the default cache expiration for Linode API calls.
+	DefaultLinodeAPICacheExpiration = 5 * time.Second
+	// DefaultLinodeAPICacheMaxSizeItem is the default cache max size item for Linode API calls.
+	DefaultLinodeAPICacheMaxSizeItem = 500
+	// DefaultLinodeAPICacheMaxMemory is the default cache max memory for Linode API calls.
+	DefaultLinodeAPICacheMaxMemory = cache.MB
+)
+
+var (
+	initClient   sync.Once
```
cbzzz, line 33: This change seems to limit us to using one API client (and thus one access token) for all CAPL-managed clusters. Is there another way to have both caching for the API while supporting something like per-cluster Linode tokens?

mhmxs: @cbzzz we had one access token before this change, and unnecessary client creations (access token never changes). So i think my version is better than the previous one (more configurable, and less footprint). But your point is also valid, it needs a new concept how to define per-cluster Linode token. Once we change to per-cluster, this cache solution won't be the best for us. So this is an interesting question.

go.mod
srust: Bummer about ETag - oh well.

mhmxs:
The upcoming change invalidates this one: https://github.com/linode/cluster-api-provider-linode/pull/142

PR #115, opened 2024-02-13
[docs] - Add MdBook for documentation and GHA workflow

AshleyDumaine:
This adds [mdBook](https://github.com/rust-lang/mdBook) documentation and a new GHA workflow to build and publish the book to GitHub Pages. 

Also resolves #111

PR #116, opened 2024-02-14
[feat] add HelmChartProxies for linode-ccm and linode-csi-driver, install ccm and cilium for default template

### LINKED ISSUE #100: Add-on for linode-blockstorage-csi-driver
We should make it easy for clusters to opt in to having the [blockstorage CSI driver](https://github.com/linode/linode-blockstorage-csi-driver) installed via a label on the Cluster resource so stateful workloads can be created and backed by Linode block storage. 

This should be doable via the Helm chart plus CAAPH introduced in #96

### LINKED ISSUE #101: Add-on for linode-cloud-controller-manager

AshleyDumaine:
This makes it so CAAPH can install the linode-csi and linode-ccm charts and installs cilium and the linode-ccm by default since they are critical components for the cluster to work properly. CSI can still be installed via the addon template.

The linode-CCM is a necessary component for provisioned [k8s 1.29](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.29.md#api-change-1) clusters to correctly set the internal and external IPs.

Closes #101
Closes #100

Depends on https://github.com/linode/linode-cloud-controller-manager/pull/177


templates/addons/linode-blockstorage-csi-driver-helm.yaml
```diff
@@ -0,0 +1,22 @@
...
+    matchLabels:
+      csi-driver: linode
+  repoURL: https://linode.github.io/linode-blockstorage-csi-driver/
+  chartName: linode-blockstorage-csi-driver
+  namespace: kube-system
+  options:
+    waitForJobs: true
+    wait: true
+    timeout: 5m
+  valuesTemplate: |
```
eljohnson92, line 16: could we include this secret as part of the template?

Tiltfile
```diff
@@ -32,6 +32,28 @@ k8s_resource(
...
+)
+
+# Linode Blockstorage CSI driver for storage on self-managed k8s clusters
+k8s_yaml("./templates/addons/linode-blockstorage-csi-driver-helm.yaml")
+k8s_resource(
+    new_name="addon-linode-blockstorage-csi-driver-helm",
+    objects=[
+        "linode-blockstorage-csi-driver:helmchartproxy"
+    ],
+    resource_deps=["capi-controller-manager", "cluster-api-provider-linode-controller-manager", "caaph-controller-manager"]
```
eljohnson92, line 53: I don't think the CAPL is a dependency for these two addons, only capi and caaph

templates/addons/linode-blockstorage-csi-driver-helm.yaml
AshleyDumaine: Having the Secret in the template would apply it to the management cluster, not the workload cluster. I opted to have the Secret creation be a manual step once the workload cluster is provisioned and reachable, mostly because of the API token requirement (region we could technically determine via `{{ .InfraCluster.spec.region }}` but this isn't anywhere in the spec and probably shouldn't be anyway). (https://github.com/linode/cluster-api-provider-linode/pull/115/files#r1490197122)

AshleyDumaine: I was able to get this working with the experimental/alpha [ClusterResourceSet](https://cluster-api.sigs.k8s.io/tasks/experimental-features/cluster-resource-set).

templates/addons/linode-blockstorage-csi-driver-helm.yaml
```diff
@@ -0,0 +1,22 @@
...
+    secretRef:
+      name: "linode-token-region"
+      apiTokenRef: "apiToken"
+      regionRef: "region"
```
eljohnson92, line 20: does the region have to be in the secret or can we keep just the token in the secret?

templates/addons/linode-cloud-controller-manager-helm.yaml
```diff
@@ -0,0 +1,39 @@
...
+    waitForJobs: true
+    wait: true
+    timeout: 5m
+  valuesTemplate: |
+    secretRef:
+      name: "linode-token-region"
+      apiTokenRef: "apiToken"
+      regionRef: "region"
+    image:
+      tag: v0.3.24
```
eljohnson92, line 22: can we move this version into the version field of the chart? each version should set the image tag correctly

templates/addons/linode-cloud-controller-manager-helm.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: addons.cluster.x-k8s.io/v1alpha1
```
eljohnson92, line 1: considering this is needed for base cluster functionality, should we just include this in the cluster-template.yaml(along with the cilium config?)

templates/addons/linode-blockstorage-csi-driver-helm.yaml
```diff
@@ -0,0 +1,22 @@
...
+    csiLinodePlugin:
+      tag: v0.6.3
```
eljohnson92, line 22: move this to the chart version as well

templates/addons/linode-cloud-controller-manager-helm.yaml
AshleyDumaine: I was thinking about that earlier, sounds good.

templates/addons/linode-blockstorage-csi-driver-helm.yaml
AshleyDumaine: good question, let me check

AshleyDumaine: yep, we can remove these two lines

eljohnson92 (approved):

PR #117, opened 2024-02-14
add version to controller and include it in user-agent for linode api requests

eljohnson92:
This PR tries to auto-detect the controller version based on a tag(with a -dev suffix if it is between tags) and then uses `CAPL/{version}` in the request to the linode API


cmd/main.go
```diff
@@ -19,6 +19,8 @@ package main
 import (
 	"errors"
 	"flag"
+	"fmt"
+	"github.com/linode/cluster-api-provider-linode/version"
```
AshleyDumaine, line 23: nit: import ordering

luthermonson (approved):

cloud/scope/common.go
```diff
@@ -17,5 +19,7 @@ func createLinodeClient(apiKey string) *linodego.Client {
 	}
 	linodeClient := linodego.NewClient(oauth2Client)
 
+	linodeClient.SetUserAgent(fmt.Sprintf("CAPL/%s", version.GetVersion()))
```
luthermonson, line 22: 🔥

avestuk (approved):

PR #124, opened 2024-02-20
shorten cluster-api-provider-linode to capl, add release process

AshleyDumaine:
## Changes
- This shortens "cluster-api-provider-linode" to "capl" on k8s resources to follow other provider naming conventions. 
- Organizes the Makefile and removes some unused targets
- Includes kustomize tweaks so that we can produce infrastructure-components.yaml without templating in values. 

## Additions
- Adds a new release target and GHA workflow based off of https://github.com/linode/linode-cloud-controller-manager/blob/main/.github/workflows/release.yml to build and release the manager image to dockerhub on **tags** (needs extra secrets set in the repo settings to push)

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -1,34 +1,42 @@
 load("ext://k8s_attach", "k8s_attach")
 
-docker_build("controller", ".", only=("Dockerfile", "Makefile", "vendor","go.mod", "go.sum", "./api", "./cloud","./cmd", "./controller", "./util", "./version"), build_args={'VERSION': os.getenv("VERSION","")})
+docker_build(
+    "docker.io/linode/cluster-api-provider-linode",
+    context = ".",
+    only=("Dockerfile", "Makefile", "vendor","go.mod", "go.sum", "./api", "./cloud","./cmd", "./controller", "./util", "./version"),
+    build_args={'VERSION': os.getenv("VERSION","")},
+)
 
 local_resource(
     'capi-controller-manager',
     cmd='EXP_CLUSTER_RESOURCE_SET=true clusterctl init --addon helm',
 )
 
-k8s_yaml(kustomize('config/default'))
-
-# get generated secret name so we can categorize it
-token_secret_name = str(local('kustomize build config/default | grep -m1 "name: cluster-api-provider-linode-token-"', quiet=True, echo_off=True)).split()[1]
+templated_yaml = local(
+    'kustomize build config/default | envsubst',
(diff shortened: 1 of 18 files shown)


Tiltfile
```diff
@@ -1,34 +1,41 @@
 load("ext://k8s_attach", "k8s_attach")
 
-docker_build("controller", ".", only=("Dockerfile", "Makefile", "vendor","go.mod", "go.sum", "./api", "./cloud","./cmd", "./controller", "./util", "./version"), build_args={'VERSION': os.getenv("VERSION","")})
+docker_build(
+    "docker.io/linode/cluster-api-provider-linode",
+    context = ".",
+    only=("Dockerfile", "Makefile", "vendor","go.mod", "go.sum", "./api", "./cloud","./cmd", "./controller", "./util", "./version"),
+    ignore = ["templates"],
```
eljohnson92, line 7: my understanding of `only()` was it would ignore other dirs, are you seeing different behaviour requiring the ignoring of templates?

config/default/kustomization.yaml
```diff
@@ -26,124 +21,7 @@ resources:
...
-
-patchesStrategicMerge:
-# Protect the /metrics endpoint by putting it behind auth.
-# If you want your controller-manager to expose the /metrics
-# endpoint w/o any authn/z, please comment the following line.
-- manager_auth_proxy_patch.yaml
-
-
-
-# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
```
eljohnson92, line 42: should this webhook/certmanager stuff stay in here in case we want to use it in the future?

AshleyDumaine: I'd prefer to add that when we need it but I can add back if that's preferable

Tiltfile
```diff
@@ -1,34 +1,40 @@
...
 )
 
-k8s_yaml(kustomize('config/default'))
+def kustomizesub(dir):
+    yaml = local('hack/kustomize-sub.sh {}'.format(dir), quiet=True)
+    return yaml
 
-# get generated secret name so we can categorize it
-token_secret_name = str(local('kustomize build config/default | grep -m1 "name: cluster-api-provider-linode-token-"', quiet=True, echo_off=True)).split()[1]
+k8s_yaml(kustomizesub('config/default'))
```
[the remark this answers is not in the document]
AshleyDumaine: updated to remove the hack script

eljohnson92 (approved):

PR #130, opened 2024-02-21
[feat, docs] - Add ClusterClass support and create flavors

AshleyDumaine:
## Additions
- Toggles on the ClusterClass feature gate for the KIND cluster created via Tilt
- Adds a new "clusterclass-kubeadm" flavor
- Uses kustomize to generate the flavors
- Adds a new make target and hack script to generate flavors

## Changes
- Reorders and cleans up the development doc in the book

## Deletions
- Removes unused `make run` target
- Removes `templates/cluster-template*.yaml` to reduce confusion (the generated templates via kustomize should not be directly edited) 
  - 💬  Discussion: I'm open to keeping these in here but to do so we'll need more work to verify the templates are always up-to-date and include some kind of comment that the files are autogenerated and not to edit them directly.


templates/addons/cilium/cilium.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: addons.cluster.x-k8s.io/v1alpha1
+kind: HelmChartProxy
+metadata:
+  name: cilium
+spec:
+  clusterSelector:
+    matchLabels:
+      cni: cilium
+  repoURL: https://helm.cilium.io/
```
eljohnson92, line 9: something I realized we probably want to do here is to install cilium in a NS other than `default`

AshleyDumaine: good catch, updated to kube-system

Makefile
```diff
@@ -12,7 +11,7 @@ ARCH_SHORT := amd64
 else ifeq ($(ARCH_SHORT),aarch64)
 ARCH_SHORT := arm64
 endif
-VERSION ?= $(shell git describe --tags --dirty=-dev)
+VERSION ?= $(shell git describe --always --dirty=-dev)
```
eljohnson92, line 14: we need both `tag` and `always` here I think

hack/generate-flavors.sh
```diff
@@ -0,0 +1,13 @@
...
+
+REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+FLAVORS_DIR="${REPO_ROOT}/templates/flavors"
+
+for name in $(find "${FLAVORS_DIR}/"* -maxdepth 0 -type d -print0 | xargs -0 -I {} basename {} | grep -v base); do
+  kustomize build "${FLAVORS_DIR}/${name}" > "${REPO_ROOT}/templates/cluster-template-${name}.yaml"
+done
+
+# move the default template to the default file expected by clusterctl
+mv "${REPO_ROOT}/templates/cluster-template-default.yaml" "${REPO_ROOT}/templates/cluster-template.yaml"
```
eljohnson92, line 13: do we need both template-default and template.yaml to exist together or is just the tamplate.yaml enough?

Makefile
```diff
@@ -190,20 +189,24 @@ tilt-cluster: ctlptl tilt kind clusterctl
...
 set-manifest-image: ## Update kustomize image patch file for default resource.
 	sed -i'' -e 's@image: .*@image: '"${MANIFEST_IMG}:${MANIFEST_TAG}"'@' ./config/default/manager_image_patch.yaml
 
 .PHONY: release
-release: $(KUSTOMIZE)
-	rm -rf $(RELEASE_DIR)
+release: $(KUSTOMIZE) clean-release
 	mkdir -p $(RELEASE_DIR)/
-	$(MAKE) set-manifest-image MANIFEST_IMG=$(REGISTRY)/$(IMAGE_NAME) MANIFEST_TAG=$(RELEASE_TAG)
+	$(MAKE) set-manifest-image MANIFEST_IMG=$(REGISTRY)/$(IMAGE_NAME) MANIFEST_TAG=$(VERSION)
```
eljohnson92, line 200: could these `$(MAKE)` calls be refactored to be dependent targets instead?

hack/generate-flavors.sh
AshleyDumaine: just the cluster-template is expected for the default flavor (omitting the flavor flag)

templates/flavors/clusterclass/cluster.yaml
eljohnson92, line 1: should we break this out into it's own flavor/template? maybe something like `kubeadm-cluster-class` and `kubeadm-cluster` or something like that?

eljohnson92 (approved):

PR #119, opened 2024-02-16
chore: migrate from kuttl to chainsaw

eddycharly:
Chainsaw was created because kuttl stopped to be actively maintained (it's actually a single maintainer now).

We are also solving most of the kuttl limitations that exist today and wrote a command to automatically migrate from kuttl to chainsaw. It supercharges e2e tests with better logs, config maps assertions, [assertions trees](https://kyverno.io/blog/2023/12/13/kyverno-chainsaw-exploring-the-power-of-assertion-trees/) and many more things 💪

This is a relatively young project but it has already been adopted by a couple of e2e tests heavy projects (kyverno, keptn, fairwinds rbac-manager, grafana-operator, redis-operator and others).

To give more context, you can see similar PRs at:
- https://github.com/keptn/lifecycle-toolkit/pull/3000
- https://github.com/open-telemetry/opentelemetry-operator/pull/2630
- https://github.com/operator-framework/operator-sdk/pull/6676
- https://github.com/grafana/tempo-operator/pull/786 in this PR the tempo-operator maintainers asked some good questions and I explained the current status, that I suggest taking a look at.

Sadly, [kuttl](https://github.com/kudobuilder/kuttl) haven't done a release since jan 2023.

Here you can find a good video showcasing chainsaw: https://www.youtube.com/watch?v=Ejof-wtAdQM&ab_channel=CamilaMacedo

If you need more infos :point_down:
- GitHub: https://github.com/kyverno/chainsaw
- Docs: https://kyverno.github.io/chainsaw
- Slack: https://kubernetes.slack.com/archives/C067LUFL43U

PR #131, opened 2024-02-21
e2e: Replace Kuttl-in-Kuttl

### LINKED ISSUE #103: e2e: Replace Kuttl-in-Kuttl with simple kubectl apply in scripts
The Kuttl-in-Kuttl pattern introduced in https://github.com/linode/cluster-api-provider-linode/pull/80 can be replaced with defining a script that simply runs `kubectl apply` while passing dynamic data. The `TestStep` will fail if any of the commands exit with non-zero.

```sh
MACHINE_UID="$(OBJ=machines/machine-sample make getKubeUid)"

cat <<EOF | kubectl apply -n $NAMESPACE -f -
kind: LinodeMachine
metadata:
  ownerReferences:
  - apiVersion: cluster.x-k8s.io/v1beta1
    kind: Machine
    name: machine-sample
    uid: ${MACHINE_UID}
EOF
```

bcmendoza820:
Closes #103

This PR replaces kuttl-in-kuttl and removes related helper scripts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,12 +117,12 @@ test: manifests generate fmt vet envtest ## Run tests.
 e2etest:
 	make --no-print-directory _e2etest # Workaround to force the flag on Github Action
 
-_e2etest-infra: kind ctlptl tilt kuttl kustomize clusterctl
+_e2etest-infra: kind ctlptl tilt kuttl kustomize clusterctl envsubst
 	@echo -n "LINODE_TOKEN=$(LINODE_TOKEN)" > config/default/.env.linode
 	$(CTLPTL) apply -f .tilt/ctlptl-config.yaml
 	$(TILT) ci --timeout 240s -f Tiltfile
 
-_e2etest: manifests generate _e2etest-infra
+_e2etest: manifests generate envsubst _e2etest-infra
 	ROOT_DIR="$(PWD)" $(KUTTL) test --config e2e/kuttl-config.yaml
 
 ## --------------------------------------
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -324,12 +322,6 @@ envtest: $(ENVTEST) ## Download setup-envtest locally if necessary.
 $(ENVTEST): $(LOCALBIN)
 	GOBIN=$(CACHE_BIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
-.PHONY: envsubst
```
AshleyDumaine, line 327: We should keep envsubst, we're using that in the Tiltfile: https://github.com/linode/cluster-api-provider-linode/blob/main/Tiltfile#L16

AshleyDumaine (changes requested):

Makefile
eljohnson92: @AshleyDumaine  if we only need envsubst for the Tiltfile we should absolutely rip it out fo there as well. is there anywhere else we are using envsubst explicitly?

AshleyDumaine (approved):

eddycharly:
Folks did you look at my PR #119 ?

We can discuss the pros/cons if you want, there's a lot of things to win switching to chainsaw IMHO.

eljohnson92:
Hi @eddycharly we wanted to first remove the kuttl-in-kuttl pattern before considering switching tools so we can compare with a simpler use-case.

eddycharly:
@eljohnson92 👍

PR #138, opened 2024-02-26
move kubeadm specific config into default

eljohnson92:
**What type of PR is this?**
/kind cleanup

This PR splits base and default based on the usage of kubeadm. this will allow none-kubeadm resources to import `base/` directly
Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


AshleyDumaine (approved):

templates/flavors/base/cluster.yaml
```diff
@@ -1,19 +1,18 @@
+---
 apiVersion: cluster.x-k8s.io/v1beta1
 kind: Cluster
 metadata:
   name: ${CLUSTER_NAME}
   labels:
-    cni: cilium
```
AshleyDumaine, line 6: I think we should leave these on here since they're currently needed for the clusters to work

AshleyDumaine (approved):

PR #140, opened 2024-02-26
[feat] - use etcd volume on control plane nodes

AshleyDumaine:
**What type of PR is this?**

/kind feature

This PR configures etcd to be on a separate volume from the root disk for control plane nodes. The volume ID for each control plane node is stored in the status field on the LinodeMachine to save some API calls. A 10 GB volume is used since that is the smallest volume possible and is still well within the recommendations of the [etcd docs](https://etcd.io/docs/v3.5/dev-guide/limit/).
  💬 Discussion: should we make this configurable in the LinodeCluster spec ? 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


srust (changes requested):

cloud/services/volumes.go
```diff
@@ -0,0 +1,88 @@
...
+func CreateEtcdVolume(ctx context.Context,
+	linodeID int,
+	logger logr.Logger,
+	machineScope *scope.MachineScope,
+) (volume *linodego.Volume, err error) {
+	// if the instance is part of the control plane, create an etcd volume to attach
+	if !kutil.IsControlPlaneMachine(machineScope.Machine) {
+		return volume, err
+	}
+	volume, err = machineScope.LinodeClient.CreateVolume(ctx, linodego.VolumeCreateOptions{
```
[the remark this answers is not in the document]
AshleyDumaine: I redid this PR to change to split the root disk in half and use the other half for the etcd disk

controller/linodemachine_controller.go
```diff
@@ -378,6 +388,67 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
+	// carve out half the root disk space to be used for the etcd disk
+	rootDiskID := instanceConfig.Devices.SDA.DiskID
+	rootDisk, err := machineScope.LinodeClient.GetInstanceDisk(ctx, linodeInstanceID, rootDiskID)
+	if err != nil {
+		logger.Error(err, "Failed to get root disk for instance")
+
+		return err
+	}
+	//nolint:gomnd // cutting disk in half
+	diskSize := rootDisk.Size / 2
```
cbzzz, line 415: Should this be bounded? At least to the recommended limit (8GB)? https://etcd.io/docs/v3.3/dev-guide/limit/#storage-size-limit

eljohnson92: could we perhaps set the rootdisk to disksize-8GB and the etcd-data to 8gb?

cbzzz: Ah, it looks like we're already setting the quota in ETCD to 8GB here: https://github.com/linode/cluster-api-provider-linode/blob/b33f0355d1666bc4ba5b29ab9c32455988ddb8b9/templates/flavors/clusterclass-kubeadm/kubeadm-controlplane-template.yaml#L37-L38

AshleyDumaine: I've configured this to be 8 GB and updated the etcd doc to note this

controller/linodemachine_controller.go
```diff
@@ -378,6 +391,66 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
+	// carve out space for the etcd disk
+	rootDiskID := instanceConfig.Devices.SDA.DiskID
+	rootDisk, err := machineScope.LinodeClient.GetInstanceDisk(ctx, linodeInstanceID, rootDiskID)
+	if err != nil {
+		logger.Error(err, "Failed to get root disk for instance")
+
+		return err
+	}
+	diskSize := rootDisk.Size - defaultEtcdDiskSize
+	if err = machineScope.LinodeClient.ResizeInstanceDisk(ctx, linodeInstanceID, rootDiskID, diskSize); err != nil {
```
srust, line 418: Why must we carve out of the root disk and not allocate an additional disk? Why bother with the resize?

controller/linodemachine_controller.go
```diff
@@ -378,6 +391,66 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
+		},
+	)
+	if err != nil {
+		logger.Error(err, "Failed to create etcd disk")
+
+		return err
+	}
+
+	// attach etcd disk to sdc
+	instanceConfig.Devices.SDC = &linodego.InstanceConfigDevice{DiskID: etcdDisk.ID}
```
srust, line 441: these should be statitically allocated during config profile right, and never detached?

controller/linodemachine_controller.go
```diff
@@ -317,30 +320,40 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
 			return nil, err
 		}
+
+		// create etcd disk only if this is a control plane node
+		if kutil.IsControlPlaneMachine(machineScope.Machine) {
+			if err = r.configureDisks(ctx, logger, machineScope, linodeInstance.ID); err != nil {
+				logger.Error(err, "Failed to configure instance disks")
+
+				return nil, err
+			}
```
rahulait, line 354: Maybe we should put the boot step in the kutil if loop instead of configureDisks so that we can separate if it failed to configure disk or failed to boot the instance?

controller/linodemachine_controller.go
```diff
@@ -317,30 +320,40 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 		linodeInstance = &linodeInstances[0]
 	case 0:
 		// get the bootstrap data for the Linode instance and set it for create config
-		createConfig, err := r.newCreateConfig(ctx, machineScope, tags, logger)
+		createOpts, err := r.newCreateConfig(ctx, machineScope, tags, logger)
```
srust, line 323: I believe `boot` defaults to true in the API, do we need to set it to `false` explicitly so we can continue configuring it before booting it.

controller/linodemachine_controller.go
```diff
@@ -49,6 +49,9 @@ import (
 	"github.com/linode/cluster-api-provider-linode/cloud/services"
 )
 
+// default etcd Disk size in MB
+const defaultEtcdDiskSize = 8192
```
srust, line 53: I don't think this is enough.

the max etcd data size *on-disk* is 8GiB. However, there is filesystem overhead, etc. so surely the disk itself must be larger than the etcd on-disk size?

controller/linodemachine_controller.go
AshleyDumaine: sure, sounds good

rahulait (approved):
The change looks good to me. I am not sure if we should have extra room for snapshots or anything like that, we'll be using very tiny bit of 8G most of the times, so it might be just fine. I wonder if we should throw an error if someone selects a flavor/plan which has less than 50G of storage, but that can be done in a separate ticket if required. This PR looks good to me.

controller/linodemachine_controller.go
```diff
@@ -317,30 +320,40 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
 
-		linodeInstance, err = machineScope.LinodeClient.CreateInstance(ctx, *createConfig)
+		linodeInstance, err = machineScope.LinodeClient.CreateInstance(ctx, *createOpts)
 		if err != nil {
 			logger.Error(err, "Failed to create Linode machine instance")
 
 			return nil, err
 		}
+
+		// create etcd disk only if this is a control plane node
```
srust, line 348: nit: remove etcd from comment. Refer to it as "control plane disk configuring"

And the r.configureDisks could take a param for control plane, or probably better be configureDisksControlPlane so that we can future proof.

Likely, we are going to want to do different partitions for either types, when we talk about upgrades or immutable OS where we could potentially use different partitions to reboot in the OS, and have separate disk configurations that we can boot.

controller/linodemachine_controller.go
```diff
@@ -378,6 +391,66 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
+		logger.Error(err, "Failed to create etcd disk")
+
+		return err
+	}
+
+	// attach etcd disk to sdc
+	instanceConfig.Devices.SDC = &linodego.InstanceConfigDevice{DiskID: etcdDisk.ID}
+
+	// finally, boot the machine
+	err = machineScope.LinodeClient.BootInstance(ctx, linodeInstanceID, instanceConfig.ID)
```
srust, line 444: can move this out into the caller, so we can disable booting during create, and boot in a common way once the disks are configured here.

controller/linodemachine_controller.go
AshleyDumaine: We have to do this or we hit storage plan limits. I cannot add an extra disk without shrinking the root disk or swap.

controller/linodemachine_controller.go
AshleyDumaine: yes, that was Rahul's suggestion here: https://github.com/linode/cluster-api-provider-linode/pull/140/files#r1506266046

controller/linodemachine_controller.go
AshleyDumaine: see https://github.com/linode/cluster-api-provider-linode/pull/140/files#diff-84b19020b945417b38ec4b095741150823859a32633c49a650471dc6717cd3b8R62

controller/linodemachine_controller.go
AshleyDumaine: What would be your suggestion for a size if not 8 GB? If we allow nanodes for instance type on control plane nodes we have at most 25 GB for all disks attached to the instance (root, swap, and etcd).

srust (approved):
Added one comment about the booted=false and bootinstance making them common cases. I think that makes it clearer, less conditionals.

Otherwise, this looks good to go.

Thanks!

srust:
@AshleyDumaine 

We're going to need the etcd partition size to be configurable as a parameter. Different users are going to want different.

To hold backups, work space for tarring the files, working with snapshots,etc. some users are going to want up to 16 or 32 GiB for a partition.

We can safely have a nice default but allow it configurable.

AshleyDumaine:
> @AshleyDumaine
> 
> We're going to need the etcd partition size to be configurable as a parameter. Different users are going to want different.
> 
> To hold backups, work space for tarring the files, working with snapshots,etc. some users are going to want up to 16 or 32 GiB for a partition.
> 
> We can safely have a nice default but allow it configurable.

Sounds good, I created https://github.com/linode/cluster-api-provider-linode/issues/152 to address this. I think we'll want to include validating admission webhooks for this to reject if the requested etcd size exceeds the instance plan for the control plane nodes. This should be added in another PR.

controller/linodemachine_controller.go
AshleyDumaine: yes, I also checked to make sure this stays in the event of an instance reboot.

eljohnson92 (approved):

PR #142, opened 2024-02-27
feat: multi-tenancy

cbzzz:
**What type of PR is this?**
/kind feature

Adds multi-tenancy for CAPL-managed clusters. Clusters may now reference an optional Secret containing their Linode credentials (i.e. API token) to be used for the deployment of "things" associated with their cluster.

Fixes #

- From the design, it appears `LinodeVPC` and `LinodeCluster` objects are loosely-coupled and managed somewhat independently by the controller. I thought it was best to use a separate `CredentialsRef` field in the Spec for `LinodeVPC`s rather than doing a bigger refactor and trying to couple the two types.

- The `$LINODE_TOKEN` environment variable is _always_ propagated into a Secret that will be created for every workload cluster. While a credentials Secret may be referenced by multiple workload clusters (or not at all), I thought generating a per-cluster Secret was best in the current cluster templating workflow.

**TODOs**:

- [x] squashed commits
- [X] includes documentation
- [ ] adds unit tests

**Testing**:
1. Create two Linode tokens, one with full "Read/Write" permissions (`$GOOD_TOKEN`) and one with full "None" permissions (`$BAD_TOKEN`).

2. Deploy two workload clusters, one per token:
```sh
❯ LINODE_TOKEN=$GOOD_TOKEN clusterctl generate cluster ${CLUSTER_NAME}-good \
  --kubernetes-version v1.29.1 \
  --infrastructure linode:0.0.0 \
  | kubectl apply -f -

❯ LINODE_TOKEN=$BAD_TOKEN clusterctl generate cluster ${CLUSTER_NAME}-bad \
  --kubernetes-version v1.29.1 \
  --infrastructure linode:0.0.0 \
  | kubectl apply -f -
```

3. A new credentials Secret is created for every workload cluster:
```sh
❯ kubectl get secrets
NAME                             TYPE                                   DATA   AGE
cbang-capl-bad-credentials       Opaque                                 1      10m
cbang-capl-good-credentials      Opaque                                 1      10m
```

4. The workload cluster with the valid token will provision successfully while the one with the invalid token will fail:
```sh
❯ kubectl get clusters
NAMESPACE   NAME              CLUSTERCLASS   PHASE         AGE     VERSION
default     cbang-capl-good                  Provisioned   10m
default     cbang-capl-bad                   Failed        10m

❯ kubectl logs -n capl-system capl-controller-manager-67cdf96488-q682z
2024-03-04T14:00:00Z    DEBUG   events  [401] Your OAuth token is not authorized to use this endpoint.  {"type": "Warning", "object": {"kind":"LinodeCluster","namespace":"default","name":"cbang-capl-bad","uid":"9ec42cbe-3214-4bf7-b9be-5eadcdf6d2bd","apiVersion":"infrastructure.cluster.x-k8s.io/v1alpha1","resourceVersion":"38418"}, "reason": "CreateError"}
```
NOTE: These tokens will be different from the default one (`capl-system/capl-manager-credentials`) that is used by CAPL if the Cluster does not reference a credentials Secret.


api/v1alpha1/linodecluster_types.go
```diff
@@ -40,6 +40,11 @@ type LinodeClusterSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	// +optional
 	VPCRef *corev1.ObjectReference `json:"vpcRef,omitempty"`
+
+	// CredentialsRef is a reference to a Secret that contains the credentials to use for provisioning this cluster. If not
+	// supplied then the credentials of the controller will be used.
+	// +optional
+	CredentialsRef *corev1.SecretReference `json:"credentialsRef,omitempty"`
```
bcm820, line 47: Since `LinodeCluster` and others are namespace-scoped, should the type here just be a string and expect the secret exist in the same namespace?

cloud/scope/common.go
```diff
@@ -23,3 +28,22 @@ func createLinodeClient(apiKey string) *linodego.Client {
...
+		Name:      credentialsRef.Name,
+		Namespace: credentialsRef.Namespace,
+	}
+
+	credSecret := &corev1.Secret{}
+	if err := crClient.Get(ctx, secretRefName, credSecret); err != nil {
+		return nil, fmt.Errorf("getting credentials secret %s\\%s: %w", secretRefName.Namespace, secretRefName.Name, err)
+	}
+
+	rawData, ok := credSecret.Data["apiToken"]
```
bcm820, line 43: Do we want the key to be configurable? For flexibility, we can use `corev1.SecretKeySelector` which allows for specifying a secret `name` and `key` (this assumes we expect the Secret to exist in the same namespace).

api/v1alpha1/linodecluster_types.go
cbzzz: I think it could, but I wanted to leave the option open to use Secrets from a different namespace (like the GCP version). I'll defer to the team if we want to strictly enforce same-namespace Secrets.

I was planning to have the manifest templating default the `SecretReference.Namespace` value to where the cluster is installed though.

cloud/scope/common.go
cbzzz: I think it could, but I wanted to keep the key consistent with the key for the "global" API token we use throughout controller right now.

I can make it configurable if we want to start generalizing out this logic. I was thinking about removing `LINODE_TOKEN` as an environment variable completely (in a separate PR 😅) and just having the reconciller(s) reference Secrets directly.

bcm820: Understood! If there's precedence for using `apiToken` I don't see why we shouldn't stick with it and document it. And +1 on removing the environment variable as well.

api/v1alpha1/linodecluster_types.go
bcm820: Now that you mention it, I could see the use-case of referencing the same secret in some other namespace and not needing to propagate it everywhere. We'll adopt this convention as well in our ticket.

cloud/scope/common.go
cbzzz: That's fair! I can add the `apiToken` key to the documentation 🫡

AshleyDumaine (approved):
Can we add a docs update on multi-tenancy under the topics section? 🙏

cbzzz:
> Can we add a docs update on multi-tenancy under the topics section? 🙏

Can do! That one in-line comment wasn't doing it for you? 🙈

cloud/scope/common.go
```diff
@@ -23,3 +28,26 @@ func createLinodeClient(apiKey string) *linodego.Client {
...
+		Name:      credentialsRef.Name,
+		Namespace: credentialsRef.Namespace,
+	}
+	if secretRef.Namespace == "" {
+		secretRef.Namespace = defaultNamespace
+	}
+
+	credSecret := &corev1.Secret{}
+	if err := crClient.Get(ctx, secretRef, credSecret); err != nil {
+		return nil, fmt.Errorf("get credentials secret %s\\%s: %w", secretRef.Namespace, secretRef.Name, err)
```
AshleyDumaine, line 43:
```suggestion
		return nil, fmt.Errorf("get credentials secret %s/s: %w", secretRef.Namespace, secretRef.Name, err)
```

config/crd/bases/infrastructure.cluster.x-k8s.io_linodevpcs.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.14.0
+    controller-gen.kubebuilder.io/version: v0.13.0
```
AshleyDumaine, line 6: might need to update devbox dependencies

cbzzz: Isn't this a Go dependency?

eljohnson92: if you pull in the latest from `main` and regenerate there is a fix in the Makefile. basically devbox doesn't have the latest controller-tools version

docs/src/topics/multi-tenancy.md
```diff
@@ -0,0 +1,37 @@
+# Multi-Tenancy
+
+CAPL can manage multi-tenant workload clusters across Linode accounts.
+LinodeCluster objects may reference an optional Secret containing their Linode
+credentials (i.e. API token) to be used for the deployment of "things"
```
AshleyDumaine, line 5: I would slightly change the wording to note that this is for Linode resources associated with individual workload clusters such as Linodes, VPCs, and NodeBalancers.

cloud/scope/vpc.go
```diff
@@ -54,11 +54,19 @@ func validateVPCScopeParams(params VPCScopeParams) error {
...
 // NewVPCScope creates a new Scope from the supplied parameters.
 // This is meant to be called for each reconcile iteration.
-func NewVPCScope(apiKey string, params VPCScopeParams) (*VPCScope, error) {
+func NewVPCScope(ctx context.Context, apiKey string, params VPCScopeParams) (*VPCScope, error) {
 	if err := validateVPCScopeParams(params); err != nil {
 		return nil, err
 	}
 
+	// Override the controller credentials with ones from the VPC's Secret reference (if supplied).
+	if params.LinodeVPC.Spec.CredentialsRef != nil {
```
eljohnson92, line 63: is `LinodeCluster` in the scope of this as well? I think it would be nice if it could just be set at the linodeCluster level and be inherited by the other resources under it.  we can still keep this part in, but just default to the cluster if this isn't specified possibly?

cbzzz: Please see the Reviewer Notes in the Description. I'm happy to discuss if we want to take on a bigger refactor of our reconciliation logic though.

config/crd/bases/infrastructure.cluster.x-k8s.io_linodevpcs.yaml
cbzzz: Do you mean regenerate with `make manifests` (and controller-gen `v0.14.0`) or just run the controller-gen (`v1.13.0`) binary provided by Devbox?

docs/src/topics/multi-tenancy.md
cbzzz: Better? Worse? 🤔 https://github.com/linode/cluster-api-provider-linode/pull/142/commits/9942590798b51169f5fcd60fb7d223f02e5390c1

config/crd/bases/infrastructure.cluster.x-k8s.io_linodevpcs.yaml
eljohnson92: if you rerun `make manifests` or `make build` from main it *should* upgrade things back to 0.14.0

cloud/scope/vpc.go
eljohnson92: ah right, thanks for pointing me back at that I had read it and forgot while reviewing. happy to chat through this but given we are creating VPCs for lkeClusters I think being able to just define it there if possible would be preferable as the default

config/crd/bases/infrastructure.cluster.x-k8s.io_linodevpcs.yaml
cbzzz: I should be already using controller-gen `v0.14.0` for these specific changes 🤔

eljohnson92: in your latest commit I still see the `v0.13.0` label on the resources

cloud/scope/vpc.go
cbzzz: Sounds good! Since the ObjectStorageBucket reconciler (#154) also uses this pattern, I think we might want this be a bigger conversation about redesigning how the various controllers work together (and what information controllers should wait for/need when creating the backing Linode resource).

config/crd/bases/infrastructure.cluster.x-k8s.io_linodevpcs.yaml
cbzzz: No, I think that's the suggested change?

eljohnson92 (approved):

AshleyDumaine (approved):
LGTM

PR #145, opened 2024-02-28
[feat] Add deleteBucket logic for Linode OBJ

amold1:
**What type of PR is this?**
/kind feature

This PR adds the delete bucket logic for the Linode OBJ

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


AshleyDumaine:
Seems the `api/va1alpha1/linodeobjectstoragebucket_types.go` file is missing. Did you create the new CRD via kubebuilder?

bcm820:
> Seems the `api/va1alpha1/linodeobjectstoragebucket_types.go` file is missing. Did you create the new CRD via kubebuilder?

This is @amold1's merge into my feature branch. I'll make a draft PR for it soon!

cloud/scope/object_storage_bucket.go
```diff
@@ -90,6 +90,15 @@ func (s *ObjectStorageBucketScope) AddFinalizer(ctx context.Context) error {
...
 
+// RemoveFinalizer removes a finalizer immediately patches the
+// object to avoid any race conditions.
+func (s *ObjectStorageBucketScope) RemoveFinalizer(ctx context.Context) error {
+	if controllerutil.RemoveFinalizer(s.Object, infrav1alpha1.GroupVersion.String()) {
+		return s.Close(ctx)
+	}
+	return nil
+}
+
```
bcm820, line 101: I see that the other controllers are calling `controllerutil.RemoveFinalizer` directly. IMO we should be somewhat consistent. I wouldn't mind having all scopes implement `RemoveFinalizer` since they all have `AddFinalizer`. We could also make a separate helper for it and add it as an inline struct, or maybe require that all scopes implement some "`FinalizerOps`" interface. What are your thoughts?

controller/linodeobjectstoragebucket_controller.go
```diff
@@ -191,11 +204,47 @@ func (r *LinodeObjectStorageBucketReconciler) reconcileCreate(ctx context.Contex
...
+	if err := bucketScope.RemoveFinalizer(ctx); err != nil {
+		return nil
+	}
+
+	return nil
+}
+
+// bucketExists indicates whether or not an Object Storage Bucket exists.
+// It uses the Linode API.
+func (r *LinodeObjectStorageBucketReconciler) bucketExists(ctx context.Context, logger logr.Logger, bucketScope *scope.ObjectStorageBucketScope) (bool, error) {
```
bcm820, line 237: I think this function should be in the `cloud/services` package since all Linode client code lives there.

controller/linodeobjectstoragebucket_controller.go
```diff
@@ -191,11 +204,47 @@ func (r *LinodeObjectStorageBucketReconciler) reconcileCreate(ctx context.Contex
...
+	if bucketScope.Object == nil {
+		return false, errors.New("nil bucket")
+	}
```
bcm820, line 240: This check isn't needed since `NewObjectStorageBucketScope` already validates params.

controller/linodeobjectstoragebucket_controller.go
amold1: Agreed. Moved it

controller/linodeobjectstoragebucket_controller.go
amold1: Agreed. Removed

cloud/services/object_storage_buckets.go
```diff
@@ -114,3 +114,18 @@ func CreateObjectStorageKeys(ctx context.Context, bucketScope *scope.ObjectStora
...
 	return newKeys, nil
 }
+
+func DeleteObjectStorageKeys(ctx context.Context, bucketScope *scope.ObjectStorageBucketScope, logger logr.Logger, keys [2]float64) error {
+
+	for _, keyID := range keys {
+		logger.Info("deleting object storage access key %s", keyID)
+		if err := bucketScope.LinodeClient.DeleteObjectStorageKey(ctx, int(keyID)); err != nil {
+			logger.Info("access key does not exist")
+		} else if err != nil {
```
bcm820, line 124: This is a duplicate error check. Did you mean for the check above to eval if the error is a 404?

cloud/scope/object_storage_bucket.go
```diff
@@ -161,6 +162,50 @@ func (s *ObjectStorageBucketScope) ApplyAccessKeySecret(ctx context.Context, key
...
+		{"read_only", "ro"},
+	} {
+		secretDataForKey, isset := secret.Data[e.permission]
+		if !isset {
+			return newKeys, fmt.Errorf("secret %s missing data field: %s", secretName, e.permission)
+		}
+		decodedSecretDataForKey := string(secretDataForKey)
+
+		var jsonMap map[string]interface{}
+		json.Unmarshal([]byte(string(decodedSecretDataForKey)), &jsonMap)
```
bcm820, line 194: This needs to be error checked

cloud/scope/object_storage_bucket.go
```diff
@@ -161,6 +162,50 @@ func (s *ObjectStorageBucketScope) ApplyAccessKeySecret(ctx context.Context, key
...
+	var secret corev1.Secret
+	if err := s.client.Get(ctx, objkey, &secret); err != nil {
+		if apierrors.IsNotFound(err); err != nil {
+			return newKeys, err
+		}
+	}
+	for i, e := range []struct {
+		permission string
+		suffix     string
+	}{
```
bcm820, line 183: Since we only need `permission`, this can just loop over `[]string` versus `[]struct{}

cloud/scope/object_storage_bucket.go
```diff
@@ -161,6 +162,50 @@ func (s *ObjectStorageBucketScope) ApplyAccessKeySecret(ctx context.Context, key
...
+
+		var jsonMap map[string]interface{}
+		json.Unmarshal([]byte(string(decodedSecretDataForKey)), &jsonMap)
+
+		accessKeyID, ok := jsonMap["id"]
+		if !ok {
+			err := fmt.Errorf("obj bucket key not found")
+			return newKeys, err
+		}
+		key := accessKeyID.(float64)
```
bcm820, line 201: Isn't the `id` field just an `int`? We can unmarshal into a `linodego.ObjectStorageKey` to not have to deal with `any` types.

PR #146, opened 2024-02-28
[fix] - add token field to CRS secret for csi-driver chart

AshleyDumaine:
**What type of PR is this?**

/kind bug

The linode-blockstorage-csi-driver expects for the secretRef that the token field is `token`, not `apiToken` like in the linode-ccm:
- https://github.com/linode/linode-blockstorage-csi-driver/blob/main/helm-chart/csi-driver/templates/csi-secrets.yaml#L8
- https://github.com/linode/linode-cloud-controller-manager/blob/main/deploy/chart/templates/ccm-linode.yaml#L8

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/templates/addons/linode-blockstorage-csi-driver/linode-bs-csi.yaml
+++ b/templates/addons/linode-blockstorage-csi-driver/linode-bs-csi.yaml
@@ -17,3 +17,4 @@ spec:
   valuesTemplate: |
     secretRef:
       name: "linode-token-region"
+      apiTokenRef: "apiToken"


cbzzz (approved):

templates/addons/cluster-resource-set/secret.yaml
```diff
@@ -12,6 +12,7 @@ stringData:
       namespace: kube-system
     stringData:
       apiToken: ${LINODE_TOKEN}
+      token: ${LINODE_TOKEN}
```
eljohnson92, line 15: Could we instead update the values on the CSI chart to use apiToken instead of token?

AshleyDumaine: updated

eljohnson92 (approved):

PR #156, opened 2024-02-29
[fix] [rke2] - set node-ip, use internal-only-ips for rke2 registration, set the ip type properly for the machine addresses

AshleyDumaine:
**What type of PR is this?**

/kind bug

Adds the private IP to the `node-ip` and the public IP to the `tls-san` config for RKE2 like we had to do for K3s so the TLS certs are valid. Without `node-ip`, the container logs can't be retrieved and without the `tls-san` addition on top of that, server joining stops working. Ideally we need to have registration done via [controlPlaneEndpoint registration though that's still in progress](https://github.com/rancher-sandbox/cluster-api-provider-rke2/pull/266).
I'm aware `hostname -I` is brittle with the ordering per the manpage, but I don't have a much better solution at this time for this workaround.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/templates/flavors/rke2/rke2ConfigTemplate.yaml
+++ b/templates/flavors/rke2/rke2ConfigTemplate.yaml
@@ -12,7 +12,11 @@ spec:
         kubelet:
           extraArgs:
             - "provider-id=linode://{{ ds.meta_data.id }}"
+      # TODO: use MDS to get public and private IP instead because hostname ordering can't always be assumed
       preRKE2Commands:
+        - |
+          mkdir -p /etc/rancher/rke2/config.yaml.d/
+          echo "node-ip: $(hostname -I | grep -oE 192\.168\.[0-9]+\.[0-9]+)" >> /etc/rancher/rke2/config.yaml.d/capi-config.yaml
         - sed -i '/swap/d' /etc/fstab
         - swapoff -a
         - hostnamectl set-hostname '{{ ds.meta_data.label }}' && hostname -F /etc/hostname
--- a/controller/linodemachine_controller.go
+++ b/controller/linodemachine_controller.go
@@ -378,10 +378,14 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 	machineScope.LinodeMachine.Spec.ProviderID = util.Pointer(fmt.Sprintf("linode://%d", linodeInstance.ID))
 
 	machineScope.LinodeMachine.Status.Addresses = []clusterv1.MachineAddress{}
-	for _, add := range linodeInstance.IPv4 {
+	for _, addr := range linodeInstance.IPv4 {
+		addrType := clusterv1.MachineExternalIP
+		if addr.IsPrivate() {
+			addrType = clusterv1.MachineInternalIP
+		}
 		machineScope.LinodeMachine.Status.Addresses = append(machineScope.LinodeMachine.Status.Addresses, clusterv1.MachineAddress{
-			Type:    clusterv1.MachineExternalIP,
-			Address: add.String(),
+			Type:    addrType,
+			Address: addr.String(),
 		})
 	}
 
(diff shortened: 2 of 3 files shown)


rahulait (approved):
tested this, works fine. LGTM

AshleyDumaine:
> ~tested this, works fine. LGTM~
> 
> Seeing issues with other control plane nodes joining the cluster:
> 
> ```
> Feb 29 19:13:21 rah1-control-plane-wrg4n rke2[3412]: time="2024-02-29T19:13:21Z" level=fatal msg="starting kubernetes: preparing server: CA cert validation failed: Get \"https://172.234.204.152:9345/cacerts\": tls: failed to verify certificate: x509: certificate is valid for 10.43.0.1, 127.0.0.1, 172.234.208.182, 192.168.144.181, ::1, not 172.234.204.152"
> Feb 29 19:13:21 rah1-control-plane-wrg4n systemd[1]: rke2-server.service: Main process exited, code=exited, status=1/FAILURE
> Feb 29 19:13:21 rah1-control-plane-wrg4n systemd[1]: rke2-server.service: Failed with result 'exit-code'.
> Feb 29 19:13:21 rah1-control-plane-wrg4n systemd[1]: Failed to start Rancher Kubernetes Engine v2 (server).
> ```
> 
> We might need node-external-ip flag or something similar which adds node's public ip as well.

Unfortunately we can't set that for cloud-provider=external, rke2 fails to start in that case:
```
Feb 29 21:37:08 test-rke2-control-plane-jm9sv rke2[1092]: time="2024-02-29T21:37:08Z" level=fatal msg="can't set node-external-ip while using cloud provider"
Feb 29 21:37:08 test-rke2-control-plane-jm9sv systemd[1]: rke2-server.service: Main process exited, code=exited, status=1/FAILURE
Feb 29 21:37:08 test-rke2-control-plane-jm9sv systemd[1]: rke2-server.service: Failed with result 'exit-code'.
Feb 29 21:37:08 test-rke2-control-plane-jm9sv systemd[1]: Failed to start Rancher Kubernetes Engine v2 (server).
```

AshleyDumaine:
This might actually be something that can be fixed if we set the IP type correctly on the machine controller, I just noticed we always set it to `clusterv1.MachineExternalIP` on https://github.com/linode/cluster-api-provider-linode/blob/main/controller/linodemachine_controller.go#L383 which will cause issues with RKE2 registration. I'll try out the fix there instead of this workaround and see if it helps.

rahulait:
Tested this, LGTM

rahulait (approved):

templates/flavors/rke2/rke2ConfigTemplate.yaml
```diff
@@ -12,7 +12,12 @@ spec:
         kubelet:
           extraArgs:
             - "provider-id=linode://{{ ds.meta_data.id }}"
+      # TODO: use MDS to get public and private IP instead because hostname ordering can't always be assumed
       preRKE2Commands:
+        - |
+          mkdir -p /etc/rancher/rke2/config.yaml.d/
+          echo "node-ip: $(hostname -I | awk '{print $2}')" >> /etc/rancher/rke2/config.yaml.d/capi-config.yaml
+          sed -i "/tls-san:/a - $(hostname -I | awk '{print $1}')" /etc/rancher/rke2/config.yaml
```
eljohnson92, line 20: I am pretty worried about the reliance here on the ordering from `hostname -I` I think we should do this only as the absolutely last option

AshleyDumaine: The other option is a gnarly regex or curling some endpoint that can return the public IP

eljohnson92: I was able to start a cluster and get logs successfully with just setting the `node-ip` to the internal IP, are there other places where I could look for errors requiring the external IP to be set in the SAN as well?

AshleyDumaine: added, thanks 👍 We were seeing errors when trying to view container logs so if both `kubectl logs` works and an HA control plane is able to form we should be good. This seems to be the case with the latest commit.

eljohnson92 (approved):

rahulait (approved):
LGTM

PR #159, opened 2024-03-01
add vpc support for capl clusters

rahulait:
**What type of PR is this?**
/kind feature

With this change, k8s clusters are provisioned by default within a VPC.

1. Since nodebalancers don't yet work with VPC subnets, we attach two interfaces to a linode:
    - eth0 for public and private nodebalancer traffic
    - eth1 for VPC pod-to-pod traffic

    We have to keep pub+priv on eth0 because we want cloud-init to work within VPC. If eth0 is on VPC, then cloud-init fails unless we explicitly add route to make it go over eth1. Once nodebalancer and cloud-init works natively with VPC, we'll use just one interface.

2. We have cilium installed and configured with native routing (no vxlan). All pod-to-pod traffic goes over eth1 within the VPC.
```
root@rah6-control-plane-g8zh6:~# ip a | grep vxlan
root@rah6-control-plane-g8zh6:~#
```
3. Each node in a k8s cluster gets assigned a podCIDR. Linode ccm running in cluster with route-controller feature enabled automatically updates the routes within VPC so that pod-to-pod traffic can flow over the VPC subnet. This PR depends on that feature: see https://github.com/linode/linode-cloud-controller-manager/pull/184

One should be able to ping to pod ip's from any linode within the VPC.

To check if routes are added to interface or not, one can use curl as well:
```
curl --header 'Authorization: Bearer $LINODE_API_TOKEN' -X GET https://api.linode.com/v4/linode/instances/55519597/configs | jq .data[0].interfaces[].ip_ranges
```
The range returned in the output should match with the range present in node's spec `k get node <nodename> -o yaml | yq .spec.podCIDRs`

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controller/linodemachine_controller.go
+++ b/controller/linodemachine_controller.go
@@ -355,16 +355,24 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 			return ctrl.Result{RequeueAfter: reconciler.DefaultMachineControllerWaitForRunningDelay}, nil
 		}
 
+		addrs, err := r.buildInstanceAddrs(ctx, machineScope, linodeInstance.ID)
+		if err != nil {
+			logger.Error(err, "Failed to get instance ip addresses")
+
+			if reconciler.RecordDecayingCondition(machineScope.LinodeMachine,
+				ConditionPreflightReady, string(cerrs.CreateMachineError), err.Error(),
+				reconciler.DefaultMachineControllerPreflightTimeout(r.ReconcileTimeout)) {
+				return ctrl.Result{}, err
+			}
+
+			return ctrl.Result{RequeueAfter: reconciler.DefaultMachineControllerWaitForRunningDelay}, nil
+		}
+		machineScope.LinodeMachine.Status.Addresses = addrs
+
 		conditions.MarkTrue(machineScope.LinodeMachine, ConditionPreflightReady)
 	}
 
 	machineScope.LinodeMachine.Spec.ProviderID = util.Pointer(fmt.Sprintf("linode://%d", linodeInstance.ID))
-
-	addrs, err := r.buildInstanceAddrs(ctx, machineScope, linodeInstance.ID)
-	if err != nil {
-		return linodeInstance, err
-	}
-	machineScope.LinodeMachine.Status.Addresses = addrs
 
 	// Set the instance state to signal preflight process is done
 	machineScope.LinodeMachine.Status.InstanceState = util.Pointer(linodego.InstanceOffline)
--- a/controller/linodemachine_controller.go
+++ b/controller/linodemachine_controller.go
@@ -61,7 +61,7 @@ const (
 	ConditionPreflightAdditionalDisksCreated clusterv1.ConditionType = "PreflightAdditionalDisksCreated"
 	ConditionPreflightConfigured             clusterv1.ConditionType = "PreflightConfigured"
 	ConditionPreflightBootTriggered          clusterv1.ConditionType = "PreflightBootTriggered"
-	ConditionPreflightNBConfigured           clusterv1.ConditionType = "PreflightNBConfigured"
+	ConditionPreflightNetworking             clusterv1.ConditionType = "PreflightNetworking"
 	ConditionPreflightReady                  clusterv1.ConditionType = "PreflightReady"
 )
 
@@ -351,12 +351,12 @@ func (r *LinodeMachineReconciler) reconcileInstanceCreate(
 		conditions.MarkTrue(machineScope.LinodeMachine, ConditionPreflightBootTriggered)
 	}
 
-	if !reconciler.ConditionTrue(machineScope.LinodeMachine, ConditionPreflightNBConfigured) {
+	if !reconciler.ConditionTrue(machineScope.LinodeMachine, ConditionPreflightNetworking) {
 		if err := services.AddNodeToNB(ctx, logger, machineScope); err != nil {
 			logger.Error(err, "Failed to add instance to Node Balancer backend")
 
 			if reconciler.RecordDecayingCondition(machineScope.LinodeMachine,
-				ConditionPreflightNBConfigured, string(cerrs.CreateMachineError), err.Error(),
+				ConditionPreflightNetworking, string(cerrs.CreateMachineError), err.Error(),
 				reconciler.DefaultMachineControllerPreflightTimeout(r.ReconcileTimeout)) {
 				return ctrl.Result{}, err
 			}
@@ -364,7 +364,7 @@ func (r *LinodeMachineReconciler) reconcileInstanceCreate(
 			return ctrl.Result{RequeueAfter: reconciler.DefaultMachineControllerWaitForRunningDelay}, nil
 		}
 
-		conditions.MarkTrue(machineScope.LinodeMachine, ConditionPreflightNBConfigured)
+		conditions.MarkTrue(machineScope.LinodeMachine, ConditionPreflightNetworking)
 	}
 
 	if !reconciler.ConditionTrue(machineScope.LinodeMachine, ConditionPreflightReady) {
--- a/controller/linodemachine_controller.go
+++ b/controller/linodemachine_controller.go
@@ -83,6 +83,11 @@ var requeueInstanceStatuses = map[linodego.InstanceStatus]bool{
 	linodego.InstanceResizing:     true,
 }
 
+type nodeIP struct {
+	ip     string
+	ipType clusterv1.MachineAddressType
+}
+
 // LinodeMachineReconciler reconciles a LinodeMachine object
 type LinodeMachineReconciler struct {
 	client.Client
@@ -359,7 +364,12 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 	}
 
 	machineScope.LinodeMachine.Spec.ProviderID = util.Pointer(fmt.Sprintf("linode://%d", linodeInstance.ID))
-	machineScope.LinodeMachine.Status.Addresses = buildInstanceAddrs(linodeInstance)
+
+	addrs, err := r.buildInstanceAddrs(ctx, logger, machineScope, linodeInstance.ID)
+	if err != nil {
+		return linodeInstance, err
+	}
+	machineScope.LinodeMachine.Status.Addresses = addrs
 
 	// Set the instance state to signal preflight process is done
 	machineScope.LinodeMachine.Status.InstanceState = util.Pointer(linodego.InstanceOffline)
--- a/controller/linodemachine_controller.go
+++ b/controller/linodemachine_controller.go
@@ -83,11 +83,6 @@ var requeueInstanceStatuses = map[linodego.InstanceStatus]bool{
 	linodego.InstanceResizing:     true,
 }
 
-type nodeIP struct {
-	ip     string
-	ipType clusterv1.MachineAddressType
-}
-
 // LinodeMachineReconciler reconciles a LinodeMachine object
 type LinodeMachineReconciler struct {
 	client.Client
@@ -365,7 +360,7 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 
 	machineScope.LinodeMachine.Spec.ProviderID = util.Pointer(fmt.Sprintf("linode://%d", linodeInstance.ID))
 
-	addrs, err := r.buildInstanceAddrs(ctx, logger, machineScope, linodeInstance.ID)
+	addrs, err := r.buildInstanceAddrs(ctx, machineScope, linodeInstance.ID)
 	if err != nil {
 		return linodeInstance, err
 	}
--- a/controller/linodemachine_controller.go
+++ b/controller/linodemachine_controller.go
@@ -61,6 +61,7 @@ const (
 	ConditionPreflightAdditionalDisksCreated clusterv1.ConditionType = "PreflightAdditionalDisksCreated"
 	ConditionPreflightConfigured             clusterv1.ConditionType = "PreflightConfigured"
 	ConditionPreflightBootTriggered          clusterv1.ConditionType = "PreflightBootTriggered"
+	ConditionPreflightNBConfigured           clusterv1.ConditionType = "PreflightNBConfigured"
 	ConditionPreflightReady                  clusterv1.ConditionType = "PreflightReady"
 )
 
@@ -248,7 +249,6 @@ func (r *LinodeMachineReconciler) reconcile(
 	return
 }
 
-//nolint:cyclop // keep top-level preflight condition checks in the same function for readability
 func (r *LinodeMachineReconciler) reconcileCreate(
 	ctx context.Context,
 	logger logr.Logger,
@@ -312,8 +312,17 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 		return ctrl.Result{}, err
 	}
 
+	return r.reconcileInstanceCreate(ctx, logger, machineScope, linodeInstance)
+}
+
+func (r *LinodeMachineReconciler) reconcileInstanceCreate(
+	ctx context.Context,
+	logger logr.Logger,
+	machineScope *scope.MachineScope,
+	linodeInstance *linodego.Instance,
+) (ctrl.Result, error) {
 	if !reconciler.ConditionTrue(machineScope.LinodeMachine, ConditionPreflightConfigured) {
-		if err = r.configureDisks(ctx, logger, machineScope, linodeInstance.ID); err != nil {
+		if err := r.configureDisks(ctx, logger, machineScope, linodeInstance.ID); err != nil {
 			if reconciler.RecordDecayingCondition(machineScope.LinodeMachine,
 				ConditionPreflightConfigured, string(cerrs.CreateMachineError), err.Error(),
 				reconciler.DefaultMachineControllerPreflightTimeout(r.ReconcileTimeout)) {
@@ -327,7 +336,7 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 	}
 
 	if !reconciler.ConditionTrue(machineScope.LinodeMachine, ConditionPreflightBootTriggered) {
-		if err = machineScope.LinodeClient.BootInstance(ctx, linodeInstance.ID, 0); err != nil {
+		if err := machineScope.LinodeClient.BootInstance(ctx, linodeInstance.ID, 0); err != nil {
 			logger.Error(err, "Failed to boot instance")
 
 			if reconciler.RecordDecayingCondition(machineScope.LinodeMachine,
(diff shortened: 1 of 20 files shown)


avestuk:
What happens to NodePort services in this model? Are they bound on eth0 still?

templates/flavors/base/linodeVPC.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
+kind: LinodeVPC
+metadata:
+  name: vpc-${CLUSTER_NAME}
```
eljohnson92, line 5: can we drop the vpc from the name of this? if we want to call it something other than just `cluster_name` we could use some descriptor to mention it is for cluster networking, but I would also be ok with just `$CLUSTER_NAME`

templates/addons/cilium/cilium.yaml
```diff
@@ -15,8 +15,17 @@ spec:
...
+    routingMode: native
+    kubeProxyReplacement: true
+    ipv4NativeRoutingCIDR: 10.0.0.0/8
+    tunnelProtocol: ""
+    enableIPv4Masquerade: true
     ipam:
       mode: kubernetes
+    ipv4:
+      enabled: true
+    ipv6:
```
eljohnson92, line 27: not in scope for this PR, but do we know what the lift would be to enable dual stack networking?

templates/addons/cilium/cilium.yaml
```diff
@@ -15,8 +15,17 @@ spec:
     wait: true
     timeout: 5m
   valuesTemplate: |
+    routingMode: native
+    kubeProxyReplacement: true
```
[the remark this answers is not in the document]
eljohnson92: I think we might need something for all three distros to disable kube-proxy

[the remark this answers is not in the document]
rahulait: Yeah, I think we do. Based on [this](https://docs.cilium.io/en/latest/network/kubernetes/kubeproxy-free/) documentation, for kubeadm we need to skip kube-proxy addon. RK2 and k3s I think might it automatically if we set some flag. I am going to investigate that.
I am also seeing issues with services exposed as of type NodePort and currently investigating why its not working fine with cilium direct routing enabled.

rahulait:
> What happens to NodePort services in this model? Are they bound on eth0 still?

Request for service exposed as NodePort can come on any ip-address (public, private or vpc address). So they are bound to all interfaces on the host. However, we are going to configure cilium-node to have `--direct-routing-device=eth1` so that when cilium-node is forwarding the traffic to the actual pod, it uses eth1's ip-address when doing SNAT. 
If we don't specify this, then it modifies the incoming request with src as the node which received it and dst as pod ip. It sets src as eth0's public ip and dst as pod ip (10/10 subnet) which gets sent over eth1. Since eth1 is connected to VPC, it only allows 10/8 subnet traffic(both src and dst) and hence drops the packet. By setting `--direct-routing-device=eth1`, cilium uses eth1's ip as src ip and now the packet can route over eth1. I am going to push that fix in this PR.

templates/flavors/base/linodeVPC.yaml
rahulait: Updated, added `VPC_NAME` flag. If the flag is not set, it uses `CLUSTER_NAME`

templates/addons/cilium/cilium.yaml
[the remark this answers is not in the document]
rahulait: Updated the codebase and disabled kube-proxy. I couldn't find kube-proxy running in k3s clusters, so it looks like its disabled by default?

templates/addons/cilium/cilium.yaml
```diff
@@ -15,12 +15,31 @@ spec:
...
       requireIPv4PodCIDR: true
     hubble:
       relay:
         enabled: true
       ui:
         enabled: true
+#    ipMasqAgent:
+#      enabled: true
+#    bpf:
+#      masquerade: true
```
rahulait, line 45: Not using bpf by default as I want to get VPC working first. Once the PR is merged, we can have a minor PR to switch from iptables based masquerading to bpf based masquerading.

templates/flavors/base/linodeMachineTemplate.yaml
```diff
@@ -25,4 +25,4 @@ spec:
       region: ${LINODE_REGION}
       authorizedKeys:
       # uncomment to include your ssh key in linode provisioning
-      # - ${LINODE_SSH_PUBKEY:=""}
```
eljohnson92, line 28: same here with comment

templates/flavors/k3s/secret.yaml
```diff
@@ -39,11 +39,15 @@ stringData:
      name: ccm-linode
     spec:
      targetNamespace: kube-system
-     version: ${LINODE_CCM_VERSION:=v0.3.24}
+     version: ${LINODE_CCM_VERSION:=v0.3.25}
```
eljohnson92, line 42: should we bump this to 0.4.1?

templates/flavors/rke2/secret.yaml
```diff
@@ -21,11 +21,15 @@ stringData:
      name: ccm-linode
     spec:
      targetNamespace: kube-system
-     version: ${LINODE_CCM_VERSION:=v0.3.24}
+     version: ${LINODE_CCM_VERSION:=v0.3.25}
```
eljohnson92, line 24: bump to 0.4.1

templates/flavors/k3s/secret.yaml
rahulait: v0.3.25 works for us, it won't work for anyone running without VPC. I updated it to v0.4.1 just to keep using the latest version.

templates/flavors/rke2/secret.yaml
rahulait: done

templates/flavors/base/linodeMachineTemplate.yaml
rahulait: done

eljohnson92:
@rahulait given our default will now leverage VPC+direct routing, can we create a new flavor that includes VPCless deployments so we still have documented support for DCs without VPC support yet?

templates/common-init-files/secret.yaml
```diff
@@ -25,6 +25,7 @@ stringData:
     set -euo pipefail
     export DEBIAN_FRONTEND=noninteractive
     hostnamectl set-hostname "$1" && hostname -F /etc/hostname
+    echo "$(ip a s eth1 |grep 'inet ' |cut -d' ' -f6|cut -d/ -f1) $1" >> /etc/hosts
```
eljohnson92, line 28: if the hostname of a node is an IP rather than the `label` as we are setting it now does that solve anything for us here?

controller/linodemachine_controller_helpers.go
```diff
@@ -94,6 +94,14 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
 		createConfig.RootPass = uuid.NewString()
 	}
 
+	// add public interface to linode (eth0)
+	iface := &linodego.InstanceConfigInterfaceCreateOptions{
+		Purpose: linodego.InterfacePurposePublic,
+		Primary: true,
+	}
+	createConfig.Interfaces = append(createConfig.Interfaces, *iface)
```
cbzzz, line 102: nit: If we're going to assign a pointer to `iface` and then immediately deference it, we can just assign the struct itself to `iface`

controller/linodemachine_controller_helpers.go
```diff
@@ -107,20 +115,59 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
...
+	if len(addresses.IPv4.Public) != 0 {
+		ips = append(ips, nodeIP{ip: addresses.IPv4.Public[0].Address, ipType: clusterv1.MachineExternalIP})
+	}
+
+	// Iterate over interfaces in config and find VPC specific ips
+	for _, iface := range configs[0].Interfaces {
+		if iface.VPCID != nil && iface.IPv4.VPC != "" {
+			ips = append(ips, nodeIP{ip: iface.IPv4.VPC, ipType: clusterv1.MachineInternalIP})
+		}
+	}
```
cbzzz, line 161: Should we also filter these interfaces by the specific VPC referenced in the owner LinodeCluster's spec?

controller/linodemachine_controller.go
```diff
@@ -70,6 +70,11 @@ var requeueInstanceStatuses = map[linodego.InstanceStatus]bool{
 	linodego.InstanceResizing:     true,
 }
 
+type nodeIP struct {
+	ip     string
+	ipType clusterv1.MachineAddressType
+}
```
cbzzz, line 76: nit: Replace this with [`clusterv1.MachineAddress`](https://pkg.go.dev/sigs.k8s.io/cluster-api/api/v1beta1#MachineAddress)

controller/linodemachine_controller_helpers.go
```diff
@@ -107,20 +115,59 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
...
+	addresses, err := machineScope.LinodeClient.GetInstanceIPAddresses(ctx, instanceID)
+	if err != nil {
+		return nil, err
+	}
+
+	// get the default instance config
+	configs, err := machineScope.LinodeClient.ListInstanceConfigs(ctx, instanceID, &linodego.ListOptions{})
+	if err != nil || len(configs) == 0 {
+		logger.Error(err, "Failed to list instance configs")
+		return nil, err
```
cbzzz, line 147: nit: 
```suggestion
		return nil, fmt.Errorf("list instance configs: %w", err)
```
For helper functions like this and in the case where you're both logging and returning an error, you can surface the context of an error with `fmt.Errorf`

controller/linodemachine_controller_helpers.go
```diff
@@ -107,20 +115,59 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
...
 	addrs := []clusterv1.MachineAddress{}
-	for _, addr := range linodeInstance.IPv4 {
-		addrType := clusterv1.MachineExternalIP
-		if addr.IsPrivate() {
-			addrType = clusterv1.MachineInternalIP
-		}
+	ips, err := r.getInstanceIPv4Addresses(ctx, logger, machineScope, instanceID)
+	if err != nil {
+		logger.Error(err, "Failed to get instance ip addresses")
+		return nil, err
```
cbzzz, line 123: nit:
```suggestion
		return nil, fmt.Errorf("get instance ips: %w", err)
```
For helper functions like this and in the case where you're both logging and returning an error, you can surface the context of an error with `fmt.Errorf`

rahulait:
> @rahulait given our default will now leverage VPC+direct routing, can we create a new flavor that includes VPCless deployments so we still have documented support for DCs without VPC support yet?

Sure, that might need a bit more change. Based on our discussions, my understanding was that everything will be there within VPCs and we won't support VPCless. Let me see how much change it would be and if we can also add support for VPCless deployments.

templates/common-init-files/secret.yaml
rahulait: This is needed just in case of kubeadm to tell kubelet to use eth1's ip as node ip. Hostname is still the same. This makes `hostname -i` to return one ip address. If we don't set it, kubelet starts using the ip on eth0 as node ip.
```
oot@rah8-control-plane-24ggw:~# hostname
rah8-control-plane-24ggw
root@rah8-control-plane-24ggw:~# hostname -i
10.0.0.2
root@rah8-control-plane-24ggw:~# 
```

controller/linodemachine_controller_helpers.go
rahulait: I don't think we'll have machines connected to more than one VPC for now. If we do, we'll then need more than 2 interfaces and have to fix linode-ccm to figure out which interface is primary VPC interface etc etc. I feel current logic should be enough. Let me know if you think otherway and we can check further.

templates/flavors/rke2/kustomization.yaml
```diff
@@ -14,6 +15,17 @@ patches:
...
+      group: cluster.x-k8s.io
+      version: v1beta1
+      kind: Cluster
+    patch: |-
+      apiVersion: cluster.x-k8s.io/v1beta1
+      kind: Cluster
+      metadata:
+        name: ${CLUSTER_NAME}
+        labels:
+            cni: cilium
```
cbzzz, line 28: Should this patch also be applied for K3s-flavored clusters?

cbzzz, line 28: Would it be better to use a `helm.cattle.io/v1/HelmChart` resource on the workload cluster to manage it's own Cilium Helm chart instead of [CAAPH](https://github.com/kubernetes-sigs/cluster-api-addon-provider-helm) here?

I see we're using already using this^ pattern for the CCM and for Cilium on the K3s flavor.

templates/common-init-files/secret.yaml
eljohnson92: could we use the `--node-ip` flag instead of this to set the IP of the node?

controller/linodemachine_controller_helpers.go
rahulait: Updated

controller/linodemachine_controller.go
rahulait: Updated

templates/common-init-files/secret.yaml
rahulait: `kubeadm init` doesn't have the `--node-ip` flag. It can be set for kubelet once kubelet is running but the certs get generated by kubeadm using the default ip it detects when its started the first time. I don't see any option to provide node-ip using a kubelet config file as its not supported using config file https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/. Not sure if there is a way we can specify the node-ip with `--cloud-provider=external` during kubeadm init phase to get things working.

templates/flavors/rke2/kustomization.yaml
rahulait: I am not sure about which one is better. We currently have mixed implementations and would like to pick one or the other.

controller/linodemachine_controller_helpers.go
cbzzz: Yeah, let's keep things simple for now then! 🙈

QQ: If users attempt to create a LinodeMachine with some interfaces already specified (e.g. via [`LinodeMachine.Spec.Interfaces`](https://github.com/linode/cluster-api-provider-linode/blob/db4b61cf1ec6a40512789689edc0c6f916aa9eaf/api/v1alpha1/linodemachine_types.go#L62C2-L62C12)), will that just break this feature?

This logic seems to _append_ to the `Interfaces` list in the `linodego.InstanceCreateOptions` and not overwrite it.

templates/common-init-files/secret.yaml
rahulait: It looks like this will work but we need to get the node-ip here somehow:
```
apiVersion: kubeadm.k8s.io/v1beta3
kind: InitConfiguration
nodeRegistration:
  kubeletExtraArgs:
    node-ip: 10.0.0.2
```

rahulait:
CCM e2e tests were failing when run against VPC. We need the PR https://github.com/linode/linode-cloud-controller-manager/pull/182 to be merged as well. Without this fix, it doesn't break cluster installs but breaks exposing services of type loadbalancer.

templates/common-init-files/secret.yaml
rahulait: Updated the PR. PTAL 🙏

controller/linodemachine_controller_helpers.go
rahulait: TIL. Thank you for pointing to this. I tested and updated the logic to use machinetemplate to add additional interface.

templates/flavors/base/linodeVPC.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
+kind: LinodeVPC
+metadata:
+  name: ${VPC_NAME:=${CLUSTER_NAME}}
+spec:
+  region: ${LINODE_REGION}
+  subnets:
+    - ipv4: 10.0.0.0/8
+      label: default
```
komer3, line 10: can we add the following label here: `cluster.x-k8s.io/cluster-name: ${CLUSTER_NAME}`

This is needed to make sure we know what cluster is associated with a VPC and will help us trigger a reconcile loop for VPC when a cluster is paused and unpaused during `clusterctl move` operation.

rahulait: Thanks, updated and added the label.

controller/linodemachine_controller_helpers.go
```diff
@@ -101,20 +102,39 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
...
 }
 
-func buildInstanceAddrs(linodeInstance *linodego.Instance) []clusterv1.MachineAddress {
-	addrs := []clusterv1.MachineAddress{}
-	for _, addr := range linodeInstance.IPv4 {
-		addrType := clusterv1.MachineExternalIP
-		if addr.IsPrivate() {
-			addrType = clusterv1.MachineInternalIP
+func (r *LinodeMachineReconciler) buildInstanceAddrs(ctx context.Context, machineScope *scope.MachineScope, instanceID int) ([]clusterv1.MachineAddress, error) {
+	addresses, err := machineScope.LinodeClient.GetInstanceIPAddresses(ctx, instanceID)
```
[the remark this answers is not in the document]
cbzzz: I think this is the helper that gets called when setting `.status.addresses` during machine creation 😅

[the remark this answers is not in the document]
rahulait: Yeah, this sets the `.status.addresses` during machine creation

cbzzz (approved):

rahulait:
Linode CCM tests pass when running within VPC and route-controller enabled
```
[AfterSuite] PASSED [0.028 seconds]
------------------------------

Ran 22 of 23 Specs in 495.701 seconds
SUCCESS! -- 22 Passed | 0 Failed | 0 Pending | 1 Skipped
PASS

Ginkgo ran 1 suite in 8m18.534643696s
Test Suite Passed
```

controller/linodemachine_controller.go
```diff
@@ -62,6 +62,7 @@ const (
 	ConditionPreflightAdditionalDisksCreated clusterv1.ConditionType = "PreflightAdditionalDisksCreated"
 	ConditionPreflightConfigured             clusterv1.ConditionType = "PreflightConfigured"
 	ConditionPreflightBootTriggered          clusterv1.ConditionType = "PreflightBootTriggered"
+	ConditionPreflightNBConfigured           clusterv1.ConditionType = "PreflightNBConfigured"
```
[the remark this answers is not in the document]
rahulait: updated.

eljohnson92 (approved):

cbzzz (approved):

rahulait:
Going to merge it in and open doc and VPCLess PR soon.

PR #161, opened 2024-03-04
Test LinodeObjectStorageBucket controller

bcm820:
**What type of PR is this?**

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests


controller/linodeobjectstoragebucket_controller.go
```diff
@@ -224,6 +226,8 @@ func (r *LinodeObjectStorageBucketReconciler) reconcileDelete(ctx context.Contex
 		return err
 	}
 
+	r.Recorder.Event(bScope.Bucket, corev1.EventTypeNormal, "Ready", "Object storage bucket deleted")
```
AshleyDumaine, line 229:
```suggestion
	r.Recorder.Event(bScope.Bucket, corev1.EventTypeNormal, clusterv1.DeletedReason, "Object storage bucket deleted")
```

controller/suite_test.go
```diff
@@ -74,15 +74,13 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	Expect(cfg).NotTo(BeNil())
 
-	err = infrastructurev1alpha1.AddToScheme(scheme.Scheme)
-	Expect(err).NotTo(HaveOccurred())
+	Expect(infrastructurev1alpha1.AddToScheme(scheme.Scheme)).To(Succeed())
```
AshleyDumaine, line 77: nit: can we shorten this to `infrav1alpha1` for consistency with the other go files?

cmd/main.go
```diff
@@ -149,6 +150,9 @@ func main() {
 		Recorder:         mgr.GetEventRecorderFor("LinodeObjectStorageBucketReconciler"),
 		WatchFilterValue: objectStorageBucketWatchFilter,
 		LinodeApiKey:     linodeToken,
+		LinodeClientFactory: func(apiKey string) scope.LinodeObjectStorageClient {
+			return scope.CreateLinodeClient(apiKey)
+		},
```
komer3, line 155: We can't use a lambda function here. I recommend creating a function in utils/helpers.go and pass it in. Golangci won't be happy

controller/linodeobjectstoragebucket_controller_test.go
```diff
@@ -0,0 +1,183 @@
...
+	recorder := record.NewFakeRecorder(3)
+
+	secretName := fmt.Sprintf(scope.AccessKeyNameTemplate, *obj.Spec.Label)
+
+	var secret corev1.Secret
+	var mockCtrl *gomock.Controller
+
+	BeforeEach(func() {
+		// Create a new gomock controller for each test run
+		mockCtrl = gomock.NewController(GinkgoT())
```
komer3, line 67: I'm curious: why create a new ctrl before each run? Why not just create it once and use it?

bcm820: It's needed because each gomock controller has its own scope. Invoking `mockctrl.Finish` asserts all of the declared mock expectations for its particular instance.

Dockerfile
```diff
@@ -19,6 +19,7 @@ COPY controller/ controller/
 COPY cloud/ cloud/
 COPY util/ util/
 COPY version/ version/
+COPY mock/ mock/
```
eljohnson92, line 22: do we need to add this just for testing or is it also needed while the app is running?

go.mod
```diff
@@ -37,7 +37,6 @@ require (
 	github.com/gobuffalo/flect v1.0.2 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
-	github.com/golang/mock v1.6.0 // indirect
```
eljohnson92, line 40: is this `mock` removal on purpose?

bcm820: Yes. The correct module is `go.uber.org/mock v0.4.0`.

Dockerfile
bcm820: I was getting errors with the build since I'm importing `mock` in `cloud/scope` to avoid a circular dependency. But I think I can get around that so we don't need to include it in the build.

cloud/scope/client.go
```diff
@@ -3,10 +3,27 @@ package scope
...
+	DeleteObjectStorageKey(ctx context.Context, keyID int) error
+}
+
+// LinodeObjectStorageClientBuilder is a function that returns a LinodeObjectStorageClient.
+type LinodeObjectStorageClientBuilder func(apiKey string) (LinodeObjectStorageClient, error)
+
+// CreateLinodeObjectStorageClient is the main implementation of LinodeObjectStorageClientBuilder.
+func CreateLinodeObjectStorageClient(apiKey string) (LinodeObjectStorageClient, error) {
+	return CreateLinodeClient(apiKey)
+}
```
komer3, line 25: Question: What is this func being used for?

cloud/scope/object_storage_bucket.go
```diff
@@ -58,7 +62,7 @@ func NewObjectStorageBucketScope(ctx context.Context, apiKey string, params Obje
 		}
 		apiKey = string(data)
 	}
-	linodeClient, err := createLinodeClient(apiKey)
+	linodeClient, err := params.LinodeClientBuilder(apiKey)
```
komer3, line 65: Couldn't we just call `CreateLinodeClient(apiKey)` since that returns a `*linodego.client` struct which satisfies `LinodeObjectStorageClient` interface?

controller/linodeobjectstoragebucket_controller_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	It("should reconcile an object delete", func() {
+		mockClient := mock.NewMockLinodeObjectStorageClient(mockCtrl)
+
+		for i := range 2 {
+			mockClient.EXPECT().
+				DeleteObjectStorageKey(gomock.Any(), i).
+				Return(nil).
+				Times(1)
+		}
```
komer3, line 161: Does gomock let us do this?
```suggestion
	mockClient.EXPECT().
		DeleteObjectStorageKey(gomock.Any(), i).
		Return(nil).
		Times(2)
```

controller/linodeobjectstoragebucket_controller_test.go
```diff
@@ -0,0 +1,184 @@
...
+	recorder := record.NewFakeRecorder(3)
+
+	secretName := fmt.Sprintf(scope.AccessKeyNameTemplate, *obj.Spec.Label)
+
+	var secret corev1.Secret
+	var mockCtrl *gomock.Controller
+
+	BeforeEach(func() {
+		// Create a new gomock controller for each test run
+		mockCtrl = gomock.NewController(GinkgoT())
```
komer3, line 72:
```suggestion

	BeforeEach(func() {
		// Create a new gomock controller for each test run
		mockCtrl := gomock.NewController(GinkgoT())
```

cloud/scope/client.go
bcm820: I added this since we don't want to use an anonymous function [here](https://github.com/linode/cluster-api-provider-linode/blob/befb7302c346c65168ba044e4f10e064e7ce583e/cmd/main.go#L150) (to pass golangci-lint rules).

controller/linodeobjectstoragebucket_controller_test.go
bcm820: The mock recorder returned from `EXPECT()` counts each distinct function call (and not how many times a function is called), which is why we assert that it will call both `DeleteObjectStorageKey(any, 0)` and `DeleteObjectStorageKey(any, 1)` exactly once each.

controller/linodeobjectstoragebucket_controller_test.go
bcm820: `mockCtrl` is initialized outside of the function closures because it must be accessed inside of each `It` test spec node.

komer3 (approved):
LGTM!

PR #162, opened 2024-03-04
chore: migrate from kuttl to chainsaw (linodecluster-controller)

eddycharly:
Migrate from kuttl to chainsaw.

### DIFF
--- a/e2e/linodecluster-controller/minimal-linodecluster/chainsaw-test.yaml
+++ b/e2e/linodecluster-controller/minimal-linodecluster/chainsaw-test.yaml
@@ -1,0 +1,81 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  creationTimestamp: null
+  name: minimal-linodecluster
+spec:
+  steps:
+  - name: step-00
+    try:
+    - assert:
+        file: 00-assert.yaml
+  - name: step-01
+    try:
+    - apply:
+        file: 01-create-cluster.yaml
+  - name: step-02
+    try:
+    - apply:
+        file: 02-create-linodecluster.yaml
+    - assert:
+        file: 02-assert.yaml
+  - name: step-03
+    try:
+    - script:
+        env:
+        - name: TARGET_API
+          value: api.linode.com
+        - name: TARGET_API_VERSION
+          value: v4beta
+        - name: URI
+          value: nodebalancers
+        - name: FILTER
+          value: '{"label":"linodecluster-test-api-server"}'
+        content: |
+          set -e
+          curl -s \
+            -H "Authorization: Bearer $LINODE_TOKEN" \
+            -H "X-Filter: $FILTER" \
+            -H "Content-Type: application/json" \
+            "https://$TARGET_API/$TARGET_API_VERSION/$URI"
+        check:
+          ($error): ~
+          (json_parse($stdout)):
+            results: 1
+  - name: step-04
+    try:
+    - delete:
+        ref:
+          apiVersion: cluster.x-k8s.io/v1beta1
+          kind: Cluster
+          name: cluster-test
+    - error:
+        file: 04-error.yaml
+  - name: step-05
+    try:
+    - script:
+        env:
+        - name: TARGET_API
+          value: api.linode.com
+        - name: TARGET_API_VERSION
+          value: v4beta
+        - name: URI
+          value: nodebalancers
+        - name: FILTER
+          value: '{"label":"linodecluster-test-api-server"}'
+        content: |
+          set -e
+          curl -s \
+            -H "Authorization: Bearer $LINODE_TOKEN" \
+            -H "X-Filter: $FILTER" \
+            -H "Content-Type: application/json" \
+            "https://$TARGET_API/$TARGET_API_VERSION/$URI"
+        check:
+          ($error): ~
+          (json_parse($stdout)):
+            results: 0
+  - name: step-06
+    try:
+    - apply:
+        file: 06-cleanup-cluster.yaml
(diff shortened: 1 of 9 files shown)


eddycharly:
```
        content: 'URI="nodebalancers" FILTER="{\"label\":\"linodecluster-test-api-server\"}"
          make callLinodeApiGet | grep ''results": 1'''
```
Is this supposed to fail the test if the `grep` does not find the expected results ? 🤔

eljohnson92:
yes, that is the goal of it

eddycharly:
@eljohnson92 but grep does not exit with `1` if not found right ?

eddycharly:
And why do we sometimes have `FILTER="{\"label\":\"linodecluster-test-api-server\"}"` and other times `FILTER="{\"tags\":\"linodecluster-test-api-server\"}"` shouldn't the filter be the same ?

e2e-chainsaw/linodecluster-controller/minimal-linodecluster/chainsaw-test.yaml
```diff
@@ -0,0 +1,75 @@
...
+        - name: FILTER
+          value: '{"label":"linodecluster-test-api-server"}'
+        content: |
+          set -e
+          curl -s \
+            -H "Authorization: Bearer $LINODE_TOKEN" \
+            -H "X-Filter: $FILTER" \
+            -H "Content-Type: application/json" \
+            "https://$TARGET_API/$TARGET_API_VERSION/$URI"
+        # TODO: check results
```
eddycharly, line 42: @eljohnson92 i replaced the makefile target with an inline script

eddycharly: If the payload returned is json it's easy to write a robust assertion

eljohnson92: ah, yes, I like the way that is laid out. it is a json response and I would much prefer checking values within the json, is that something we can do natively with Chainsaw or would we have to use JQ to handle checking fields/values?

eddycharly: @eljohnson92 i implemented the robust checks 🤞

eddycharly: @eljohnson92 i implemented the check, it's natively supported

e2e-chainsaw/linodecluster-controller/minimal-linodecluster/chainsaw-test.yaml
```diff
@@ -0,0 +1,81 @@
...
+    - assert:
+        file: 02-assert.yaml
+  - name: step-03
+    try:
+    - script:
+        env:
+        - name: TARGET_API
+          value: api.linode.com
+        - name: TARGET_API_VERSION
+          value: v4beta
```
eddycharly, line 30: Those two vars could be set globally 🤷

eddycharly:
Please allow workflow to run, i'd like to see how it works 🙏

eljohnson92:
@eddycharly we have an issue because we are using API tokens to talk to the Linode API it can't be run by forks, I'm trying to see if there is a quick fix I can get in to allow us to trigger this workflow even from forks

eddycharly:
@eljohnson92 ah i see, this PR doesn't remove kuttl tests, it runs both kuttl and chainsaw side by side.

eddycharly:
Indeed
```
capl-control… │ [manager] 2024-03-04T22:47:48Z	ERROR	setup	unable to start operator	{"error": "failed to get LINODE_TOKEN environment variable"}
```

eljohnson92:
I was able to test this locally so I would be open to merging this, before we do that could we change this to just replace the existing e2e test version?

eddycharly:
> I was able to test this locally so I would be open to merging this, before we do that could we change this to just replace the existing e2e test version of minimal-linodecluster?

You want me to remove the kuttl version of the minimal-linodecluster test ?

eddycharly:
> instead of creating a separate e2e-chainsaw/ directory structure I'd rather we just replace the existing KUTTL test with this new chainsaw test in place. unless that will make test targets way too messy. it would be good to see the diff here in place for review

Doing this is going to be challenging, kuttl will think it can run the test but it will fail.
That's why i created two separate folders, the plan is to replace the kuttl e2e test folder by the chainsaw one at the end (and keep only chainsaw). We will see the diff at this time.

eddycharly:
@eljohnson92 let me try something 🤞

eljohnson92:
understand that might be challenging, I would be fine with merging this separately if it isn't immediately easy to fix

eddycharly:
@eljohnson92 I just pushed, chances are it will work 🤞

eljohnson92:
awesome, just ran another test and it passed. LGTM but I'll let some other folks take a look at it tomorrow before merging it in

eddycharly:
> just ran another test and it passed

Awesome, I will take care of other tests in follow-up PRs.
Thanks!

AshleyDumaine (approved):

e2e/linodecluster-controller/minimal-linodecluster/chainsaw-test.yaml
```diff
@@ -0,0 +1,81 @@
...
+          set -e
+          curl -s \
+            -H "Authorization: Bearer $LINODE_TOKEN" \
+            -H "X-Filter: $FILTER" \
+            -H "Content-Type: application/json" \
+            "https://$TARGET_API/$TARGET_API_VERSION/$URI"
+        check:
+          ($error): ~
+          (json_parse($stdout)):
+            results: 1
```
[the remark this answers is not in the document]
eddycharly: Doc is spread between https://kyverno.github.io/chainsaw/latest/operations/check/ and https://kyverno.github.io/kyverno-json/latest/

Chainsaw uses assertion trees from kyverno-json, we didn't repeat the same doc in chainsaw.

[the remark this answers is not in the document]
eddycharly: I plan to rewrite docs for the v0.2.x series :)

PR #163, opened 2024-03-05
feat: extend multi-tenancy for LinodeMachines

cbzzz:
**What type of PR is this?**
/kind feature

Updates LinodeMachines to be able to supply an (optional) CredentialsRef field to override the credentials of their owner LinodeCluster.

This allows for a (potentially future) enhancement where clusters can provision Linode resources cross-account.

Fixes #

This is being introduced now (rather than when the feature is determined to be absolutely needed) to provide a more stable API during development. However, we may wait to add this feature since:
1. The setting introduced is optional. Adding it later would have no effect on existing clusters post-addition.
2. We are allowed to break, drop support, etc. for alpha APIs without notice: https://kubernetes.io/docs/reference/using-api/#api-versioning

**TODOs**:

- [X] squashed commits
- [X] includes documentation
- [ ] adds unit tests

### DIFF
--- a/api/v1alpha1/linodemachine_types.go
+++ b/api/v1alpha1/linodemachine_types.go
@@ -18,6 +18,7 @@ package v1alpha1
 
 import (
 	"github.com/linode/linodego"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


eljohnson92 (approved):

eljohnson92:
LGTM

AshleyDumaine (approved):

PR #167, opened 2024-03-06
support byovpc and clean up list filtering

### LINKED ISSUE #149: trying to import  a VPC for BYOVPC does not work
## General:

* [x] Have you removed all sensitive information, including but not limited to access keys and passwords?
* [x] Have you checked to ensure there aren't other open or closed [Pull Requests](../../pulls) for the same bug/feature/question?

----

## Bug Reporting
When specifying vpcID on linodeVPC creation I would expect the VPC matching that ID is used, instead it appears a new VPC is being created
### Expected Behavior
with a vpc labeled `test-vpc`
applied file: 
```
kind: LinodeVPC
metadata:
  name: linodevpc-test
spec:
  vpcID: 12345
  region: us-ord

```
generated VPC
```
kind: LinodeVPC
metadata:
  name: linodevpc-test
spec:
  label: test-vpc
  region: us-ord
  vpcID: 12345
status:
  ready: true

```
### Actual Behavior

generated VPC
```
kind: LinodeVPC
metadata:
  name: linodevpc-test
spec:
  label: linodevpc-test
  region: us-ord
  vpcID: 897934
status:
  ready: true

```
### Steps to Reproduce the Problem

  1. Apply the above minimal linodeVPC

#### Additional Notes

----

For general help or discussion, join the [Kubernetes Slack team](https://kubernetes.slack.com/messages/CD4B15LUR/details/) channel `#linode`. To sign up, use the [Kubernetes Slack inviter](http://slack.kubernetes.io/).

The [Linode Community](https://www.linode.com/community/questions/) is a great place to get additional support.

eljohnson92:
/kind cleanup

This PR adds support for byovpc and changes all linode resource listing to use the same utils method
Fixes #149

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


util/helpers.go
```diff
@@ -13,14 +14,15 @@ func Pointer[T any](t T) *T {
...
+	case id != 0:
+		filter["id"] = strconv.Itoa(id)
+	case label != "":
 		filter["label"] = label
-	}
-
-	if len(tags) != 0 {
+	case len(tags) != 0:
 		filter["tags"] = strings.Join(tags, ",")
 	}
```
nesv, line 27: Using the switch will only ever result in the execution of the first case expression that evaluates to true; if `id != 0` and `len(tags) != 0`, only the `id != 0` case block would be executed. Is this the intended behaviour?

eljohnson92: id's and labels are both completely unique so if you specify a label there isn't a point of also filtering by tags so the goal is if you specify ID just use that, if there isn't an ID just use label, and if neither of them then fall back to filtering by tags

eljohnson92: tried to capture this on line 17 but maybe I can make that message more clear that each option is used exclusively

nesv: Sounds good!

As a follow-up, if `id == 0`, `label == ""` and `len(tags) == 0`, the JSON filter string will still be constructed, but with nothing in it. Is this okay? Will it have any unintended effects when calling the Linode API? Could the function short-circuit, and return an empty string if no filters were set?

nesv: > each option is used exclusively

:thinking: Should this be three, separate functions perhaps? It would help make the intent clearer when the caller wants to use filtering.

eljohnson92: it is likely this logic is going to be something all of the controllers will need with "if we have an ID then find the resource by an ID, if we just have a label try and find it by the label, otherwise search for it by the labels" separating the functions might mean we just move that duplicate logic to each of the controllers?

eljohnson92: I was also considering maybe setting tags as the default so we always have something, we shouldn't be creating resources without any tags or known labels or anything so it might be safe to do that? the default would be to fail anyways because it would not filter anything and the next check would fail because more than 1 resource is returned

nesv: Just to capture a discussion that happened out of band of this PR:

We floated the idea of using a `type Filter struct` instead of a function, that implements the `encoding/json.Marshaler` and `fmt.Stringer` interfaces instead, with full documentation on the preferences of each of the fields in the `Filter` struct.

nesv (approved):
:partying_face:

nesv (approved):

PR #172, opened 2024-03-07
🧹 [cleanup] golangci-lint updates

AshleyDumaine:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:
- Turns on the `gci` linter for checking import ordering on lint. Interesting caveat with `gci`: https://github.com/daixiang0/gci/issues/135. For this reason, we exclude `gci` linter on `cmd/main.go` and `controller/suite_test.go`
- Turns off nlreturn
- Turns off gocyclo since it's abandoned
- Bumps cyclop to 17

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #174, opened 2024-03-07
🧹 [cleanup] - refactoring to reduce cyclomatic complexity in machine controller, add missing autogenerated changes

AshleyDumaine:
**What type of PR is this?**

/kind cleanup

- Some cleanup to reduce cyclomatic complexity in the machine controller
- Adds missing autogenerated changes from #163 (cc @cbzzz)

Fixes #

**Special notes for your reviewer**: Based on #172 which was just focused on some lint changes

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


AshleyDumaine:
Created #175 to update GHA for diff checking (out of scope for this PR)

controller/linodemachine_controller.go
```diff
@@ -370,31 +297,12 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 		return nil, err
 	}
 
-	if linodeInstance == nil {
-		err = errors.New("missing instance")
-
-		logger.Error(err, "Panic! Failed to create isntance")
-
-		return nil, err
-	}
```
cbzzz, line 379: I'm not totally sure that we can remove this check? I assume it double-checks this part: https://github.com/linode/cluster-api-provider-linode/blob/9a5dfbdfd473c8e8b8b1cd38ba72a38cc9232ea4/controller/linodemachine_controller.go#L246-L251

But I suspect this might be related to a similar issue with the Linode Firewall, where if you call the `linodego.Client` with an incorrect `util.Filter` you can get a `nil` object back with no error.

cbzzz (approved):
Only **_you_** 🐻👉 can prevent complexity! 🔥🔥

controller/linodemachine_controller.go
```diff
@@ -103,76 +103,31 @@ func (r *LinodeMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-
 	machineScope, err := scope.NewMachineScope(
 		ctx,
 		r.LinodeApiKey,
 		scope.MachineScopeParams{
 			Client:        r.Client,
 			Cluster:       cluster,
 			Machine:       machine,
-			LinodeCluster: linodeCluster,
+			LinodeCluster: &infrav1alpha1.LinodeCluster{},
```
nesv, line 130: Can this field be elided from the inline struct literal, and the zero-value (`nil`) be used instead?

controller/linodemachine_controller_helpers.go
```diff
@@ -94,9 +94,93 @@ func (*LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineScop
...
+			if match = linodeMachine.OwnerReferences[i].UID == machine.UID; match {
+				break
+			}
+		}
+		if !match {
+			log.Info("Failed to find the referenced owner machine, skipping reconciliation", "references", linodeMachine.OwnerReferences, "machine", machine.ObjectMeta)
+
+			return nil, err
+		}
+	}
```
nesv, line 154: style nit: Since there are no variables scoped to the switch block, could this default case be moved out of the switch and into the main function body?

controller/linodemachine_controller_helpers.go
```diff
@@ -94,9 +94,93 @@ func (*LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineScop
...
+		if err = client.IgnoreNotFound(err); err != nil {
+			log.Error(err, "Failed to fetch cluster by label")
+		}
+
+		return nil, err
+	case cluster == nil:
+		err = errors.New("missing cluster")
+		log.Error(err, "Missing cluster")
+
+		return nil, err
```
nesv, line 172: The error isn't adding anything useful into the context of the log message.

```suggestion
		log.Error(nil, "Missing cluster")
		return nil, errors.New("missing cluster")
```

controller/linodemachine_controller_helpers.go
```diff
@@ -94,9 +94,93 @@ func (*LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineScop
...
+	case cluster.Spec.InfrastructureRef == nil:
+		err = errors.New("missing infrastructure reference")
+
+		log.Error(err, "Missing infrastructure reference")
+
+		return nil, err
```
nesv, line 178: Same as above: no useful context is being added to the log message.

controller/linodemachine_controller.go
```diff
@@ -103,76 +103,31 @@ func (r *LinodeMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-			}
-		}
-
-		if !match {
-			log.Info("Failed to find the referenced owner machine, skipping reconciliation", "references", linodeMachine.OwnerReferences, "machine", machine.ObjectMeta)
-
-			return ctrl.Result{}, nil
-		}
 	}
+	log = log.WithValues("LinodeMachine: ", machine.Name)
```
nesv, line 116: The `WithValues` method takes a key and value pair. I don't think you need to specify the `: ` separator in the key argument.

controller/linodemachine_controller.go
```diff
@@ -103,76 +103,31 @@ func (r *LinodeMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-
 	machineScope, err := scope.NewMachineScope(
 		ctx,
 		r.LinodeApiKey,
 		scope.MachineScopeParams{
 			Client:        r.Client,
 			Cluster:       cluster,
 			Machine:       machine,
-			LinodeCluster: linodeCluster,
+			LinodeCluster: nil,
```
nesv, line 130: For what it's worth, you can simply _not_ specify this field, and it will default to the zero-value for the field's type. Since `LinodeCluster` is a `*T` (pointer to a type), the zero value is `nil`.

nesv (approved):
LGTM!

PR #188, opened 2024-03-11
[fix] Remove OBJ label field and add key ID status fields

bcm820:
**What type of PR is this?**

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


controller/linodeobjectstoragebucket_controller.go
```diff
@@ -186,38 +184,16 @@ func (r *LinodeObjectStorageBucketReconciler) reconcileApply(ctx context.Context
...
-	if err := r.Client.Update(ctx, secret); err != nil {
-		bScope.Logger.Error(err, "Failed to remove finalizer from secret; will not be deleted")
-		r.setFailure(bScope, err)
-
-		return err
-	}
-
 	if !controllerutil.RemoveFinalizer(bScope.Bucket, infrav1alpha1.GroupVersion.String()) {
-		bScope.Logger.Error(err, "Failed to remove finalizer from bucket; will not be deleted")
+		err := fmt.Errorf("failed to remove finalizer from bucket; unable to delete")
```
AshleyDumaine, line 195: nit: `errors.New` since we're not formatting the error message

cloud/services/object_storage_buckets.go
```diff
@@ -109,21 +96,13 @@ func createObjectStorageKey(ctx context.Context, bScope *scope.ObjectStorageBuck
...
-		bScope.Logger.Error(err, "Failed to read secret with access keys to revoke; must be manually revoked")
-
-		return fmt.Errorf("failed to read secret %s with access keys: %w", secret.Name, err)
-	}
-
+func RevokeObjectStorageKeys(ctx context.Context, bScope *scope.ObjectStorageBucketScope) error {
 	var errs []error
-	for _, keyID := range keyIDs {
-		if err := revokeObjectStorageKey(ctx, bScope, keyID); err != nil {
+	for _, keyID := range []*int{bScope.Bucket.Status.ReadWriteKeyID, bScope.Bucket.Status.ReadOnlyKeyID} {
```
AshleyDumaine, line 101: using a for loop here seems a bit odd unless we were iterating over something like `Status.AccessKeys`. Will there only ever a single key for each rather than a list for both RW and RO? Also do we need to distinguish between RW vs RO?

bcm820: There should only ever be a single key for each. I think storing the IDs as an array makes sense. Will make that change.

eljohnson92 (approved):

PR #190, opened 2024-03-13
e2e: template resources

cbzzz:
**What type of PR is this?**
/kind cleanup

This templates resource names (using the abbreviated commit hash) across the E2E tests to prevent naming collisions.

Fixes #

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


Makefile
```diff
@@ -138,7 +139,7 @@ test: manifests generate fmt vet envtest ## Run tests.
 
 .PHONY: e2etest
 e2etest: manifests generate local-deploy chainsaw
-	$(CHAINSAW) test ./e2e
+	echo "name: capl-e2e-$(GIT_REF)" | $(CHAINSAW) test ./e2e --values -
```
eljohnson92, line 142: could this be supplied as just a flag rather than piping it to the command?

cbzzz: No, we only have the following options: https://kyverno.github.io/chainsaw/latest/configuration/values/#example

eljohnson92: I asked the chainsaw devs and they mentioned there is an `env()` function we could use to just read env variables in the tests, perhaps we could do that instead to just read in the GIT_REF?

cbzzz: Unfortunately, Chainsaw seems to run tests in something like a subshell (e.g. `$USER` is accessible) so Make variables are not available.

eljohnson92 (approved):

PR #195, opened 2024-03-14
[BUG] - Improved error handling to avoid nil ptrs

komer3:
**What type of PR is this?**

BUG FIX: Using the correct method to unwrap errors from linodego and handling it

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


controller/linodecluster_controller.go
```diff
@@ -165,7 +164,12 @@ func (r *LinodeClusterReconciler) reconcileCreate(ctx context.Context, logger lo
 	linodeNB, err := services.CreateNodeBalancer(ctx, clusterScope, logger)
 	if err != nil {
 		setFailureReason(clusterScope, cerrs.CreateClusterError, err, r)
+		return err
+	}
 
+	if linodeNB == nil {
```
rahulait, line 170: I wonder if api will ever return nil nodebalancer and no error.

rahulait (dismissed):
Except the minor comment/question, the PR looks good to me.

cloud/services/object_storage_buckets.go
```diff
@@ -18,9 +17,11 @@ func EnsureObjectStorageBucket(ctx context.Context, bScope *scope.ObjectStorageB
 		bScope.Bucket.Spec.Cluster,
 		bScope.Bucket.Name,
 	)
-	linodeErr := &linodego.Error{}
-	if errors.As(err, linodeErr) && linodeErr.StatusCode() != http.StatusNotFound {
-		return nil, fmt.Errorf("failed to get bucket from cluster %s: %w", bScope.Bucket.Spec.Cluster, err)
+	if err != nil {
+		linodeErr := linodego.NewError(err)
+		if linodeErr.StatusCode() != http.StatusNotFound {
```
AshleyDumaine, line 22: I think for this we should be able to use `util.IgnoreLinodeAPIError(err, http.StatusNotFound)`

komer3: Use I was able to use that. Thanks for pointing that out!

rahulait (approved):

PR #209, opened 2024-03-21
Install etcd-backup-restore for child clusters

amold1:
**PR Summary:**
Install upstream [etcd-backup-restore](https://github.com/gardener/etcd-backup-restore) on child clusters to enable snaphotting, and uploading to LinodeOBJ, with server-side encryption enabled, for restoring

---------

**What this PR does / why we need it:**
This PR creates a new addon etcd-backup-restore. The addon applies four resources.
1. A `LinodeObjectStorageBucket` to create the bucket that etcdbrctl will interact with.
2. A `ClusterResourceSet` to copy the access details `secret`, that the OBJ controller creates, on to the child cluster.
3. A `configMap` that encapsulates the `etcd-backup-restore` `statefulset`. The `statefulset` itself has 2 containers for the pod. An `initContainer` that creates a credentials config file for the etcdbrctl binary to use for auth and the etcdbrctl `container` itself.
4. A `ClusterResourceSet` to copy the `etcd-backup-restore` `statefulset` on to the child cluster.

---------

**Additonal Notes:**
Because the `ClusterResourceSet` needs a secret to be of type `addons.cluster.x-k8s.io/resource-set`, we have changed the OBJ behavior when creating the access key secret. If the `secretType` in the `spec` for `LinodeObjectStorageBucket` is set to that type, we will create the `generic` secret and encapsulate it in a secret of type  `addons.cluster.x-k8s.io/resource-set`

---------

**Related PRs:**
Upstream doesn't currently support Server-Side Encryption so we have a [PR](https://github.com/gardener/etcd-backup-restore/pull/719) open to add that feature

---------

**Test:**
```
## Setup Env Vars
export KUBERNETES_VERSION=v1.29.1
export LINODE_REGION=us-ord
export LINODE_TOKEN=<your-api-token>
export LINODE_CONTROL_PLANE_MACHINE_TYPE=g6-standard-2
export LINODE_MACHINE_TYPE=g6-standard-2
export CLUSTER_NAME=<cluster-name>
export OBJ_BUCKET_REGION=us-ord-1
export ETCDBR_IMAGE=docker.io/amoldeodhar/etcdbrctl:sse.2
export SSE_KEY=<aes266-key>   # sample: cdQdZ3PrKgm5vmqxeqwQCuAWJ7pPVyHg

# Deploy management cluster
make local-deploy

# Create manifests locally
make local-release

# Create cluster config
"""
providers:
  - name: linode
    url: ${HOME}/cluster-api-provider-linode/infrastructure-linode/0.0.0/infrastructure-components.yaml
    type: InfrastructureProvider
"""

# Deploy child cluster
clusterctl generate cluster $CLUSTER_NAME \
  --kubernetes-version v1.29.1 \
  --infrastructure linode:0.0.0 \
  --flavor etcd-backup-restore \
  | kubectl apply -f -

# Check if cluster is ready
kubectl get cluster $CLUSTER_NAME -o yaml

# Get child cluster kubeconfig
clusterctl get kubeconfig $CLUSTER_NAME > $CLUSTER_NAME-kubeconfig
export KUBECONFIG=$CLUSTER_NAME-kubeconfig

# Check logs for the backup pod
kubectl -n kube-system logs etcd-backup-0

You can now check on CloudManager to see if the bucket was created and if it has objects under `bucket_name/etcd-backup/v2/Full******.gz`
```


cloud/scope/object_storage_bucket.go
```diff
@@ -101,15 +101,46 @@ func (s *ObjectStorageBucketScope) AddFinalizer(ctx context.Context) error {
 
 // ApplyAccessKeySecret applies a Secret containing keys created for accessing the bucket.
 func (s *ObjectStorageBucketScope) ApplyAccessKeySecret(ctx context.Context, keys [NumAccessKeys]linodego.ObjectStorageKey, secretName string) error {
+	var secretStringData map[string]string
+
+	// If the secret is of ClusterResourceSet type, encapsulate real data in the outer data
+	if s.Bucket.Spec.SecretType == "addons.cluster.x-k8s.io/resource-set" {
+		bucketDataSecret := `kind: Secret
```
eljohnson92, line 108: instead of doing this templating, could we create a normal `secret` object and just convert that to a string for the access-keys-secret value?

templates/flavors/etcd-backup-restore/kustomization.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../base
```
eljohnson92, line 4: instead of pulling in `../base` and these other files could we just pull in `../default`?

amold1: done!

templates/addons/etcd-backup-restore/etcd-backup-restore.yaml
```diff
@@ -0,0 +1,191 @@
...
+            resources:
+              limits:
+                cpu: 100m
+                memory: 1Gi
+              requests:
+                cpu: 23m
+                memory: 128Mi
+            env:
+            - name: AWS_APPLICATION_CREDENTIALS_JSON
+              value: ${CREDS_DIRNAME}${CREDS_FILENAME}
```
eljohnson92, line 157: Can this just be hard coded? any reason a user would want to configure this?

templates/addons/etcd-backup-restore/etcd-backup-restore.yaml
```diff
@@ -0,0 +1,191 @@
...
+spec:
+  clusterSelector:
+    matchLabels:
+      etcd-backup: "true"
+      cluster: ${CLUSTER_NAME}
+  resources:
+  - kind: Secret
+    name: ${CLUSTER_NAME}-ssekey
+  strategy: ApplyOnce
+---
```
eljohnson92, line 30: can we break this linodeObjectStorageBucket into it's own file in the etcd-backup-restore flavor?

templates/addons/etcd-backup-restore/etcd-backup-restore.yaml
amold1: It certainly can be. The directory name is being used for the mountPath so I figured we could split it but that just increases the number of env vars we need so I'll hardcode it

templates/addons/etcd-backup-restore/etcd-backup-restore.yaml
```diff
@@ -0,0 +1,191 @@
...
+          app.kubernetes.io/name: etcd
+      updateStrategy:
+        type: RollingUpdate
+      template:
+        metadata:
+          labels:
+            app.kubernetes.io/name: etcd
+        spec:
+          initContainers:
+          - name: create-creds-file
```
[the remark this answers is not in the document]
amold1: The chart isn't creating the creds file. And I did not see any implementation details about this. The only doc I found only mentions that the "file needs to be made available". [doc link](https://github.com/gardener/etcd-backup-restore/blob/master/docs/deployment/getting_started.md#various-ways-to-pass-credentials)

templates/addons/etcd-backup-restore/etcd-backup-restore.yaml
amold1: I have broken this up into 3. OBJ, SSE key, and the etcdbr itself. Each also has its own ClusterResourceSet in the same file

templates/addons/etcd-backup-restore/etcd-backup-restore.yaml
amold1: Done. Hard-coded

api/v1alpha1/linodeobjectstoragebucket_types.go
```diff
@@ -43,6 +43,11 @@ type LinodeObjectStorageBucketSpec struct {
 	// +optional
 	// +kubebuilder:default=0
 	KeyGeneration *int `json:"keyGeneration,omitempty"`
+
+	// secretType sets the type for the access-keys secret that will be generated by the controller.
+	// +optional
+	// +kubebuilder:default=addons.cluster.x-k8s.io/resource-set
+	SecretType string `json:"secretType,omitempty"`
```
bcm820, line 50: What do you think about defaulting `SecretType` to `Opaque` to support the general use case of provisioning buckets? Otherwise, the formatting of the secret data could throw users off.

cloud/scope/object_storage_bucket.go
```diff
@@ -17,6 +17,18 @@ import (
 	infrav1alpha1 "github.com/linode/cluster-api-provider-linode/api/v1alpha1"
 )
 
+const bucketDataSecret = `kind: Secret
+apiVersion: v1
+metadata:
+  name: etcd-backup
+  namespace: kube-system
+stringData:
+  bucket_name: %s
```
bcm820, line 26: How about `bucket_label` versus `bucket_name`?

api/v1alpha1/linodeobjectstoragebucket_types.go
```diff
@@ -43,6 +43,11 @@ type LinodeObjectStorageBucketSpec struct {
 	// +optional
 	// +kubebuilder:default=0
 	KeyGeneration *int `json:"keyGeneration,omitempty"`
+
+	// secretType sets the type for the access-keys secret that will be generated by the controller.
```
bcm820, line 47: nit: consistent with other field comments

```suggestion
	// SecretType sets the type for the access-keys secret that will be generated by the controller.
```

api/v1alpha1/linodeobjectstoragebucket_types.go
amold1: So Evan and I had a chat about this and the OBJ use case is going to be for the etcd backup almost 90% of the times. @eljohnson92 thoughts? Should we keep this as is or swap the defaults?

cloud/scope/object_storage_bucket.go
amold1: I went with `bucket_name` to be consistent with the variable that is going fill in this value `s.Bucket.Name`

bcm820: Ah, makes sense.

templates/addons/etcd-backup-restore/sse-key-secret.yaml
```diff
@@ -0,0 +1,29 @@
+---
+apiVersion: v1
+kind: Secret
+type: addons.cluster.x-k8s.io/resource-set
+metadata:
+  name: ${CLUSTER_NAME}-ssekey
+stringData:
+  linode-token-region.yaml: |-
```
eljohnson92, line 8: can we rename this to reflect the different secret

templates/addons/etcd-backup-restore/etcd-backup-restore.yaml
```diff
@@ -0,0 +1,135 @@
...
+              valueFrom:
+                secretKeyRef:
+                  name: sse-key
+                  key: sse-key
+          containers:
+          - name: backup-restore
+            command:
+            - /etcdbrctl
+            - snapshot
+            - --schedule=0 0 * * *
```
eljohnson92, line 67: do we want to make these schedules optional env variables to start that default to these crons?

amold1: done

templates/addons/etcd-backup-restore/sse-key-secret.yaml
amold1: done

controller/linodeobjectstoragebucket_controller_test.go
```diff
@@ -24,6 +24,7 @@ import (
 
 	"github.com/linode/linodego"
 	"go.uber.org/mock/gomock"
+	"gopkg.in/yaml.v2"
```
bcm820, line 27: Can you use `sigs.k8s.io/yaml` instead? It has better support for parsing nested fields.

bcm820 (dismissed):

bcm820 (approved):

PR #211, opened 2024-03-25
add support for stackscript based bootstrapping

eljohnson92:
**What type of PR is this?**
/kind feature

This PR introduced the option to use a stackscript instead of cloud-init to bootstrap a cluster. cloud-init support is currently only rolled out to a few datacenters and the Akamai datasource has only been added/backported to a few official Linode OSes. Datacenter capabilities are fairly static so from a deployed DC we can infer whether or not you need stackscript based bootstrapping. For OSes, it is harder to confidently infer this information so a flag `useStackScriptBootstrap` is introduced with the corresponding template variable `USE_STACKSCRIPT_BOOTSTRAP` to enable this functionality if your OS/DC needs it and it is not enabled automatically. Stackscript [1319647](https://cloud.linode.com/stackscripts/1319647) has been introduced specifically for bridging user-data, stackscripts and cloud-init.
Fixes #

**TODOs**:

- [ ] squashed commits
- [x] includes documentation
- [x] adds unit tests

### DIFF
--- a/controller/linodemachine_controller_helpers.go
+++ b/controller/linodemachine_controller_helpers.go
@@ -39,6 +39,7 @@ import (
 
 	infrav1alpha1 "github.com/linode/cluster-api-provider-linode/api/v1alpha1"
 	"github.com/linode/cluster-api-provider-linode/cloud/scope"
+	"github.com/linode/cluster-api-provider-linode/cloud/services"
 	"github.com/linode/cluster-api-provider-linode/util"
 	"github.com/linode/cluster-api-provider-linode/util/reconciler"
 )
@@ -46,11 +47,6 @@ import (
 // Size limit in bytes on the decoded metadata.user_data for cloud-init
 // The decoded user_data must not exceed 16384 bytes per the Linode API
 const maxBootstrapDataBytes = 16384
-
-// Hard code the stack script ID for our CAPI shim stackscript.
-// This is a very specific stackscript which requires exact inputs
-// from this code, so it is unlikely we will want to configure it without a code change.
-const capiStackScriptID = 1319647
 
 func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineScope *scope.MachineScope, tags []string, logger logr.Logger) (*linodego.InstanceCreateOptions, error) {
 	var err error
@@ -98,10 +94,18 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
 	} else {
 		logger.Info(fmt.Sprintf("using StackScripts for bootstrapping. imageMetadataSupport: %t, regionMetadataSupport: %t",
 			imageMetadataSupport, regionMetadataSupport))
+		capiStackScriptID, err := services.EnsureStackscript(ctx, machineScope)
+		if err != nil {
+			return nil, err
+		}
 		createConfig.StackScriptID = capiStackScriptID
+		// ###### WARNING, currently label, region and type are supported as cloud-init variables, any changes ######
+		// any changes to this could be potentially backwards incompatible and should be noted through a backwards incompatible version update #####
+		instanceData := fmt.Sprintf("label: %s\nregion: %s\ntype: %s", machineScope.LinodeMachine.Name, machineScope.LinodeMachine.Spec.Region, machineScope.LinodeMachine.Spec.Type)
+		// ###### WARNING ######
 		createConfig.StackScriptData = map[string]string{
-			"label":    machineScope.LinodeMachine.Name,
-			"userdata": b64.StdEncoding.EncodeToString(bootstrapData),
+			"instancedata": b64.StdEncoding.EncodeToString([]byte(instanceData)),
+			"userdata":     b64.StdEncoding.EncodeToString(bootstrapData),
 		}
 	}
 
--- a/controller/linodemachine_controller_helpers.go
+++ b/controller/linodemachine_controller_helpers.go
@@ -86,17 +86,22 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
 		return nil, err
 	}
 	regionMetadataSupport := slices.Contains(region.Capabilities, "Metadata")
-	if machineScope.LinodeMachine.Spec.UseStackScriptBootstrap || !regionMetadataSupport {
-		logger.Info(fmt.Sprintf("using StackScripts for bootstrapping. useStackScriptBootstrap: %b, regionMetadataSupport: %b",
-			&machineScope.LinodeMachine.Spec.UseStackScriptBootstrap, &regionMetadataSupport))
+	image, err := machineScope.LinodeClient.GetImage(ctx, machineScope.LinodeMachine.Spec.Image)
+	if err != nil {
+		return nil, err
+	}
+	imageMetadataSupport := slices.Contains(image.Capabilities, "cloud-init")
+	if imageMetadataSupport && regionMetadataSupport {
+		createConfig.Metadata = &linodego.InstanceMetadataOptions{
+			UserData: b64.StdEncoding.EncodeToString(bootstrapData),
+		}
+	} else {
+		logger.Info(fmt.Sprintf("using StackScripts for bootstrapping. imageMetadataSupport: %t, regionMetadataSupport: %t",
+			imageMetadataSupport, regionMetadataSupport))
 		createConfig.StackScriptID = capiStackScriptID
 		createConfig.StackScriptData = map[string]string{
 			"label":    machineScope.LinodeMachine.Name,
 			"userdata": b64.StdEncoding.EncodeToString(bootstrapData),
-		}
-	} else {
-		createConfig.Metadata = &linodego.InstanceMetadataOptions{
-			UserData: b64.StdEncoding.EncodeToString(bootstrapData),
 		}
 	}
 
--- a/controller/linodemachine_controller_helpers.go
+++ b/controller/linodemachine_controller_helpers.go
@@ -85,10 +85,10 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
 	if err != nil {
 		return nil, err
 	}
-
-	if machineScope.LinodeMachine.Spec.UseStackScriptBootstrap || !slices.Contains(region.Capabilities, "Metadata") {
-		logger.Info("metadata is not enabled in this region or 'useStackScriptBootstrap' is set to true, " +
-			"using StackScripts for bootstrapping")
+	regionMetadataSupport := slices.Contains(region.Capabilities, "Metadata")
+	if machineScope.LinodeMachine.Spec.UseStackScriptBootstrap || !regionMetadataSupport {
+		logger.Info(fmt.Sprintf("using StackScripts for bootstrapping. useStackScriptBootstrap: %b, regionMetadataSupport: %b",
+			&machineScope.LinodeMachine.Spec.UseStackScriptBootstrap, &regionMetadataSupport))
 		createConfig.StackScriptID = capiStackScriptID
 		createConfig.StackScriptData = map[string]string{
 			"label":    machineScope.LinodeMachine.Name,
--- a/controller/linodemachine_controller_helpers.go
+++ b/controller/linodemachine_controller_helpers.go
@@ -23,6 +23,7 @@ import (
 	"encoding/gob"
 	"errors"
 	"fmt"
+	"slices"
 	"sort"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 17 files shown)


docs/src/topics/getting-started.md
```diff
@@ -31,6 +31,19 @@ export LINODE_TOKEN=<your linode PAT>
 export LINODE_CONTROL_PLANE_MACHINE_TYPE=g6-standard-2
 export LINODE_MACHINE_TYPE=g6-standard-2
 ```
+## Specifying the region and image
+
+### Image
+The default image is linode/ubuntu22.04 if nothing is specified. This can be overridden, but currently only ubuntu is
```
AshleyDumaine, line 37: There should be some other supported versions too and Debian 11:
```
linode-cli images ls | grep cloud-init
│ linode/debian11            │ Debian 11                  │ Debian    │                           │ True      │ 1500  │ available │ cloud-init   │
│ linode/ubuntu20.04         │ Ubuntu 20.04 LTS           │ Ubuntu    │                           │ True      │ 2000  │ available │ cloud-init   │
│ linode/ubuntu22.04         │ Ubuntu 22.04 LTS           │ Ubuntu    │                           │ True      │ 3500  │ available │ cloud-init   │
│ linode/ubuntu23.10         │ Ubuntu 23.10               │ Ubuntu    │                           │ True      │ 3500  │ available │ cloud-init   │
```

eljohnson92: the issue is our install scripts for kubeadm are ubuntu specific, let me try and rephrase it to say our kubeadm template only supports ubuntu currently

templates/flavors/base/linodeMachineTemplate.yaml
```diff
@@ -9,6 +9,7 @@ spec:
       image: ${LINODE_OS:="linode/ubuntu22.04"}
       type: ${LINODE_CONTROL_PLANE_MACHINE_TYPE}
       region: ${LINODE_REGION}
+      useStackScriptBootstrap: ${USE_STACKSCRIPT_BOOTSTRAP:""}
```
AshleyDumaine, line 12: should this be `false` instead to make it explicit?

docs/src/topics/getting-started.md
AshleyDumaine: I'm not sure what the timeline is on getting the other Ubuntu versions compatible with cloud-init, doesn't seem like we can just say 20.04 and up:
```
linode-cli images ls | grep ubuntu | sort
│ linode/ubuntu16.04lts      │ Ubuntu 16.04 LTS           │ Ubuntu    │                           │ True      │ 2700  │ available │              │
│ linode/ubuntu18.04         │ Ubuntu 18.04 LTS           │ Ubuntu    │                           │ True      │ 2700  │ available │              │
│ linode/ubuntu20.04         │ Ubuntu 20.04 LTS           │ Ubuntu    │                           │ True      │ 2000  │ available │ cloud-init   │
│ linode/ubuntu22.04         │ Ubuntu 22.04 LTS           │ Ubuntu    │                           │ True      │ 3500  │ available │ cloud-init   │
│ linode/ubuntu22.10         │ Ubuntu 22.10               │ Ubuntu    │                           │ True      │ 3500  │ available │              │
│ linode/ubuntu23.04         │ Ubuntu 23.04               │ Ubuntu    │                           │ True      │ 3500  │ available │              │
│ linode/ubuntu23.10         │ Ubuntu 23.10               │ Ubuntu    │                           │ True      │ 3500  │ available │ cloud-init   │
```
Until then we might want to call out specific versions?

eljohnson92: with the cloud-init shim any version of ubuntu should work, we are just doing some ubuntu specific stuff in our preKubeadm commands so we can't support other OSes for kubeadm

controller/linodemachine_controller_helpers.go
```diff
@@ -74,8 +80,23 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
...
-	createConfig.Metadata = &linodego.InstanceMetadataOptions{
-		UserData: b64.StdEncoding.EncodeToString(bootstrapData),
+
+	region, err := machineScope.LinodeClient.GetRegion(ctx, machineScope.LinodeMachine.Spec.Region)
+	if err != nil {
+		return nil, err
+	}
+
+	if machineScope.LinodeMachine.Spec.UseStackScriptBootstrap || !slices.Contains(region.Capabilities, "Metadata") {
+		logger.Info("metadata is not enabled in this region, using StackScripts for bootstrapping")
```
AshleyDumaine, line 90: We should only log this line if the second conditional evaluates to true

eljohnson92: good call, let me rewrite the message to take into account both conditions

controller/linodemachine_controller_helpers.go
```diff
@@ -74,8 +80,24 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
...
-		UserData: b64.StdEncoding.EncodeToString(bootstrapData),
+
+	region, err := machineScope.LinodeClient.GetRegion(ctx, machineScope.LinodeMachine.Spec.Region)
+	if err != nil {
+		return nil, err
+	}
+
+	if machineScope.LinodeMachine.Spec.UseStackScriptBootstrap || !slices.Contains(region.Capabilities, "Metadata") {
+		logger.Info("metadata is not enabled in this region or 'useStackScriptBootstrap' is set to true, " +
+			"using StackScripts for bootstrapping")
```
AshleyDumaine, line 91:
```suggestion
		logger.Info(
			"using StackScripts for bootstrapping",
			"useStackScriptBootstrap",
			machineScope.LinodeMachine.Spec.UseStackScriptBootstrap,
			"metadataSupport",
			slices.Contains(region.Capabilities, "Metadata"),
		)
```

AshleyDumaine (approved):

controller/linodemachine_controller_helpers.go
```diff
@@ -46,6 +47,11 @@ import (
 // The decoded user_data must not exceed 16384 bytes per the Linode API
 const maxBootstrapDataBytes = 16384
 
+// Hard code the stack script ID for our CAPI shim stackscript.
+// This is a very specific stackscript which requires exact inputs
+// from this code, so it is unlikely we will want to configure it without a code change.
+const capiStackScriptID = 1319647
```
cbzzz, line 53: I would guess this is hard-coded to simplify the user-experience (i.e. toggle a switch to use StackScript configurations). However, I think this decision raises some questions since there is now a unversion-controlled, externally-managed piece of code that has no direct link with CAPL versions. 

How do we handle updates (specifically breaking changes) to this specific StackScript? If we opaquely update this StackScript underneath CAPL users, new StackScript-based cluster installations for consumers using CAPL versions will break. If we decide to just update this reference every time we change the Metadata cloud-config, how do we enforce parity between the StackScript configuration method and the Metadata configuration method (especially during development)?

controller/linodemachine_controller_helpers.go
```diff
@@ -74,8 +80,24 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
 
 		return nil, err
 	}
-	createConfig.Metadata = &linodego.InstanceMetadataOptions{
-		UserData: b64.StdEncoding.EncodeToString(bootstrapData),
+
+	region, err := machineScope.LinodeClient.GetRegion(ctx, machineScope.LinodeMachine.Spec.Region)
+	if err != nil {
+		return nil, err
+	}
```
cbzzz, line 87: nit: Why is an error here allowed to stop the controller from reconciling the creation of a LinodeMachine? If this fails, we can gracefully handle this by just trying the Metadata service anyways and falling back on the StackScripts if it fails.

controller/linodemachine_controller_helpers.go
```diff
@@ -74,8 +80,24 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
...
+	regionMetadataSupport := slices.Contains(region.Capabilities, "Metadata")
+	if machineScope.LinodeMachine.Spec.UseStackScriptBootstrap || !regionMetadataSupport {
```
cbzzz, line 89: We should also to check that the image also support the Metadata service here. Couldn't we check the `capabilities` in the Images List response? https://www.linode.com/docs/api/images/#images-list__response-samples

I see that this is implicitly called it out in the documentation changes (only Ubuntu is supported for now), but for future-proofing, this check should be made and failures (via log or error) communicated to the user to inform them of an invalid configuration.

controller/linodemachine_controller_helpers.go
```diff
@@ -74,8 +80,24 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
...
+			&machineScope.LinodeMachine.Spec.UseStackScriptBootstrap, &regionMetadataSupport))
+		createConfig.StackScriptID = capiStackScriptID
+		createConfig.StackScriptData = map[string]string{
+			"label":    machineScope.LinodeMachine.Name,
+			"userdata": b64.StdEncoding.EncodeToString(bootstrapData),
+		}
+	} else {
+		createConfig.Metadata = &linodego.InstanceMetadataOptions{
+			UserData: b64.StdEncoding.EncodeToString(bootstrapData),
+		}
```
cbzzz, line 100: nit: This structure biases the reader towards StackScript configurations. I think this logic should be reversed. We should try defaulting to using the (idiomatic™️) Metadata service first and then falling back on StackScripts if Metadata is not available.

controller/linodemachine_controller_helpers.go
eljohnson92: we need to know before linode creation if we are using metadata or stackscripts so we can't fail back to anything once the linode is created

controller/linodemachine_controller_helpers.go
eljohnson92: Stackscripts have no concept of versioning. it is likely if we had a breaking interface change to a stackscript with relation to a CAPL change we would have to create a brand new stack script and point the new version of CAPL to that new stackscript

controller/linodemachine_controller_helpers.go
eljohnson92: I did consider doing that at first, a few things I ran into looking into this. Image list/get is not cached by the linode client like the region endpoint is so we would have to make an extra call out to the linode API every time a new machine came up. additionally, even if we did make this call, private user images capabilities do not always reflect whether or not the image actually supports cloud init, they have to set it as a capability at image creation time and cannot update it after the image is created. If you still think it is worth it I can look into adding it in as a default, but I think we should still support overwriting the default

cbzzz: > Image list/get is not cached by the linode client

I don't think we can use the caching provided by the `linodego` client in our current implementation anyways since we create a new client on every reconciliation loop.

> private user images capabilities do not always reflect whether or not the image actually supports cloud init, they have to set it as a capability at image creation time and cannot update it after the image is created.

Could we skip the `capabilities` check for private images then?

eljohnson92: we can check if they happen to have it enabled whether it is public/private and then just let them override it with this flag if it was incorrectly tagged?

eljohnson92: or we could limit the check to just `linode/` images

cbzzz: Actually, thinking more about it, is the core issue here just user misconfiguration? Would this validation be better served in a admission webhook?

templates/flavors/base/linodeMachineTemplate.yaml
```diff
@@ -9,6 +9,7 @@ spec:
       image: ${LINODE_OS:="linode/ubuntu22.04"}
       type: ${LINODE_CONTROL_PLANE_MACHINE_TYPE}
       region: ${LINODE_REGION}
+      useStackScriptBootstrap: ${USE_STACKSCRIPT_BOOTSTRAP:=false}
```
AshleyDumaine, line 12: leftover from before the refactor?

eljohnson92: good catch, updated it

AshleyDumaine (dismissed):
LGTM

controller/linodemachine_controller_helpers.go
cbzzz: nit: Now that we have removed the `LinodeMachine.spec.UseStackScriptBootstrap` setting, can we gracefully handle these two API call errors here instead of just returning them? 

We can default to setting the `createConfig.StackScript*` fields.

AshleyDumaine (approved):
LGTM

PR #216, opened 2024-03-25
Avoid disk resize during Linode creation

bcm820:
**What type of PR is this?**

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests


controller/linodemachine_controller.go
```diff
@@ -310,41 +317,48 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
+	// create the root disk
+	rootDisk, err := machineScope.LinodeClient.CreateInstanceDisk(
+		ctx,
+		linodeInstanceID,
+		linodego.InstanceDiskCreateOptions{
+			Label:           "root",
+			Size:            instanceType.Disk - defaultEtcdDiskSize,
+			Image:           createOpts.Image,
+			RootPass:        createOpts.RootPass,
+			Filesystem:      string(linodego.FilesystemExt4),
```
bcm820, line 347: TODO: Confirm we can default to ext4

controller/linodemachine_controller.go
```diff
@@ -310,41 +317,48 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 	return linodeInstance, nil
 }
 
-func (r *LinodeMachineReconciler) configureDisksControlPlane(
+func (r *LinodeMachineReconciler) configureInstanceProfile(
```
AshleyDumaine, line 320: We might want to keep in `ControlPlane` since this is just for control plane nodes

controller/linodemachine_controller.go
```diff
@@ -310,41 +317,48 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
+			AuthorizedUsers: createOpts.AuthorizedUsers,
+			StackscriptID:   createOpts.StackScriptID,
+			StackscriptData: createOpts.StackScriptData,
+		},
+	)
+	if err != nil {
+		logger.Error(err, "Failed to create root disk")
 
 		return err
 	}
```
AshleyDumaine, line 359: [Similar to the resize wait that I recently added](https://github.com/linode/cluster-api-provider-linode/blob/main/controller/linodemachine_controller.go#L349), we might want to wait for the root and etcd disks to be ready?

bcm820: Ah, sorry I missed that change. I'll add it as well

cloud/services/loadbalancers.go
```diff
@@ -110,11 +110,6 @@ func AddNodeToNB(
 	logger logr.Logger,
 	machineScope *scope.MachineScope,
 ) error {
-	// Update the NB backend with the new instance if it's a control plane node
-	if !kutil.IsControlPlaneMachine(machineScope.Machine) {
-		return nil
-	}
```
bcm820, line 116: nb: I moved this check to the caller. We don't expect the function to be used elsewhere correct?

controller/linodemachine_controller.go
```diff
@@ -253,108 +250,210 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
+		logger.Error(err, "multiple instances found", "tags", tags)
+
+		return ctrl.Result{}, err
+	}
+
+	ok, err := r.checkControlPlaneDiskStatuses(ctx, machineScope, linodeInstance.ID, instanceConfig)
+	if err != nil {
+		logger.Error(err, "Failed to check instance disks statuses")
+
+		return ctrl.Result{RequeueAfter: reconciler.DefaultMachineControllerWaitForControlPlaneDisksDelay}, err
```
AshleyDumaine, line 388: I'm not sure if we necessarily want to requeue if there was an error on the API side unless it was a transient error. We should probably just do `return ctrl.Result{}, err` like in the other error cases for now.

AshleyDumaine (dismissed):

PR #217, opened 2024-03-26
use the cluster-api operator for tilt installs instead of clusterctl

eljohnson92:
**What type of PR is this?**
/kind feature

This PR replaces `clusterctl init ...` local_commands in tilt with usage of the cluster-api operator and installs of the resources directly. The main benefit of doing it this way is we can easily stream container logs and see the health of the various parts of the CAPL installation. this otherwise should be a no-op for other workflows
Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -1,55 +1,133 @@
 load("ext://k8s_attach", "k8s_attach")
+load("ext://helm_resource", "helm_resource", "helm_repo")
+load("ext://namespace", "namespace_create")
+update_settings(k8s_upsert_timeout_secs=60)
 
-docker_build(
-    "docker.io/linode/cluster-api-provider-linode",
-    context = ".",
-    only=("Dockerfile", "Makefile", "vendor","go.mod", "go.sum", "./api", "./cloud","./cmd", "./controller", "./util", "./version"),
-    build_args={'VERSION': os.getenv("VERSION","")},
+helm_repo("capi-operator-repo", "https://kubernetes-sigs.github.io/cluster-api-operator",labels=["helm-repos"])
+helm_repo("jetstack-repo", "https://charts.jetstack.io", labels=["helm-repos"])
+helm_resource(
+    "cert-manager",
+    "jetstack-repo/cert-manager",
+    namespace="cert-manager",
+    resource_deps=["jetstack-repo"],
+    flags=[
+        "--create-namespace",
+        "--set=installCRDs=true",
+        "--set=global.leaderElection.namespace=cert-manager",
+    ],
+    labels=["cert-manager"],
 )
 
-local_resource(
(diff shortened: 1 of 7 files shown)


komer3 (commented):
With this change, we won't be able to use `clusterclt move`. Which can be an issue since people who create a mgmt cluster using `make local-deploy` won't be able to utilize clusterclt functionality completely.

eljohnson92:
@komer I think the operator sets the right labels?
```
  labels:
    cluster.x-k8s.io/provider: bootstrap-k3s
    cluster.x-k8s.io/v1beta1: v1beta1
    clusterctl.cluster.x-k8s.io: ""
 ```

File: Tiltfile [hunk 1]
```diff
@@ -1,53 +1,130 @@
...
+    k8s_resource(
+        new_name="capi-rke2-controller-manager",
+        objects=[
+            "rke2-bootstrap-system:namespace",
+            "rke2-control-plane-system:namespace",
+            "rke2:bootstrapprovider",
+            "rke2:controlplaneprovider",
+        ],
+        resource_deps=["capi-controller-manager"],
+        labels=["CAPI"],
```
komer3, line 92: Question: Do we want to install k3s and rke2 provider by default? I thought we would only want to install it when the env var is set to true

hack/manifests/core.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: operator.cluster.x-k8s.io/v1alpha2
+kind: CoreProvider
+metadata:
+  name: cluster-api
+  namespace: capi-system
+spec:
+  manager:
+    featureGates:
+      ClusterResourceSet: true
+      ClusterTopology: true
```
komer3, line 11: nit: new line

Tiltfile
eljohnson92: oops, good catch, fixed

Tiltfile
```diff
@@ -1,55 +1,132 @@
...
+        "capl-leader-election-role:role",
+        "capl-manager-role:clusterrole",
+        "capl-metrics-reader:clusterrole",
+        "capl-proxy-role:clusterrole",
+        "capl-leader-election-rolebinding:rolebinding",
+        "capl-manager-rolebinding:clusterrolebinding",
+        "capl-proxy-rolebinding:clusterrolebinding",
+        "capl-manager-credentials:secret",
+    ],
+    labels=["CAPL"],
```
komer3, line 131: I think we should add a dependency here to capi controller manager. I noticed that CAPL gets installed first before cert-manager and CAPI CRDs which is causing it to throw errors

komer3 (approved):
LGTM!

PR #143, opened 2024-02-27
[tweak] - Switch linode labels to LinodeMachine Name instead of UID

komer3:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:
This PR switched to using linode machine names instead of UID for less confusion

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #219, opened 2024-03-27
fix: add additional name validation for custom resources

cbzzz:
**What type of PR is this?**
/kind bug

This propagates the label constraints of Linode resources to their associated CAPL CustomResourceDefinitions via the Kubernetes Validation Rules feature. When a custom resource is created, the Kubernetes object name is validated against the label constraints of its backing Linode resources. This allows CAPL-managed resources to maintain a human-readable naming scheme between its Kubernetes representation and the backing Linode resource implementations.

See:
- https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules
- https://github.com/kubernetes/kubernetes/issues/74620

**Testing**
1. Create `Linode*` resources with invalid names:
```sh
## LinodeCluster
$ export NAME="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx" && kubectl apply -f - <<EOF
kind: LinodeCluster
metadata:
  name: ${NAME}
spec:
  region: us-sea
EOF

The LinodeCluster "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx" is invalid: metadata.name: Invalid value: "string": custom validation: linode nodebalancer: labels must be between 3..32 characters

## LinodeMachine
$ export NAME="bad--name" && kubectl apply -f - <<EOF
kind: LinodeMachine
metadata:
    name: ${NAME}
spec:
    region: us-sea
    type: g5-nanode-1
EOF
The LinodeMachine "bad--name" is invalid: metadata.name: Invalid value: "string": custom validation: linode instance: labels cannot have two hyphens (--), underscores (__) or periods (..) in a row

## LinodeVPC
export NAME="bad.name"
kubectl apply -f - <<EOF
kind: LinodeVPC
metadata:
  name: ${NAME}
spec:
  region: us-sea
EOF
The LinodeVPC "bad.name" is invalid: metadata.name: Invalid value: "string": custom validation: linode vpc: labels: can only contain ASCII letters, numbers, and hyphens (-), cannot have two consecutive hyphens (--), regex used for validation is: '^[-[:alnum:]]*$', see: https://www.linode.com/docs/api/vpcs/#vpc-create
```

Fixes #

Linode resources have constraints on the format of their labels, e.g.:
- https://www.linode.com/docs/api/nodebalancers/#nodebalancer-create
- https://www.linode.com/docs/api/linode-instances/#linode-create
- https://www.linode.com/docs/api/vpcs/#vpc-create

**UIDs vs. Names**
Prior to #143, CAPL was using the Kubernetes resource UID as a generalized [injection(?)](https://en.wikipedia.org/wiki/Injective_function) between the "space" of valid Kubernetes resource names ([RFC 1123](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names)) to the "space" of valid Linode labels across various resource types. This provided three important properties:
1. Uniqueness
5. Identity: A Kubernetes resource can be mapped to a Linode resource label via its UID in the ObjectMeta.
6. Formatting: [Kubernetes UIDs](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids) are standardized and well-formatted such that they do not break _any_ existing Linode label constraints.

**Why not validate in the controller?**
It didn't seem as user-friendly to require users to debug through controller logs, resource events, etc. to fix what is essentially a naming error. Kubernetes already enforces metadata validation on resources and we can ratchet off that mechanism to inject in our own additive validation constraints to the built-in ones.

See: https://github.com/kubernetes/kubernetes/issues/74620

**Why not automatically format Linode labels**?
For many Linode resources, labels also have an additional constraint that they _must be unique across a resource type in an account_. Auto-formatting labels could lead to a conflict where two independent Kubernetes resources attempt to use the same label for two independent Linode resources.

Unfortunately, this issue is still occurs if a user creates an object of the same name, but in a different namespace. 😢

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/cloud/services/loadbalancers.go
+++ b/cloud/services/loadbalancers.go
@@ -23,7 +23,7 @@ const (
 func CreateNodeBalancer(ctx context.Context, clusterScope *scope.ClusterScope, logger logr.Logger) (*linodego.NodeBalancer, error) {
 	var linodeNB *linodego.NodeBalancer
 
-	NBLabel := fmt.Sprintf("%s-api-server", clusterScope.LinodeCluster.Name)
+	NBLabel := clusterScope.LinodeCluster.Name
 	clusterUID := string(clusterScope.LinodeCluster.UID)
 	tags := []string{string(clusterScope.LinodeCluster.UID)}
 	listFilter := util.Filter{
@@ -53,9 +53,9 @@ func CreateNodeBalancer(ctx context.Context, clusterScope *scope.ClusterScope, l
 		return &linodeNBs[0], nil
 	}
 
-	logger.Info(fmt.Sprintf("Creating NodeBalancer %s-api-server", clusterScope.LinodeCluster.Name))
+	logger.Info(fmt.Sprintf("Creating NodeBalancer %s", clusterScope.LinodeCluster.Name))
 	createConfig := linodego.NodeBalancerCreateOptions{
-		Label:  util.Pointer(fmt.Sprintf("%s-api-server", clusterScope.LinodeCluster.Name)),
+		Label:  util.Pointer(clusterScope.LinodeCluster.Name),
 		Region: clusterScope.LinodeCluster.Spec.Region,
 		Tags:   tags,
 	}
--- a/config/crd/patches/validation_in_linodeclusters.yaml
+++ b/config/crd/patches/validation_in_linodeclusters.yaml
@@ -1,0 +1,11 @@
+# The following patch adds additional constraints after the built-in name validation for the CRD
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/metadata/properties
+  value:
+    name:
+      type: string
+      x-kubernetes-validations:
+      - rule: 3 <= size(self) && size(self) <= 32
+        message: >-
+          custom validation:
+          linode nodebalancer: labels must be between 3..32 characters
--- a/config/crd/patches/validation_in_linodemachines.yaml
+++ b/config/crd/patches/validation_in_linodemachines.yaml
@@ -1,0 +1,26 @@
+# The following patch adds additional constraints after the built-in name validation for the CRD
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/metadata/properties
+  value:
+    name:
+      type: string
+      x-kubernetes-validations:
+      - rule: 3 <= size(self) && size(self) <= 64
+        message: >-
+          custom validation:
+          linode instance: labels must be between 3..64 characters
+      - rule: self.matches('^[[:alnum:]]([-_.[:alnum:]]+[[:alnum:]])*$')
+        message: >-
+          custom validation:
+          linode instance: labels:
+          must begin and end with an alphanumeric character,
+          may only consist of alphanumeric characters, hyphens (-), underscores (_) or periods (.),
+          cannot have two hyphens (--), underscores (__) or periods (..) in a row,
+          regex used for validation is: '^[[:alnum:]]([-_.[:alnum:]]+[[:alnum:]])*$',
+          see: https://www.linode.com/docs/api/linode-instances/#linode-create
+      # TODO: Consider combining this into the regex above to minimize time complexity
+      # See: https://github.com/google/cel-spec/blob/master/doc/langdef.md#time-complexity
+      - rule: "!(self.contains('--') || self.contains('__') || self.contains('..'))"
+        message: >-
+          custom validation:
+          linode instance: labels cannot have two hyphens (--), underscores (__) or periods (..) in a row
--- a/config/crd/patches/validation_in_linodevpcs.yaml
+++ b/config/crd/patches/validation_in_linodevpcs.yaml
@@ -1,0 +1,25 @@
+# The following patch adds additional constraints after the built-in name validation for the CRD
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/metadata/properties
+  value:
+    name:
+      type: string
+      x-kubernetes-validations:
+      - rule: 1 <= size(self) && size(self) <= 64
+        message: >-
+          custom validation:
+          linode vpc: labels must be between 1..64 characters
+      - rule: self.matches('^[-[:alnum:]]*$')
+        message: >-
+          custom validation:
+          linode vpc: labels:
+          can only contain ASCII letters, numbers, and hyphens (-),
+          cannot have two consecutive hyphens (--),
+          regex used for validation is: '^[-[:alnum:]]*$',
+          see: https://www.linode.com/docs/api/vpcs/#vpc-create
+      # TODO: Consider combining this into the regex above to minimize time complexity
+      # See: https://github.com/google/cel-spec/blob/master/doc/langdef.md#time-complexity
+      - rule: "!self.contains('--')"
+        message: >-
+          custom validation:
+          linode vpc: labels cannot have two consecutive hyphens (--)
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -36,6 +36,27 @@ patches:
 #- path: patches/cainjection_in_linodeobjectstoragebuckets.yaml
 #+kubebuilder:scaffold:crdkustomizecainjectionpatch
 
+# [VALIDATION]
+# patches here are for additional validation for each CRD
+- target:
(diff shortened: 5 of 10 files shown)


config/crd/patches/validation_in_linodeclusters.yaml
```diff
@@ -0,0 +1,9 @@
+# The following patch adds additional constraints after the built-in name validation for the CRD
```
eljohnson92, line 1: could we use kubebuilder [validations](https://book.kubebuilder.io/reference/markers/crd-validation) for these usecases, or are they complex enough that they need to be built right into the CRD?

cbzzz: Unfortunately not 😢 See 0d41fd3ba77c03a5895ce30d1e9612c8588fcbe7 😢

eljohnson92: oh, is that because we are validating the names of the resources and not some other spec field?

cloud/services/loadbalancers.go
```diff
@@ -23,7 +23,7 @@ const (
 func CreateNodeBalancer(ctx context.Context, clusterScope *scope.ClusterScope, logger logr.Logger) (*linodego.NodeBalancer, error) {
 	var linodeNB *linodego.NodeBalancer
 
-	NBLabel := fmt.Sprintf("%s-api-server", clusterScope.LinodeCluster.Name)
```
eljohnson92, line 26: is the reason for this change so cluster names have the full 32 Chars for the nodebalancer name?

cbzzz: Yes, it allows clusters to use the full 32-character limit for NodeBalancer labels: https://www.linode.com/docs/api/nodebalancers/#nodebalancer-create

config/crd/patches/validation_in_linodeclusters.yaml
cbzzz: Pretty much. I was able to "hack" around this limitation, but it breaks the validation messaging from the API server: https://kubernetes.slack.com/archives/CAR30FCJZ/p1711475793972459

config/crd/patches/validation_in_linodemachines.yaml
```diff
@@ -0,0 +1,17 @@
...
+      - rule: 3 <= size(self) && size(self) <= 64
+        message: linode instance labels must be between 3..64 characters
+      # TODO: Combine these into one regex to validate ALL constraints on Linode instance labels at once?
+      # https://www.linode.com/docs/api/linode-instances/#linode-create
+      # - rule: self.contains('^[a-z0-9](.*[a-z0-9])*$')
+      #   message: linode instance labels must begin and end with an alphanumeric character
+      # - rule: self.contains('^[-_.[:alnum:]]*$')
+      #   message: linode instance labels may only consist of alphanumeric characters, hyphens (-), underscores (_) or periods (.)
+      - rule: "!(self.contains('--') || self.contains('__') || self.contains('..'))"
+        message: linode instance labels cannot have two hyphens (--), underscores (__) or periods (..) in a row
```
eljohnson92, line 17: label is a bit of an overloaded term here that I'm not sure will be completely obvious to end users how it correlates to the `name` of a resource. could we possibly reword these to mention it is just a requirement on the naming of the resource itself?

config/crd/patches/validation_in_linodemachines.yaml
```diff
@@ -0,0 +1,17 @@
+# The following patch adds additional constraints after the built-in name validation for the CRD
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/metadata/properties
+  value:
+    name:
+      type: string
+      x-kubernetes-validations:
+      - rule: 3 <= size(self) && size(self) <= 64
+        message: linode instance labels must be between 3..64 characters
+      # TODO: Combine these into one regex to validate ALL constraints on Linode instance labels at once?
```
eljohnson92, line 10: would these commented-out rules be covered by the k8s default `metadata.name` validation already?

cbzzz: Not necessarily. For example, [RFC 1123](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names) allows double-hypens (which is disallowed by some Linode resource labels).

eljohnson92: yeah, I'm talking specifically lines 11-15 here

cbzzz: Yes, they should be. However, I opted to add in a 1:1 mapping with all the Linode label validation rules just to make this complete in itself and easier to read.

config/crd/patches/validation_in_linodemachines.yaml
cbzzz: If it helps, the formatting of this validation within the CRD means that API server will automatically reference the `metadata.name` field in the error message like so:
```sh
Error from server (Invalid): error when creating "STDIN": LinodeMachineTemplate.infrastructure.cluster.x-k8s.io "bad--name-control-plane" is invalid: metadata.name: Invalid value: "string": linode instance labels cannot have two hyphens (--), underscores (__) or periods (..) in a row
Error from server (Invalid): error when creating "STDIN": LinodeMachineTemplate.infrastructure.cluster.x-k8s.io "bad--name-md-0" is invalid: metadata.name: Invalid value: "string": linode instance labels cannot have two hyphens (--), underscores (__) or periods (..) in a row
``` 

I can still change this though if you feel it's unclear 🙂

config/crd/patches/validation_in_linodemachines.yaml
eljohnson92: that is fair, if we are going to leave it commented out could we just write something describe why we are keeping them in there but not enabling them?

config/crd/patches/validation_in_linodemachines.yaml
eljohnson92: that is fairly clear, I am still on the fence on using the label terminology but it is more of a nit at this point so I'm good with that not being a blocking concern

config/crd/patches/validation_in_linodemachines.yaml
cbzzz: I've cleaned up the commented-out rules and updated the error messages. I've also updated the "Testing" section of the Description to better show what the error messages from the API server would look like.

Technically, this rule in the `LinodeMachine` and `LinodeMachineTemplate` definition is covered by the default validation, but I've left it in for clarity and context: 
```yaml
      - rule: self.matches('^[[:alnum:]]([-_.[:alnum:]]+[[:alnum:]])*$')
        message: >-
          custom validation:
          linode instance: labels:
          must begin and end with an alphanumeric character,
          may only consist of alphanumeric characters, hyphens (-), underscores (_) or periods (.),
          cannot have two hyphens (--), underscores (__) or periods (..) in a row,
          regex used for validation is: '^[[:alnum:]]([-_.[:alnum:]]+[[:alnum:]])*$',
          see: https://www.linode.com/docs/api/linode-instances/#linode-create
```

(and in case we would like to optimize this in the future and combine all the rules into one regex idk 🤷‍♀️)

eljohnson92 (approved):

config/crd/patches/validation_in_linodevpcs.yaml
```diff
@@ -0,0 +1,25 @@
...
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/metadata/properties
+  value:
+    name:
+      type: string
+      x-kubernetes-validations:
+      - rule: 1 <= size(self) && size(self) <= 64
+        message: >-
+          custom validation:
+          linode vpc: labels must be between 3..64 characters
```
komer3, line 11:
```suggestion
          linode vpc: labels must be between 1..64 characters
```
You might have missed this :)

cbzzz: 👁️👄👁️ Thanks!

komer3 (approved):
LGTM!

eljohnson92 (dismissed):

komer3 (approved):

PR #218, opened 2024-03-27
Use URL from linodego create response for etcd-backup endpoint

amold1:
**What type of PR is this?**
Use URL from linodego create response for etcd-backup endpoint

**What this PR does / why we need it**:
This will allow us to use the hostname response from creating the Linode OBJ bucket as the endpoint for the etcd-backup-restore

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests
- [x] adds or updates e2e tests

PR #221, opened 2024-03-27
Add docs for etcd backups and set default for etcd backup restore image

amold1:
**What type of PR is this?**
Add docs for etcd backups

- [ ] squashed commits
- [x] includes documentation
- [ ] adds unit tests

### DIFF
--- a/templates/addons/etcd-backup-restore/etcd-backup-restore.yaml
+++ b/templates/addons/etcd-backup-restore/etcd-backup-restore.yaml
@@ -80,7 +80,7 @@ data:
             - --cacert=/etc/kubernetes/pki/etcd/ca.crt
             - --cert=/etc/kubernetes/pki/etcd/healthcheck-client.crt
             - --key=/etc/kubernetes/pki/etcd/healthcheck-client.key
-            image: ${ETCDBR_IMAGE:docker.io/amoldeodhar/etcdbrctl:withcredsfile}
+            image: ${ETCDBR_IMAGE}
             securityContext:
               allowPrivilegeEscalation: false
               runAsUser: 0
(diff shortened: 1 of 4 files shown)


templates/addons/etcd-backup-restore/etcd-backup-restore.yaml
```diff
@@ -80,7 +80,7 @@ data:
             - --cacert=/etc/kubernetes/pki/etcd/ca.crt
             - --cert=/etc/kubernetes/pki/etcd/healthcheck-client.crt
             - --key=/etc/kubernetes/pki/etcd/healthcheck-client.key
-            image: ${ETCDBR_IMAGE}
+            image: ${ETCDBR_IMAGE:docker.io/amoldeodhar/etcdbrctl:withcredsfile}
```
eljohnson92, line 83: Can this also be used with the upstream image? or does it only work with the upstream changes?

amold1: snapshotting and uploads should work but SSE will definitely not work with the current upstream image

amold1: Also, given that upstream makes breaking changes often(and they might change the approach we have taken if and when they also incorporate AWS KMS Key support alongwith our SSE-C support), our addon implementation will often be tightly coupled with an upstream image tag

eljohnson92: I am fine with pinning to a specific version, I just want to make sure we aren't using a fork because we can't use the upstream version unless it has the exact spec we have currently just in case it takes a while to merge/things change and we are stuck on a fork. could we point at the latest supported upstream version without SSE for now and add in the SSE version once it is supported?

amold1: I've reverted this change and moved it to [this commit](https://github.com/linode/cluster-api-provider-linode/pull/218/commits/1799ebe29cd4c0c2f23923328d861655e2ab1e30) on [PR-218](https://github.com/linode/cluster-api-provider-linode/pull/218). I've tested that we can use that image and the snapshot is uploaded unencrypted.

eljohnson92 (approved):

PR #241, opened 2024-04-05
[feat] Create new dual-stack flavor

AshleyDumaine:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**: Adds a new dual-stack flavor to enable ipv6 in kubeadm clusters. Also creates a different ipv6 HelmChartProxy for cilium since enabling ipv6 on cilium without the correct kubeadm settings to support ipv6 will keep cilium from starting properly.

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

### DIFF
--- a/templates/common-init-files/secret.yaml
+++ b/templates/common-init-files/secret.yaml
@@ -22,6 +22,7 @@ stringData:
     net.bridge.bridge-nf-call-iptables  = 1
     net.bridge.bridge-nf-call-ip6tables = 1
     net.ipv4.ip_forward                 = 1
+    net.ipv6.conf.all.forwarding        = 1
   kubeadm-pre-init.sh: |
     #!/bin/bash
     set -euo pipefail
--- a/templates/flavors/dual-stack/kustomization.yaml
+++ b/templates/flavors/dual-stack/kustomization.yaml
@@ -1,0 +1,42 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../default
+
+patches:
+  - target:
+      group: cluster.x-k8s.io
+      version: v1beta1
+      kind: Cluster
+    patch: |-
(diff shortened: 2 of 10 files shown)


templates/common-init-files/secret.yaml
```diff
@@ -22,6 +22,7 @@ stringData:
     net.bridge.bridge-nf-call-iptables  = 1
     net.bridge.bridge-nf-call-ip6tables = 1
     net.ipv4.ip_forward                 = 1
+    net.ipv6.conf.all.forwarding        = 1
```
eljohnson92, line 25: can this be set if dual-stack is not enabled?

AshleyDumaine: Yeah, I figured to add it regardless since we already set `net.bridge.bridge-nf-call-ip6tables = 1` regardless of dual-stack as well

rahulait (approved):
Overall, the changes look good to me.

templates/flavors/dual-stack/kustomization.yaml
```diff
@@ -0,0 +1,42 @@
...
+      group: cluster.x-k8s.io
+      version: v1beta1
+      kind: Cluster
+    patch: |-
+      apiVersion: cluster.x-k8s.io/v1beta1
+      kind: Cluster
+      metadata:
+        name: ${CLUSTER_NAME}
+        labels:
+          ipv6: true
```
rahulait, line 17: Should this be `ipv6: "true"` as in addons/cilium/cilium-ipv6.yaml or without quotes here? Didn't install it, hence checking.

AshleyDumaine: Yeah I had that at one point on a different branch, updated 👍

rahulait (approved):

eljohnson92 (approved):

PR #242, opened 2024-04-08
[fix] Requeue during machine preflight

bcm820:
**What type of PR is this?**

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/cloud/scope/common.go
+++ b/cloud/scope/common.go
@@ -14,9 +14,7 @@ import (
 	"github.com/linode/cluster-api-provider-linode/version"
 )
 
-type clientOption func(*linodego.Client)
-
-func CreateLinodeClient(apiKey string, opts ...clientOption) (*linodego.Client, error) {
+func CreateLinodeClient(apiKey string) (*linodego.Client, error) {
 	if apiKey == "" {
 		return nil, errors.New("missing Linode API key")
 	}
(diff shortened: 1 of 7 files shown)


cloud/scope/common.go
```diff
@@ -14,7 +14,9 @@ import (
 	"github.com/linode/cluster-api-provider-linode/version"
 )
 
-func CreateLinodeClient(apiKey string) (*linodego.Client, error) {
+type clientOption func(*linodego.Client)
+
+func CreateLinodeClient(apiKey string, opts ...clientOption) (*linodego.Client, error) {
```
nesv, line 19: I'm not a huge fan of the way `clientOption` isn't exported despite it being used and present in the package's exported API. This change also seems disproportionate to the (presumably-intended) effect.

In the one case in this PR where an option is provided &mdash; `cloud/scope/machine.go:85-87` &mdash; it is being used to disable retries by calling `(*linode.Client).SetRetryCount(0)`. Can that lone call be moved into `CreateLinodeClient` (thus disabling retries for all clients instantiated through this function), or can it be directly called on the `*linode.Client` returned from this function in the one place it's needed?

nesv (approved):
LGTM! Thank you, @bcm820!

cloud/scope/common.go
bcm820: I've observed several other failing E2E suites, but haven't looked into them enough to know the impact. It definitely does seem that linodemachine controller relies on retries the most.

PR #247, opened 2024-04-09
Add tests for LinodeCluster controller

amold1:
**What type of PR is this?**
This PR adds tests for the Linode Cluster Controller

To improve test coverage

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests


controller/suite_test.go
```diff
@@ -52,12 +59,49 @@ func TestControllers(t *testing.T) {
...
+	gopath := envOr("GOPATH", build.Default.GOPATH)
+	return filepath.Join(gopath, "pkg", "mod", "sigs.k8s.io", fmt.Sprintf("cluster-api@v%s", clusterAPIVersion), "config", "crd", "bases")
+}
+
+func envOr(envKey, defaultValue string) string {
+	if value, ok := os.LookupEnv(envKey); ok {
+		return value
+	}
+	return defaultValue
+}
```
bcm820, line 89: Should we just hardcode the directory i.e. `cluster-api@v1.6.3`? It'd be easy enough to change when we upgrade cluster-api in go.mod.

amold1: Yeah we can certainly do that I have no strong opinions either way. Just to make sure we are clear, you are proposing we do something like this?
```
func getFilePathToCAPICRDs() string {
	gopath := os.LookupEnv("GOPATH", build.Default.GOPATH)
	return filepath.Join(gopath, "pkg", "mod", "sigs.k8s.io", "cluster-api@v1.6.3", "config", "crd", "bases")
}
```

controller/linodecluster_controller_test.go
```diff
@@ -0,0 +1,357 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	infrav1 "github.com/linode/cluster-api-provider-linode/api/v1alpha1"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("lifecycle", Ordered, Label("cluster", "lifecycle"), func() {
+	var mockCtrl *gomock.Controller
```
bcm820, line 35: It doesn't appear to be using a mock client here. If no mock client is being configured, the test is making real calls to the Linode API.

controller/linodecluster_controller_test.go
```diff
@@ -0,0 +1,357 @@
...
+				NodeBalancerID: &nodebalancerID,
+			},
+			ControlPlaneEndpoint: clusterv1.APIEndpoint{
+				Host: controlPlaneEndpointHost,
+			},
+		},
+	}
+
+	// Create a recorder with a buffered channel for consuming event strings.
+	recorder := record.NewFakeRecorder(10)
```
bcm820, line 84: No recorder is needed if we don't need to consume and test events.

controller/linodecluster_controller_test.go
amold1: I've removed all mentions of mock client now. We are not using the client because we are only creating CAPL and CAPI clusters and no calls to the Linode API at all.

controller/linodecluster_controller_test.go
amold1: Yep. I've removed them from all but the last one where I have added a check now

controller/suite_test.go
amold1: Maybe
```
f, err := modfile.Parse("go.mod", file_bytes, nil)
if err != nil {
	panic(err)
}

for _, s := range f.Require {
	if strings.Contains(string(s.Mod.Path), "cluster-api") {
		clusterAPIVersion = s.Mod.Version
	}
}
```
and then
```
filepath.Join(gopath, "pkg", "mod", "sigs.k8s.io", fmt.Sprintf("cluster-api@%s", clusterAPIVersion), "config", "crd", "bases")
```

amold1: https://github.com/linode/cluster-api-provider-linode/pull/247/commits/c4424e6f98e1f985760d278285c9b37b89fbc20b

bcm820: This approach LGTM!

bcm820 (approved):

eljohnson92 (approved):

PR #250, opened 2024-04-12
add a more generalized spec for modifying linode disks

### LINKED ISSUE #152: Allow etcd partition size to be configurable
Following up on https://github.com/linode/cluster-api-provider-linode/pull/140 we should allow the user to specify the size of their etcd disk in case they need something larger than the default. 

This will have to take into account the storage limits on the particular instance size selected for control plane nodes, possibly throwing a validation error if the requested size exceeds the plan limits plus space needed for the root and swap disks.

eljohnson92:
**What type of PR is this?**
/kind feature

This PR adds two new interfaces, `osDisk` and `dataDisks`. `osDisk` will force the root os disk to be a certain size if set, otherwise the root disk will use any space available to the linode that is not used by `dataDisks`. 

`dataDisks` adds support for carving out part of the linode total disk size for explicitly separate disks such as etcd. these can be assigned to any of the 7(sdb-sdh) additional devices on a linode.
Fixes #152 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


api/v1alpha1/linodemachine_types.go
```diff
@@ -75,6 +81,17 @@ type LinodeMachineSpec struct {
 	CredentialsRef *corev1.SecretReference `json:"credentialsRef,omitempty"`
 }
 
+// InstanceDisk InstanceDisks defines a list of disks to use for an instance
+type InstanceDisk struct {
+	// DiskID is the linode assigned ID of the disk
+	DiskID int `json:"diskID,omitempty"`
+	// SizeMB is the size in MB of the disk
+	// +kubebuilder:validation:Required
+	SizeMB int `json:"sizeMB"`
```
nesv, line 90: Instead of embedding the value's units in the field name, maybe use a [resource.Quantity](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity) instead.

controller/linodemachine_controller.go
```diff
@@ -369,41 +367,55 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
 	linodeInstanceID int,
 ) error {
-	if !kutil.IsControlPlaneMachine(machineScope.Machine) {
+	if machineScope.LinodeMachine.Spec.DataDisks == nil && machineScope.LinodeMachine.Spec.OSDisk != nil {
 		return nil
 	}
 
-	if err := r.resizeRootDisk(ctx, logger, machineScope, linodeInstanceID); err != nil {
+	err := r.resizeRootDisk(ctx, logger, machineScope, linodeInstanceID)
+	if err != nil {
```
nesv, line 381: nit:

```suggestion
	if err := r.resizeRootDisk(ctx, logger, machineScope, linodeInstanceID); err != nil {
```

controller/linodemachine_controller.go
```diff
@@ -369,41 +367,55 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
+				linodego.InstanceDiskCreateOptions{
+					Label:      label,
+					Size:       disk.SizeMB,
+					Filesystem: string(linodego.FilesystemExt4),
+				},
+			)
+			if err != nil {
+				logger.Error(err, "Failed to create disk", "DiskLabel", label)
+
+				conditions.MarkFalse(machineScope.LinodeMachine, ConditionPreflightAdditionalDisksCreated, string(cerrs.CreateMachineError), clusterv1.ConditionSeverityWarning, err.Error())
```
nesv, line 406: nit: line length consistency

```suggestion
				conditions.MarkFalse(
					machineScope.LinodeMachine, 
					ConditionPreflightAdditionalDisksCreated, 
					string(cerrs.CreateMachineError), 
					clusterv1.ConditionSeverityWarning,
					err.Error(),
				)
```

controller/linodemachine_controller.go
```diff
@@ -446,8 +457,17 @@ func (r *LinodeMachineReconciler) resizeRootDisk(
...
+		additionalDiskSize := 0
+		for _, disk := range machineScope.LinodeMachine.Spec.DataDisks {
+			additionalDiskSize += disk.SizeMB
+		}
+		diskSize := rootDisk.Size - additionalDiskSize
+		if machineScope.LinodeMachine.Spec.OSDisk != nil {
+			diskSize = machineScope.LinodeMachine.Spec.OSDisk.SizeMB
+		}
+
+		if err = machineScope.LinodeClient.ResizeInstanceDisk(ctx, linodeInstanceID, rootDiskID, diskSize); err != nil && !linodego.ErrHasStatus(err, linodeBusyCode) {
```
nesv, line 470: nit: try to not rely on `err` being in the larger scope, when it can be safely scoped to the `if` block

```suggestion
		if err := machineScope.LinodeClient.ResizeInstanceDisk(ctx, linodeInstanceID, rootDiskID, diskSize); err != nil && !linodego.ErrHasStatus(err, linodeBusyCode) {
```

Also, super awesome seeing you use `ErrHasStatus` :grin:

controller/linodemachine_controller.go
```diff
@@ -460,7 +480,7 @@ func (r *LinodeMachineReconciler) resizeRootDisk(
 
 	// wait for the disk to resize
 	if _, err := machineScope.LinodeClient.WaitForInstanceDiskStatus(ctx, linodeInstanceID, rootDiskID, linodego.DiskReady, defaultResizeWaitSeconds); err != nil {
-		logger.Error(err, fmt.Sprintf("Failed to resize root disk within resize timeout of %d seconds", defaultResizeWaitSeconds))
+		logger.Info(fmt.Sprintf("Failed to resize root disk within resize timeout of %d seconds, requeuing", defaultResizeWaitSeconds))
```
nesv, line 483: Don't put variable information into log messages. It hurts the cardinality when ingested into a logging system.

```suggestion
		logger.Info("Timed out resizing root disk",
			"timeout", defaultResizeWaitSeconds,
			"reqeue", true,
		)
```

Even though it would be a larger change, I would _strongly_ recommend using [time.Duration](https://pkg.go.dev/time#Duration) instead of doing a numeric value with the units embedded in the name.

controller/linodemachine_controller.go
```diff
@@ -473,6 +493,59 @@ func (r *LinodeMachineReconciler) resizeRootDisk(
...
+			instanceConfig.SDF = ptr.To(linodego.InstanceConfigDevice{DiskID: disk.DiskID})
+		case "sdg":
+			instanceConfig.SDG = ptr.To(linodego.InstanceConfigDevice{DiskID: disk.DiskID})
+		case "sdh":
+			instanceConfig.SDH = ptr.To(linodego.InstanceConfigDevice{DiskID: disk.DiskID})
+		default:
+			err := fmt.Errorf("unknown device name %s", deviceName)
+			return err
+		}
+	}
```
nesv, line 544: nit: repetitive

```suggestion
	for deviceName, disk := range instanceDisks {
		dev := linodego.InstanceConfigDevice{
			DiskID: disk.DiskID,
		}
		switch deviceName {
		case "sdb":
			instanceConfig.SDB = &dev
		case "sdc":
			instanceConfig.SDC = &dev
		case "sdd":
			instanceConfig.SDD = &dev
		case "sde":
			instanceConfig.SDE = &dev
		case "sdf":
			instanceConfig.SDF = &dev
		case "sdg":
			instanceConfig.SDG = &dev
		case "sdh":
			instanceConfig.SDH = &dev
		default:
			return fmt.Errorf("unknown device name: %q", deviceName)
		}
	}
```

In the error message at the end, I suggest the use of the `%q` formatting verb, since it will wrap the string in quotes, which can be helpful if `deviceName == ""` (an empty string).

api/v1alpha1/linodemachine_types.go
```diff
@@ -64,6 +64,12 @@ type LinodeMachineSpec struct {
 	Tags []string `json:"tags,omitempty"`
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	FirewallID int `json:"firewallID,omitempty"`
+	// OSDisk is configuration for the root disk the includes the OS,
```
amold1, line 67: nit
```suggestion
	// OSDisk is configuration for the root disk that includes the OS,
```

amold1, line 67: Should these two new specs have the `+optional` marker?

api/v1alpha1/linodemachine_types.go
```diff
@@ -75,6 +81,17 @@ type LinodeMachineSpec struct {
 	CredentialsRef *corev1.SecretReference `json:"credentialsRef,omitempty"`
 }
 
+// InstanceDisk InstanceDisks defines a list of disks to use for an instance
```
amold1, line 84: nit
```suggestion
// InstanceDisk defines a list of disks to use for an instance
```

nesv (dismissed):
LGTM!

bcm820 (approved):

api/v1alpha1/linodemachine_types.go
eljohnson92: I believe kubebuilder defaults to optional by default, at least based on the generated CRD spec on these changes it looks like only sizeMB is required

api/v1alpha1/linodemachine_types.go
eljohnson92: the linode client requires size be in MB. I was thinking we explicitly call the unit out here(now in GB) so that we can simplify the conversions we will have to do to use the linode API

api/v1alpha1/linodemachine_types.go
amold1: Ah yes I checked the docs and seems that the `omitempty` is equivalent to `+optional`

amold1 (dismissed):
LGTM

api/v1alpha1/linodemachine_types.go
```diff
@@ -75,6 +81,17 @@ type LinodeMachineSpec struct {
 	CredentialsRef *corev1.SecretReference `json:"credentialsRef,omitempty"`
 }
 
+// InstanceDisk defines a list of disks to use for an instance
+type InstanceDisk struct {
+	// DiskID is the linode assigned ID of the disk
+	DiskID int `json:"diskID,omitempty"`
+	// SizeGB is the size in GB of the disk
+	// +kubebuilder:validation:Required
+	SizeGB int `json:"sizeMB"`
```
rahulait, line 90: variable names with GB/MB seems to not match here.

config/crd/bases/infrastructure.cluster.x-k8s.io_linodemachines.yaml
```diff
@@ -107,6 +107,28 @@ spec:
...
+                  properties:
+                    diskID:
+                      description: DiskID is the linode assigned ID of the disk
+                      type: integer
+                    label:
+                      description: Label is the label for the instance disk, if nothing
+                        is provided it will match the device name
+                      type: string
+                    sizeMB:
+                      description: SizeGB is the size in GB of the disk
```
rahulait, line 123: Same here

api/v1alpha1/linodemachine_types.go
eljohnson92: good catch, this has now been fixed

controller/linodemachine_controller.go
```diff
@@ -50,18 +50,18 @@ import (
 )
 
 const (
-	// default etcd disk size in MB
-	defaultEtcdDiskSize      = 10240
+	linodeBusyCode           = 400
+	sizeToMB                 = 1024
```
rahulait, line 54: Should this be 1000? For disks, I thought 1GiB = 1024 MiB and 1GB = 1000 MB.

eljohnson92: good catch, that sounds correct. fixed

docs/src/topics/disks/data-disks.md
```diff
@@ -0,0 +1,70 @@
+# Data Disks
+This section describes how to specify additional data disks for a linode instance. These disks can use devices `sdb` through `sdh` 
+for a total of 7 disks. The total size of these disks + the OS Disk cannot exceed the linode plan size.
```
AshleyDumaine, line 3: Should probably add a note for how to check the plan size (https://www.linode.com/pricing/#compute-shared)

srust (approved):

api/v1alpha1/linodemachine_types.go
nesv: Just for clarity (and to be regretfully pedantic), by "MB" do you mean _megabytes_ (decimal-based) or _mebibytes_ (binary-based)? I ask because the difference is kind of important. `10MB` (megabytes) is equivalent to `10 000 000 bytes`, where `10Mi` (mebibytes) is `10 486 760 bytes`. The binary SI units (mebibytes) are more accurate when it comes to describing disk space and memory sizes.

If the smallest a disk can be is 1Gi (gibibyte), that would be a separate concern; validating the value.

My suggestion for using resource.Quantity is mostly to reuse a type that is commonly used in Kubernetes to denote a quantity of some resource (in this case, disk). If the value needs to be coerced into another representation (e.g. binary SI &rarr; decimal SI) for the Linode API, resource.Quantity has methods to help with that.

docs/src/topics/disks/os-disk.md
```diff
@@ -0,0 +1,50 @@
...
+spec:
+  template:
+    spec:
+      image: ""
+      rootPass: Akamaipass123!
+      region: us-ord
+      type: g6-standard-4
+      osDisk:
+        label: root-disk
+        sizeGB: 10
```
AshleyDumaine, line 48:
```suggestion
      [...]
      osDisk:
        label: root-disk
        sizeGB: 10
```

api/v1alpha1/linodemachine_types.go
```diff
@@ -86,11 +86,14 @@ type LinodeMachineSpec struct {
...
 	DiskID int `json:"diskID,omitempty"`
-	// Size is the disk size in resource.Quantity notation
+	// Size of the disk in resource.Quantity notation
 	// +kubebuilder:validation:Required
 	Size resource.Quantity `json:"size"`
-	// Label is the label for the instance disk, if nothing is provided it will match the device name
+	// Label for the instance disk, if nothing is provided it will match the device name
 	Label string `json:"label,omitempty"`
+	// Filesystem of disk to provision, the default disk filesystem is "ext4"
+	// +kubebuilder:validation:Enum=raw;swap;ext3;ext4;initrd
```
AshleyDumaine, line 95: Should this field also be immutable?

eljohnson92: good call, yeah I'll update that

AshleyDumaine (dismissed):
LGTM

docs/src/topics/disks/data-disks.md
```diff
@@ -0,0 +1,69 @@
+# Data Disks
+This section describes how to specify additional data disks for a linode instance. These disks can use devices `sdb` through `sdh` 
+for a total of 7 disks. The total size of these disks + the OS Disk cannot exceed the linode plan size. Disk information cannot 
+be updated in place and must be updated by creating new disks.
```
AshleyDumaine, line 4:
```suggestion
for a total of 7 disks. 

~~~admonish warning
There are a couple caveats with specifying disks for a linode instance:
1. The total size of these disks + the OS Disk cannot exceed the Linode plan size.
2. Instance disk configuration is currently immutable via CAPL after the instance is booted.
~~~
```

AshleyDumaine (approved):

PR #256, opened 2024-04-15
[feat] add self-healing flavor and docs on machine health checks

AshleyDumaine:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**: Adds a new self-healing flavor based on kubeadm and docs on machine health checking

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests


templates/addons/machine-health-check/machinehealthcheck.yaml
```diff
@@ -0,0 +1,46 @@
+apiVersion: cluster.x-k8s.io/v1beta1
```
eljohnson92, line 1: do you think this stuff is standard enough we should just add it to the default flavor? or could it be disruptive enough to keep it as an opt-in in a separate flavor?

AshleyDumaine: I believe it should be opt-in rather than included by default since it introduces extra complexity that may not be desired

eljohnson92 (approved):

eljohnson92 (approved):

PR #258, opened 2024-04-17
[feat] - add new etcd-disk flavor

AshleyDumaine:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**: This creates a new flavor to configure etcd to be on a separate disk

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

### DIFF
--- a/templates/flavors/default/kubeadmControlPlane.yaml
+++ b/templates/flavors/default/kubeadmControlPlane.yaml
@@ -47,14 +47,6 @@ spec:
       controllerManager:
         extraArgs:
           cloud-provider: external
-    diskSetup:
-      filesystems:
-        - label: etcd_data
-          filesystem: ext4
-          device: /dev/sdc
-    mounts:
-      - - LABEL=etcd_data
-        - /var/lib/etcd_data
     initConfiguration:
       nodeRegistration:
         kubeletExtraArgs:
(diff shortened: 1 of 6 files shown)


templates/flavors/default/kubeadmControlPlane.yaml
```diff
@@ -36,25 +36,12 @@ spec:
     preKubeadmCommands:
       - /kubeadm-pre-init.sh '{{ ds.meta_data.label }}' "${KUBERNETES_VERSION}"
     clusterConfiguration:
-      etcd:
```
eljohnson92, line 39: do we want to keep this even if we don't have separate disks?

AshleyDumaine: Yeah I think that sounds good to me, I'll add a note in the etcd docs

templates/flavors/etcd-disk/kustomization.yaml
```diff
@@ -0,0 +1,48 @@
...
+      metadata:
+        name: ${CLUSTER_NAME}-control-plane
+      spec:
+        kubeadmConfigSpec:
+          clusterConfiguration:
+            etcd:
+              local:
+                dataDir: /var/lib/etcd_data/etcd
+                extraArgs:
+                  quota-backend-bytes: "8589934592"
```
eljohnson92, line 40: do we still need this section if it is in the default flavor?

AshleyDumaine: We technically don't but I wanted to leave it in for clarity. I'm fine with removing it

eljohnson92 (approved):

PR #259, opened 2024-04-17
standardize k8s version and common setup steps in templates

eljohnson92:
**What type of PR is this?**
/kind cleanup

This PR does some cleanup of our templates in a few ways
* executes common steps such as turning swap off in a consistent way across flavors
* removes providerID specification since the CCM can handle setting this correctly when it is installed
* sets all flavors to use the KUBERNETES_VERSION env variable so the `clusterctl --kubernetes-version` flag will work as expected

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

### DIFF
--- a/templates/flavors/clusterclass-kubeadm/kubeadmConfigTemplate.yaml
+++ b/templates/flavors/clusterclass-kubeadm/kubeadmConfigTemplate.yaml
@@ -28,7 +28,10 @@ spec:
               key: kubeadm-pre-init.sh
           permissions: "0500"
       preKubeadmCommands:
-        - /kubeadm-pre-init.sh '{{ ds.meta_data.label }}' ${KUBERNETES_VERSION}
+        - /kubeadm-pre-init.sh ${KUBERNETES_VERSION}
+        - sed -i '/swap/d' /etc/fstab
+        - swapoff -a
+        - hostnamectl set-hostname '{{ ds.meta_data.label }}' && hostname -F /etc/hostname
       joinConfiguration:
         nodeRegistration:
           kubeletExtraArgs:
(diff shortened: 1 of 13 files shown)


templates/flavors/clusterclass-kubeadm/kubeadmConfigTemplate.yaml
```diff
@@ -28,7 +28,10 @@ spec:
...
           permissions: "0500"
       preKubeadmCommands:
-        - /kubeadm-pre-init.sh '{{ ds.meta_data.label }}' ${KUBERNETES_VERSION}
+        - /kubeadm-pre-init.sh ${KUBERNETES_VERSION}
+        - sed -i '/swap/d' /etc/fstab
+        - swapoff -a
+        - hostnamectl set-hostname '{{ ds.meta_data.label }}' && hostname -F /etc/hostname
       joinConfiguration:
         nodeRegistration:
           kubeletExtraArgs:
```
AshleyDumaine, line 37: should we also remove `provider-id` here?

templates/flavors/default/kubeadmControlPlane.yaml
AshleyDumaine, line 1: same on this file as https://github.com/linode/cluster-api-provider-linode/pull/259/files#r1569316324

templates/flavors/clusterclass-kubeadm/kubeadmConfigTemplate.yaml
eljohnson92: good catch, missed this one

AshleyDumaine (approved):
LGTM, gave it a go on the RKE2 and default flavors and works fine 👍

PR #260, opened 2024-04-18
add sseCustomerAlgorithm to aws config

amold1:
Minor change:

- Upstream made the `sseCustomerAlgorithm` be a necessary config key when using SSE alongwith the `sseCustomerKey`


amold1:
Hardcoded the algorithm because that is the only on AWS supports

nesv (approved):

PR #266, opened 2024-04-18
Adding more a complete E2E test that only runs when a PR is merged to main

komer3:
**What type of PR is this?**

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


e2e/linodecluster-controller/minimal-linodecluster/chainsaw-test.yaml
```diff
@@ -4,6 +4,9 @@ kind: Test
 metadata:
   creationTimestamp: null
   name: minimal-linodecluster
+  # Label to trigger the test on every PR
```
eljohnson92, line 7: while we are labeling these workflows can we also add a label for what component they are part of like `linodecluster:` or `linodevpc:` so we could theoretically have a test filter that just triggers one of these

Makefile
```diff
@@ -147,7 +147,11 @@ test: generate fmt vet envtest ## Run tests.
 
 .PHONY: e2etest
 e2etest: generate local-deploy chainsaw
-	GIT_REF=$(GIT_REF) $(CHAINSAW) test ./e2e
+	GIT_REF=$(GIT_REF) $(CHAINSAW) test ./e2e/default-capl-cluster --assert-timeout 600s --cluster child-cluster=/e2e/default-capl-cluster/child-cluster-kubeconfig.yaml
+
+.PHONY: e2etest-pr
+e2etest-quick: generate local-deploy chainsaw
+	GIT_REF=$(GIT_REF) $(CHAINSAW) test ./e2e --selector quick
```
eljohnson92, line 154: could we add something like an `E2E_SELECTOR` to the e2etest target and just specify something like  `E2E_SELECTOR=quick make e2etest` to reduce the code duplication here?

komer3: Yup I can do that!

e2e/linodecluster-controller/minimal-linodecluster/chainsaw-test.yaml
komer3: I can add them to all the tests

Makefile
eljohnson92: if we have a blank selector here does that default to `all` or `none`? if it defaults to `all` we might be able to just do `-- selector ${E2E_SELECTOR}` on the e2etest target and just use env variables to control the selector

e2e/default-CAPL-cluster/chainsaw-test.yaml
```diff
@@ -0,0 +1,103 @@
...
+            "https://$TARGET_API/$TARGET_API_VERSION/$URI"
+        check:
+          ($error): ~
+          (json_parse($stdout)):
+            results: 0
+  - name: Delete generated child cluster manifest yaml
+    try:
+    - script:
+        content: |
+          rm -f e2e-auto-gen*.yaml
```
eljohnson92, line 101: I think this needs to be updated to match the above file name

e2e/rke2-capl-cluster/chainsaw-test.yaml
```diff
@@ -0,0 +1,105 @@
...
+    try:
+    - script:
+        env:
+        - name: CLUSTER
+          value: ($cluster)
+        - name: NAMESPACE
+          value: ($namespace)
+        content: |
+          set -e
+          clusterctl generate cluster $CLUSTER -n $NAMESPACE --flavor rke2 --kubernetes-version v1.29.1 --infrastructure linode:v0.0.0 --control-plane-machine-count 1 --worker-machine-count 1 > default-cluster.yaml
```
eljohnson92, line 35: this should be `rke2-cluster.yaml`

e2e/linodevpc-controller/minimal-linodevpc/chainsaw-test.yaml
```diff
@@ -3,6 +3,10 @@ apiVersion: chainsaw.kyverno.io/v1alpha1
 kind: Test
 metadata:
   name: minimal-linodevpc
+  # Label to trigger the test on every PR
+  labels:
+    quick:
+    min-vpc:
```
eljohnson92, line 9: I could go either way here but I sort of think we should maybe keep the label generic to the component rather than the test, so all tests that test VPC get a `vpc` label. what are your thoughts on that approach?

Makefile
komer3: So if we have a blank selector, it throws an error. Alternatively, we can pass in the entire flag to the `E2E_SELECTOR` which is a good option. But may we should rename it to `E2E_FLAGS`?

Also, another thing I found was that we can only pass one selector. When I passed in multiple selector labels, it was not able to pick up on the tests. Maybe we can ask them to add the feature to pass in multiple selector labels?

eljohnson92: yeah, I think E2E_FLAGS would work here

e2e/rke2-capl-cluster/assert-capi-resources.yaml
```diff
@@ -0,0 +1,63 @@
...
+  availableReplicas: 1
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: rke2-control-plane-controller-manager
+  namespace: rke2-control-plane-system
+status:
+  availableReplicas: 1
+# ---
```
eljohnson92, line 32: were these commented out on purpose?

.github/workflows/build_test_ci.yml
```diff
@@ -165,11 +166,30 @@ jobs:
...
         key: docker-${{ runner.os }}-${{ hashFiles('go.sum') }}
 
-    - name: E2E test
+    - name: Complete E2E Test
+      if: github.ref == 'refs/heads/main' 
       run: make e2etest
       env:
         GITHUB_TOKEN: ${{ secrets.github_token }}
         LINODE_TOKEN: ${{ secrets.LINODE_TOKEN }}
+        INSTALL_K3S_PROVIDER: true
```
eljohnson92, line 175: could we include these shared env variables at the job scope instead of the step scope?

e2e/rke2-capl-cluster/assert-capi-resources.yaml
komer3: Yeah I was trying to test out if the test would without checking on CRDs. I was seeing a weird error thats why

.github/workflows/build_test_ci.yml
komer3: Hmmm GITHUB_TOKEN and LINODE_TOKEN can be moved to job level but other vars should be kept on step level because we would only be using them when complete e2e tests are triggered on main merge. 

What do you think?

eljohnson92: it looks like these are all duplicated between the two e2e tests
```
        LINODE_REGION: us-sea
        LINODE_CONTROL_PLANE_MACHINE_TYPE: g6-standard-2
        LINODE_MACHINE_TYPE: g6-standard-2
        CLUSTERCTL_CONFIG: /home/runner/work/cluster-api-provider-linode/cluster-api-provider-linode/e2e/clusterctl-config.yaml
```

komer3: Oh that is because I wanted to run all the tests on this PR for testing purpose. I'm gonna remove those before we merge this PR to main. 

Only complete e2e test step should have those env vars

eljohnson92: ah, understood. ok you can ignore this then

e2e/clusterctl-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+  - name: local-linode
+    url: /home/runner/work/cluster-api-provider-linode/cluster-api-provider-linode/infrastructure-local-linode/v0.0.0/infrastructure-components.yaml
+    type: InfrastructureProvider
+  - name: "k3s"
+    url: https://github.com/k3s-io/cluster-api-k3s/releases/latest/bootstrap-components.yaml
+    type: "BootstrapProvider"
+  - name: "k3s"
+    url: https://github.com/k3s-io/cluster-api-k3s/releases/latest/control-plane-components.yaml
+    type: "ControlPlaneProvider"
+  - name: "rke2"
```
eljohnson92, line 11: This should be built into clusterctl, I think we can remove it from here

komer3: I originally didn't have it but for some reason, it was throwing errors when the `clusterctl generate` command was executed. Locally didn't have that issue...

eljohnson92: maybe we need a newer clusterctl version, we can leave it for now

.github/workflows/build_test_ci.yml
```diff
@@ -117,11 +121,30 @@ jobs:
...
+        CLUSTERCTL_CONFIG: /home/runner/work/cluster-api-provider-linode/cluster-api-provider-linode/e2e/clusterctl-config.yaml
+
+    
+    - name: Quick E2E Test
       run: make e2etest
       env:
-        GITHUB_TOKEN: ${{ secrets.github_token }}
-        LINODE_TOKEN: ${{ secrets.LINODE_TOKEN }}
+        # TODO: Switch selector to quick before the PR can be merged
+        E2E_FLAGS: '--assert-timeout 15m0s --selector rke2'
```
amold1, line 140: Should this be `flavor` not `rke2`?

Makefile
```diff
@@ -146,8 +146,12 @@ test: generate fmt vet envtest ## Run tests.
 	rm cover.out.tmp
 
 .PHONY: e2etest
-e2etest: generate local-deploy chainsaw
-	GIT_REF=$(GIT_REF) $(CHAINSAW) test ./e2e
+e2etest: generate local-release local-deploy chainsaw
+    ifeq ($(strip $(E2E_FLAGS)),) ## If E2E_FLAGS not set, run quick e2e tests
+		GIT_REF=$(GIT_REF) $(CHAINSAW) test ./e2e --selector quick
```
eljohnson92, line 151: given we are always including the selector here, can we maybe have ` --selector ${E2E_SELECTOR} ${E2E_FLAGS} ` and do the switches in GHA?

amold1:
There's quite a lot of duplication where we have multiple files creating the same resources in different directories. I'll try to refactor this so we have all of them in one place.

eljohnson92 (approved):

PR #278, opened 2024-04-23
remove wait for disk call and add retry for busy error code

eljohnson92:
**What type of PR is this?**
/kind cleanup

This PR removes the `WaitForInstanceDiskStatus` polling call and instead will re-queue any call and suppress any errors if the Linode API response is `Linode Busy`(400 response code).
Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controller/linodemachine_controller.go
+++ b/controller/linodemachine_controller.go
@@ -51,9 +51,8 @@ import (
 )
 
 const (
-	linodeBusyCode           = 400
-	defaultDiskFilesystem    = string(linodego.FilesystemExt4)
-	defaultResizeWaitSeconds = 5
+	linodeBusyCode        = 400
+	defaultDiskFilesystem = string(linodego.FilesystemExt4)
 
 	// conditions for preflight instance creation
 	ConditionPreflightCreated                clusterv1.ConditionType = "PreflightCreated"
@@ -413,7 +412,9 @@ func (r *LinodeMachineReconciler) createDisks(ctx context.Context, logger logr.L
 			},
 		)
 		if err != nil {
-			logger.Error(err, "Failed to create disk", "DiskLabel", label)
+			if !linodego.ErrHasStatus(err, linodeBusyCode) {
+				logger.Error(err, "Failed to create disk", "DiskLabel", label)
+			}
 
 			conditions.MarkFalse(
 				machineScope.LinodeMachine,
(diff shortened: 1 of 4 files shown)


nesv (approved):

bcm820 (approved):

controller/linodemachine_controller.go
```diff
@@ -490,17 +491,6 @@ func (r *LinodeMachineReconciler) resizeRootDisk(
 		conditions.MarkTrue(machineScope.LinodeMachine, ConditionPreflightRootDiskResizing)
 	}
 
-	// wait for the disk to resize
-	if _, err := machineScope.LinodeClient.WaitForInstanceDiskStatus(ctx, linodeInstanceID, rootDiskID, linodego.DiskReady, defaultResizeWaitSeconds); err != nil {
```
cbzzz, line 494: If we remove this wait time, won't CAPL overly spam the Linode API (especially for larger clusters)?

eljohnson92: We were only waiting 5 seconds here, which is in line with our re-queue, so rather than polling the API and waiting for it to return successfully for up to 5 seconds and then re-queueing if it takes longer we just re-queue, I don't think this should add too much more additional spam to the API, it just moves the retry/wait logic into CAPL instead of a blocking wait in linodego

cbzzz: Would you mind pointing to where CAPL sets the re-queue on errors for `5s`?

It seems like controller-runtime just uses [`k8s.io/client-go/util/workqueue.DefaultControllerRateLimiter`](https://pkg.go.dev/k8s.io/client-go/util/workqueue#DefaultControllerRateLimiter) when setting up each controller: 
- https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.17.3/pkg/reconcile#Reconciler
- https://github.com/kubernetes-sigs/controller-runtime/blob/c25fe2f5ca0f9e10d571f85da61c3d210d6345e7/pkg/controller/controller.go#L147

And this seems to suggest that re-queues on errors are exponential with a minimum delay of `5ms` and a maximum delay of `1000s`: https://github.com/kubernetes/client-go/blob/3aa45779f2e5592d52edf68da66abfbd0805e413/util/workqueue/default_rate_limiters.go#L39-L45)

PR #286, opened 2024-04-24
create a separate workflow for go analyze

eljohnson92:
**What type of PR is this?**
/kind cleanup

This moves the code to do static code analysis into it's own workflow file. This also removes the security/vuln checks as the codeql settings on the repo should also cover that

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/go-analyze.yml
+++ b/.github/workflows/go-analyze.yml
@@ -1,0 +1,77 @@
+name: Go Analyze
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - "*"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  changes:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/go-analyze.yml
```diff
@@ -0,0 +1,77 @@
+name: Go Analyze
+
+on:
```
amold1, line 3: Can we add the paths to ignore like the original file?

eljohnson92: I moved to this approach so we can keep this a required check, but make it a noop for docs only updates. from my testing, there isn't an easy way to have required checks for only some code paths otherwise

amold1: Ok so based on our discussion we want to have this job run every time and let the filter below to ignore the paths

amold1 (approved):

PR #289, opened 2024-04-26
Add vpcless flavor

rahulait:
**What type of PR is this?**

**What this PR does / why we need it**: This creates a new flavor "vpcless" to provision k8s clusters outside of VPC. This is needed for regions which don't have VPC support.

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests


templates/addons/cilium/cilium.yaml
```diff
@@ -6,6 +6,7 @@ spec:
   clusterSelector:
     matchExpressions:
       - {key: ipv6, operator: DoesNotExist}
+      - {key: vxlan, operator: DoesNotExist}
```
AshleyDumaine, line 9: I think we might also need this on the cilium-ipv6.yaml?

templates/flavors/vpcless/kustomization.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../default
+
+patches:
+  - path: deleteLinodeVPC.yaml
```
AshleyDumaine, line 7: could we inline this?
```suggestion
  - target:
      kind: LinodeVPC
    patch: |
      $patch: delete
      apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
      kind: LinodeVPC
      metadata:
        name: ${VPC_NAME:=${CLUSTER_NAME}}
```

templates/addons/cilium/cilium-vxlan.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: addons.cluster.x-k8s.io/v1alpha1
+kind: HelmChartProxy
+metadata:
+  name: cilium-vxlan
```
eljohnson92, line 4: can we add a `-${CLUSTER_NAME}` to the name of this resource and a selector to only apply to the cluster this is associated with?

AshleyDumaine: *slight preference for prefix to follow the convention for the other resources

templates/addons/provider-linode/linode-ccm-vpcless.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: addons.cluster.x-k8s.io/v1alpha1
+kind: HelmChartProxy
+metadata:
+  name: linode-cloud-controller-manager-vpcless
```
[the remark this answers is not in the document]
rahulait: Yeah, I plan to do it as part of separate PR as it will change for every addon.

templates/flavors/vpcless/kustomization.yaml
rahulait: Updated.

templates/addons/cilium/cilium.yaml
rahulait: By adding it to cilium-ipv6, are you trying to say we need to protect from cases where someone specifies something like:
```
    cni: cilium
    ipv6: "true"
    vxlan: "true"
```
I wonder if someone would do this. Let me know if you still think we should add such checks to addons.

AshleyDumaine: yeah we want to make sure only one cilium chart gets applied on a workload cluster, with the above labels I think both cilium-ipv6 and cilium-vxlan would try to be installed by CAAPH onto the workload cluster.

rahulait: sounds good, updated.

templates/flavors/vpcless/kustomization.yaml
```diff
@@ -0,0 +1,44 @@
...
+    patch: |-
+      apiVersion: cluster.x-k8s.io/v1beta1
+      kind: Cluster
+      metadata:
+        name: ${CLUSTER_NAME}
+        labels:
+            cni: cilium
+            vxlan: "true"
+            ccm: linode
+            vpcless: "true"
```
AshleyDumaine, line 42: For some reason I don't see the vpcless label when using dual-stack, we might want to add that in the dual-stack kustomization with a comment that it's there until ipv6 support works for VPCs?

rahulait: Did you do `make local-release` before provisioning cluster? I am able to see the right labels for both vpcless and dual-stack clusters.
```
k get cluster vpcless -o yaml | yq .metadata.labels
ccm: linode
cluster: vpcless
cni: cilium
crs: vpcless-crs
csi: linode
vpcless: "true"
vxlan: "true"
```

```
k get cluster dual-stack -o yaml | yq .metadata.labels
ccm: linode
cluster: dual-stack
cni: cilium
crs: dual-stack-crs
csi: linode
ipv6: "true"
vpcless: "true"
vxlan: "true"
```

AshleyDumaine: Looks like my clusterctl config file needed to be updated since the path changed again for the local release.

rahulait: Hmm, but the above output means CNI won't get installed in dual-stack cluster. We'll need to revert the tags check. 😕

templates/flavors/vpcless/kustomization.yaml
```diff
@@ -0,0 +1,44 @@
...
+      group: cluster.x-k8s.io
+      version: v1beta1
+      kind: Cluster
+    patch: |-
+      apiVersion: cluster.x-k8s.io/v1beta1
+      kind: Cluster
+      metadata:
+        name: ${CLUSTER_NAME}
+        labels:
+            cni: cilium
```
AshleyDumaine, line 39: I think we should be able to remove these extra labels and just use `vpcless` and `vxlan` labels I see the other labels still show up in the generated dual-stack cluster when generating

rahulait: Updated.

AshleyDumaine (approved):
LGTM

eljohnson92 (approved):

PR #291, opened 2024-04-29
feat: linodemachine: add validating admission webhook on create

cbzzz:
**What type of PR is this?**
/kind feature

Adds a validating admission webhook for `infrastructure.cluster.x-k8s.io/v1alpha1/linodemachine` resources on creation.

This initial validation webhook validates the following `LinodeMachine` resource fields on object creation:

| **Field**         | **Validation(s)**                                                             |
|-------------------|-------------------------------------------------------------------------------|
| `.spec.region`    | Valid region                                                                  |
| `.spec.type`      | Valid Linode plan                                                                  |
| `.spec.osDisk`    | `.Size` <= Linode Type size                                                   |
| `.spec.dataDisks` | Disk device slot: valid,unique; num(Disks) <= 7; Sum(`.Size`) <= Linode Type size |

Fixes #

Accompanied by #311 to isolate client refactoring changes: https://github.com/linode/cluster-api-provider-linode/compare/chore.clients...feat.webhook

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Testing**:
1. Create a invalid `LinodeMachine` resource:
```sh
$ kubectl apply -f - <<EOF
kind: LinodeMachine
metadata:
    name: test
spec:
    region: test
    type: g6-nanode-1
    osDisk:
      size: 100G
    dataDisks:
      sdb:
        size: 100G
EOF

The LinodeMachine "test" is invalid:
* spec.region: Not found: "test"
* spec.osDisk: Invalid value: "100G": sum disk sizes exceeds plan storage: 24600M
```


api/v1alpha1/linodemachine_webhook.go
```diff
@@ -0,0 +1,196 @@
...
+func validateDisk(disk *InstanceDisk, path *field.Path, availSize *resource.Quantity) (*resource.Quantity, *field.Error) {
+	if disk == nil {
+		return availSize, nil
+	}
+
+	if disk.Size.Sign() == -1 {
+		return nil, field.Invalid(path, disk.Size.String(), "cannot be negative")
+	}
+	if availSize.Cmp(disk.Size) == -1 {
+		return nil, field.Invalid(path, disk.Size.String(), "sum of disk sizes cannot exceed plan storage")
```
AshleyDumaine, line 190: we should probably also say what the plan storage available is 
```suggestion
		return nil, field.Invalid(path, disk.Size.String(), fmt.Sprintf("sum of disk sizes cannot exceed plan storage of %s", availSize.Size.String()))
```

Makefile
```diff
@@ -291,10 +291,11 @@ $(LOCALBIN):
 ##@ Tooling Binaries:
 # setup-envtest does not have devbox support so always use CACHE_BIN
 
-KUBECTL        ?= kubectl
+KUBECTL        ?= $(LOCALBIN)/kubectl
```
AshleyDumaine, line 294: will this work outside of devbox without a kubectl target?

api/v1alpha1/linodemachine_webhook.go
cbzzz: FWIW, the `availSize` in this helper function is the amount of the _remaining_ storage (i.e. you decrement this value every time you process a disk).

But yeah, you got me 🙈 I didn't want to pass around non-function arguments cause it's un-aesthetic ✨✨

Makefile
cbzzz: Do you mean `kubebuilder`? Yes, it should. 

Actually, I don't even think we have a `kubectl` target in the Makefile right now so this is just a floating variable.

AshleyDumaine: yeah, specifically the kubectl binary since before it was just using the host kubectl wherever that's installed vs installing and using whatever kubectl version in the tool path

cbzzz: Oh I meant like we only have this variable definition in the Makefile to define the local `kubectl` binary path, and not any other rules/targets that do anything with this variable. 

The `kubectl` binary used should be whatever is in your `$PATH`.

api/v1alpha1/linodemachine_webhook.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func (r *LinodeMachine) validateLinodeMachineSpec(ctx context.Context, client LinodeClient) field.ErrorList {
+	var (
+		errs field.ErrorList
+	)
+	_, err := validateRegion(ctx, client, r.Spec.Region, field.NewPath("spec").Child("region"))
+	if err != nil {
+		errs = append(errs, err)
+	}
```
rahulait, line 118:
```suggestion
	_, err := validateRegion(ctx, client, r.Spec.Region, field.NewPath("spec").Child("region")); err != nil {
		errs = append(errs, err)
	}
```

rahulait (dismissed):
Overall, the PR looks good to me.

rahulait (approved):

PR #311, opened 2024-05-08
chore: refactor clients

cbzzz:
**What type of PR is this?**
/kind chore

Accompanies #291 to isolate the following client refactoring changes:

- Moves the client type definitions into a separate `clients` package. This avoids import cycles for client interfaces.
- Consolidates the `Linode*Client` sub-clients into one generic interface (`LinodeClient`).

Fixes #

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


AshleyDumaine (approved):

clients/clients.go
```diff
@@ -10,7 +10,7 @@ import (
 // LinodeClient is an interface that defines the methods that a Linode client must have to interact with Linode.
 // It defines all the functions that are required to create, delete, and get resources
 // from Linode such as object storage buckets, node balancers, linodes, and VPCs.
-type LinodeMachineClient interface {
+type LinodeClient interface {
```
bcm820, line 13: Why is this needed?

cbzzz: I would like to use a generic `LinodeClient` with the webhook tests and the `LinodeMachineClient` just seems to be a generic container embeding the other `Linode*` clients anyways.

bcm820: Ah, the original comments above "LinodeClient is an interface that defines..." are outdated. The `LinodeMachineClient` only merges the client interfaces containing methods used by the machine controller (i.e. excluding the OBJ client). Maybe at this point though it just makes sense to merge all the interfaces into a single `LinodeClient` interface like you've done here? So we'd just be left with a `LinodeClient` and a `K8sClient`.

cbzzz: Ah, I missed that the `LinodeMachineClient` excludes the `LinodeObjectStorageClient` and thought it was just a catch-all definition since it includes stuff like `LinodeNodeBalancerClient`, `LinodeVPCClient.DeleteVPC`, etc. which isn't used by the machine controller at all.

> Maybe at this point though it just makes sense to merge all the interfaces into a single LinodeClient interface like you've done here?

Actually, I wasn't sure about going this far. Are we still enforcing the separation of the different `Linode*` client definitions by the methods called by their respective `Linode*` controllers? I'm fine with merging them all like you suggested or just defining a separate catch-all `LinodeClient` type.

bcm820: I'm kinda leaning towards just having one catch-all client interface, like CCM does. It would clean things up quite a lot. We could still keep the sub-clients, but only as a reference in the `clients` package.

cbzzz: Works for me! 🫡

bcm820 (approved):

PR #320, opened 2024-05-15
feat: linodecluster: add validating admission webhook on create

cbzzz:
**What type of PR is this?**
/kind feature

Adds a validating admission webhook for `infrastructure.cluster.x-k8s.io/v1alpha1/linodecluster` resources on creation.

This initial validation webhook validates the following LinodeCluster resource fields on object creation:

| **Field**      | **Validation(s)** |
| -------------- | ----------------- |
| `.spec.region` | Valid region      |

Fixes #

**TODOs**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Testing**:
1. Create an invalid `LinodeCluster` resource:
```sh
$ kubectl apply -f - <<EOF
kind: LinodeCluster
metadata:
  name: test
spec:
  region: test
EOF

The LinodeCluster "test" is invalid: spec.region: Not found: "test"
```

### DIFF
--- a/api/v1alpha1/linodecluster_webhook.go
+++ b/api/v1alpha1/linodecluster_webhook.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2023 Akamai Technologies, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// log is for logging in this package.
+var linodeclusterlog = logf.Log.WithName("linodecluster-resource")
+
+// SetupWebhookWithManager will setup the manager to manage the webhooks
+func (r *LinodeCluster) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+// TODO(user): EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable deletion validation.
+//+kubebuilder:webhook:path=/validate-infrastructure-cluster-x-k8s-io-v1alpha1-linodecluster,mutating=false,failurePolicy=fail,sideEffects=None,groups=infrastructure.cluster.x-k8s.io,resources=linodeclusters,verbs=create;update,versions=v1alpha1,name=vlinodecluster.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &LinodeCluster{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *LinodeCluster) ValidateCreate() (admission.Warnings, error) {
+	linodeclusterlog.Info("validate create", "name", r.Name)
+
+	// TODO(user): fill in your validation logic upon object creation.
+	return nil, nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *LinodeCluster) ValidateUpdate(old runtime.Object) (admission.Warnings, error) {
+	linodeclusterlog.Info("validate update", "name", r.Name)
+
+	// TODO(user): fill in your validation logic upon object update.
+	return nil, nil
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *LinodeCluster) ValidateDelete() (admission.Warnings, error) {
+	linodeclusterlog.Info("validate delete", "name", r.Name)
+
+	// TODO(user): fill in your validation logic upon object deletion.
+	return nil, nil
+}
--- a/api/v1alpha1/linodecluster_webhook.go
+++ b/api/v1alpha1/linodecluster_webhook.go
@@ -17,11 +17,19 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"context"
+	"slices"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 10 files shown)


api/v1alpha1/linodecluster_webhook.go
```diff
@@ -0,0 +1,104 @@
...
+var linodeclusterlog = logf.Log.WithName("linodecluster-resource")
+
+// SetupWebhookWithManager will setup the manager to manage the webhooks
+func (r *LinodeCluster) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+// TODO(user): EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
eljohnson92, line 45: can we remove these scaffolding todos?

api/v1alpha1/linodecluster_webhook.go
```diff
@@ -0,0 +1,104 @@
...
+	linodeclusterlog.Info("validate create", "name", r.Name)
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultWebhookTimeout)
+	defer cancel()
+
+	return nil, r.validateLinodeCluster(ctx, &defaultLinodeClient)
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *LinodeCluster) ValidateUpdate(old runtime.Object) (admission.Warnings, error) {
```
eljohnson92, line 63: can we add some tests for each of these validate functions? even if it is just confirming they are currently doing nothing?

cbzzz: Unfortunately not. 

1. These `Validate*` methods are functionally similar to the individual `Reconcile` methods - they are _only_ intended to be called by the Controller's Manager. In our case, they're also intended to inject real dependencies (e.g. `linodego.Client` and cannot be mocked.

2. For the methods that do nothing (e.g. `ValidateUpdate`, `ValidateDelete`), they do not get called since the registration of the underlying operations into Kubernetes is configured through a [controller-gen marker](https://github.com/linode/cluster-api-provider-linode/pull/320/files#diff-4df2d4fcaeca43c236826f56e655e92482e2cf05ed11b79c08546e6440bbff77R47-R48). These no op methods are currently only scaffolding functionally. I don't think it would be particularly valuable to add a unit test here.

api/v1alpha1/linodecluster_webhook.go
```diff
@@ -0,0 +1,104 @@
...
+	linodeclusterlog.Info("validate update", "name", r.Name)
+
+	// TODO(user): fill in your validation logic upon object update.
+	return nil, nil
```
AshleyDumaine, line 67: Since we allow the region to be updated, we should probably change this to 
```
return nil, r.validateLinodeCluster(ctx, &defaultLinodeClient)
```

cbzzz: Before we consider these changes, do you know if we have a story for moving clusters to a different region as of now? 

It looks the `spec.region` field in the LinodeMachines themselves immutable and the LinodeCluster controller doesn't even reconcile updates.

api/v1alpha1/linodecluster_webhook_test.go
```diff
@@ -0,0 +1,68 @@
...
+			),
+		),
+		OneOf(
+			Path(Call("invalid region", func(ctx context.Context, mck Mock) {
+				mck.LinodeClient.EXPECT().GetRegion(gomock.Any(), gomock.Any()).Return(nil, errors.New("invalid region")).AnyTimes()
+			})),
+		),
+		Result("error", func(ctx context.Context, mck Mock) {
+			assert.Error(t, cluster.validateLinodeCluster(ctx, mck.LinodeClient))
+		}),
```
amold1, line 66: We should also validate the error messages like we do [here](https://github.com/linode/cluster-api-provider-linode/blob/main/controller/linodecluster_controller_test.go#L122)

cbzzz: If possible, I would prefer to keep these tests generic against the `Linode*`resource interface and decoupled from the the implementation details.

In this case, these error messages are more of a convenience for the user to help them triage the exact validation error rather than semantically meaningful to the controller itself.

Let me know if you'd prefer otherwise though.

api/v1alpha1/linodecluster_webhook.go
```diff
@@ -0,0 +1,104 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	. "github.com/linode/cluster-api-provider-linode/clients"
+)
+
+// log is for logging in this package.
+var linodeclusterlog = logf.Log.WithName("linodecluster-resource")
```
bcm820, line 36: nit: camelCase?

api/v1alpha1/linodecluster_webhook.go
AshleyDumaine: AFAIK there's no ask to support that... *yet*

api/v1alpha1/linodecluster_webhook_test.go
amold1: For this specific implementation, we should be ok with not looking for the error string because the method has just one error check. 

I like doing those checks in cases where we have the same method checking for errors at multiple stages like in our controllers. So we want to make sure that we are checking for the specific error to have occurred rather than just some/any error to have occurred

api/v1alpha1/linodecluster_webhook.go
cbzzz: > maybe make it immutable since the VPC would need to have its region also updated for changing the LinodeCluster region to work as expected when using a VPC

Going through the specifications again, I think this option would be the best at the moment since:
1. The `.spec.Region` of a LinodeVPC is immutable
2. The LinodeCluster controller currently doesn't reconcile `.spec.Region` updates

I think if we do decide to support this feature, we can enable update validation for LinodeClusters then.

api/v1alpha1/linodecluster_webhook.go
```diff
@@ -0,0 +1,102 @@
...
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable updation and deletion validation.
+//+kubebuilder:webhook:path=/validate-infrastructure-cluster-x-k8s-io-v1alpha1-linodecluster,mutating=false,failurePolicy=fail,sideEffects=None,groups=infrastructure.cluster.x-k8s.io,resources=linodeclusters,verbs=create,versions=v1alpha1,name=vlinodecluster.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &LinodeCluster{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *LinodeCluster) ValidateCreate() (admission.Warnings, error) {
+	linodeclusterlog.Info("validate create", "name", r.Name)
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultWebhookTimeout)
```
eljohnson92, line 54: can we move this logic into the validateLinodeCluster function since it can't be tested here?

cbzzz: I think by separating the context like so, we can:
1. More granularly tune and time the context for different webhook operations
2. The tests already provide a test-specific context we can just pass in to `validateLinodeCluster`

AshleyDumaine (approved):

PR #321, opened 2024-05-15
feat: linodevpc: add validating admission webhook on create

cbzzz:
<!-- If this is your first PR, welcome! Please make sure you read the [contributing guidelines](../CONTRIBUTING.md). -->

 <!-- Please label this pull request according to what type of issue you are addressing (see ../CONTRIBUTING.md) -->
**What type of PR is this?**
/kind feature
<!--
Add one of the following kinds:
/kind feature
/kind bug
/kind api-change
/kind cleanup
/kind deprecation
/kind design
/kind documentation
/kind testing
-->

Adds a validating admission webhook for `infrastructure.cluster.x-k8s.io/v1alpha1/linodevpc` resources on creation.

This initial validation webhook validates the following LinodeVPC resource fields on object creation:

| **Field**               | **Validation(s)**                                                                                                                                                                                                                                                                                                                                                                               |
| ----------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `.spec.region`          | - Valid region<br>- [VPC supported](https://www.linode.com/docs/products/networking/vpc/#availability)                                                                                                                                                                                                                                                                                          |
| `.spec.Subnets[].Label` | - 1..64 characters<br>- Unique among the VPC’s Subnets<br>- Can only contain ASCII letters, numbers, and hyphens (`-`)<br>- No two consecutive hyphens (`--`).                                                                                                                                                                                                                                  |
| `.spec.Subnets[].IPv4`  | - IPv4 range in CIDR canonical form<br>- Range belongs to a private address space as defined in [RFC1918](https://datatracker.ietf.org/doc/html/rfc1918)<br>- Prefix length: 1-29<br>- No overlap with [Linode Instance Private IPv4 Address](https://www.linode.com/docs/products/networking/vpc/guides/subnets/#valid-ipv4-ranges) feature<br>- No overlap with other Subnets on the same VPC |

Fixes #

~~[`gopkg.in/addrs.v1`](https://github.com/go-addrs/addrs) da 🐐 no 🧢 frfr 🙏~~
[`go4.org/netipx`](https://tailscale.com/blog/netaddr-new-ip-type-for-go) da 🐐 no 🧢 frfr 🙏

**TODOs**:
<!-- Put an "X" character inside the brackets of each completed task. Some may be optional depending on the PR. -->

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Testing**:
1. Create an invalid `LinodeVPC` resource:
```sh
$ kubectl apply -f - <<EOF
kind: LinodeVPC
metadata:
    name: test
spec:
  region: us-east
  subnets:
  - label: "---"
    ipv4: 192.168.128.0/24
  - ipv4: 9.9.9.9
EOF

The LinodeVPC "test" is invalid:
* spec.region: Invalid value: "us-east": no capability: VPCs
* spec.Subnets[0].Label: Invalid value: "---": cannot contain two consecutive hyphens (--)
* spec.Subnets[0].IPv4: Invalid value: "192.168.128.0/24": range must not overlap with [192.168.128.0/17]
* spec.Subnets[1].Label: Invalid value: "": 1..64 characters
* spec.Subnets[1].IPv4: Invalid value: "9.9.9.9": must be IPv4 range in CIDR canonical form
```


api/v1alpha1/linodevpc_webhook.go
```diff
@@ -0,0 +1,266 @@
...
+		}
+	}
+
+	if len(errs) == 0 {
+		return nil
+	}
+	return errs
+}
+
+// TODO: Replace the OpenAPI schema validation for .metadata.name.
```
eljohnson92, line 195: can you elaborate a little on what this TODO is for?

api/v1alpha1/linodevpc_webhook.go
```diff
@@ -0,0 +1,266 @@
...
+var linodevpclog = logf.Log.WithName("linodevpc-resource")
+
+// SetupWebhookWithManager will setup the manager to manage the webhooks
+func (r *LinodeVPC) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+// TODO(user): EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
eljohnson92, line 84: if we have completed these TODOs can we remove them?

api/v1alpha1/linodevpc_webhook.go
cbzzz: The VPC Subnet label schema is essentially the same as the one for the VPC label itself. We previously [implemented the latter](https://www.github.com/linode/cluster-api-provider-linode/pull/219/commits/cc5e93415b446b2ca257681b0beac87e37622d73#diff-7613e13b49df95927633cbc1b5f25a84eafc7028e0cd6f5e45161e054d6a133e) via the Kubernetes Validation Rules feature onto the CustomResourceDefinitions definitions themselves. Syntactically, it might be better to remove those and consolidate all our custom (non-Kubernetes) validations in one place (the Go API resource definitions under `api/`).

api/v1alpha1/linodevpc_webhook.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	return errs
+}
+
+// TODO: Replace the OpenAPI schema validation for .metadata.name.
+//
+// validateVPCLabel validates a label string is a valid [Linode VPC Label].
+//
+// [Linode VPC Label]: https://www.linode.com/docs/api/vpcs/#vpc-create__request-body-schema
+func validateVPCLabel(label string, path *field.Path) *field.Error {
```
eljohnson92, line 200: could this be done in a similar way to the vpc name? https://github.com/linode/cluster-api-provider-linode/blob/main/config/crd/patches/validation_in_linodevpcs.yaml

eljohnson92: yes, I understand they are different, my question is could we do this validation as part of the CRD validation rule as well or does this need to be a check in the webhook specifically?

api/v1alpha1/linodevpc_webhook.go
```diff
@@ -0,0 +1,264 @@
...
+// TODO(user): change verbs to "verbs=create;update;delete" if you want to enable updation and deletion validation.
+//+kubebuilder:webhook:path=/validate-infrastructure-cluster-x-k8s-io-v1alpha1-linodevpc,mutating=false,failurePolicy=fail,sideEffects=None,groups=infrastructure.cluster.x-k8s.io,resources=linodevpcs,verbs=create,versions=v1alpha1,name=vlinodevpc.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &LinodeVPC{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *LinodeVPC) ValidateCreate() (admission.Warnings, error) {
+	linodevpclog.Info("validate create", "name", r.Name)
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultWebhookTimeout)
```
eljohnson92, line 93: can we move this code into the validateLinodeVPC function since we cannot test this code here?

api/v1alpha1/linodevpc_webhook.go
cbzzz: Perhaps, but I think the reverse (consolidating the validations with the definitions) would be the better approach here.

api/v1alpha1/linodevpc_webhook.go
cbzzz: I believe the convention around contexts is to create the context as early as possible as pass it around?

api/v1alpha1/linodevpc_webhook.go
eljohnson92: can you elaborate a bit on why that would be the better approach? it feels to me like minimizing the amount of code we have to maintain and test by using simple regex and other CRD checks would be a benefit

api/v1alpha1/linodevpc_webhook.go
eljohnson92: that assumes multiple things are using the context though, in this case, `ValidateCreate` is just a wrapper around the `validateLinodeVPC` function

api/v1alpha1/linodevpc_webhook.go
cbzzz: 1. The validations would be "next to" the definitions in the code. This way if our custom resource definition changes or any specific validations on those definitions change, we can match each API version with the appropriate corresponding validations.

2. The Kubernetes Validation Rules feature is a separate mechanism from the validating admission webhooks, which resources would need to pass two validation gates with different rules, messages, etc. One of the goals of this feature(s) was to surface user configuration errors as early and thoroughly as possible. I believe the split implementation goes against this goal.

api/v1alpha1/linodevpc_webhook.go
cbzzz: 1. That seems like an internal implementation detail. I think we should keep the webhook framework simple, yet flexible enough to accommodate for future changes if we do decide to have a separate context for different operations, etc.

2. We do make use of passing different contexts to `validateLinodeVPC` in the unit tests.

api/v1alpha1/linodevpc_webhook.go
eljohnson92: both implementations would return an error immediately to the user right? or do the validation rules return errors in a different way?

api/v1alpha1/linodevpc_webhook.go
eljohnson92: any code directly in the validate{Create,Update,Delete} functions is untestable. keeping as much logic as possible out of these functions and in the functions with the linode clients will keep code more easy to write tests for

api/v1alpha1/linodevpc_webhook.go
cbzzz: The issue here is that it's a _separate_ mechanism and thus they're completely _separate_ validation checks. For example, if the user violated an `x-kubernetes-validations` rule, the API server would return the error early and not call the corresponding validating admission webhook at all.

eljohnson92: ah, so if we put it all in webhooks users would get all of the errors at once rather than having to fix some of them and then moving on to getting the webhook errors?

cbzzz: Yes, I believe that's a better user experience overall (within limitations). For example, we still use things like [controller-gen markers](https://book.kubebuilder.io/reference/markers/crd-validation#crd-validation) to do the bulk of our validations via `x-kubernetes-validations`.

api/v1alpha1/linodevpc_webhook.go
eljohnson92: ah, I now understand what you mean by the different contexts in the unit tests, this sounds good

eljohnson92 (approved):

eljohnson92 (approved):

PR #348, opened 2024-05-29
[fix] Create new ccnp for etcd ports

amold1:
kustomize is currently not detecting the ccnp patch for kubeadm to add etcd ports. This is down to kustomize not using env vars and failing to find a matching resources. Se we are now adding a etcd specific policy for kubeadm

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/templates/flavors/kubeadm/default/allow-etcd-policy.yaml
+++ b/templates/flavors/kubeadm/default/allow-etcd-policy.yaml
@@ -10,7 +10,7 @@ data:
     metadata:
       name: "allow-etcd-policy"
     spec:
-      description: "allow etcd & api server traffic"
+      description: "allow etcd traffic"
       nodeSelector: {}
       ingress:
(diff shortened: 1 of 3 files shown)


rahulait (dismissed):
LGTM

templates/flavors/kubeadm/default/allow-etcd-policy.yaml
```diff
@@ -0,0 +1,38 @@
...
+    metadata:
+      name: "allow-etcd-policy"
+    spec:
+      description: "allow etcd traffic"
+      nodeSelector: {}
+      ingress:
+        - fromEntities:
+            - cluster
+        - fromCIDR:
+            - 10.0.0.0/8
```
rahulait, line 19: No need to add these again

rahulait (approved):

PR #349, opened 2024-05-30
[fix] cert hostpath and kustomize patch & label fixes for flavors

amold1:
cert hostpath and kustomize patch & label fixes for flavors

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


templates/flavors/kubeadm/full/kustomization.yaml
```diff
@@ -41,6 +41,47 @@ patches:
               - label: etcd_data
                 filesystem: ext4
                 device: /dev/sdc
+            partitions:
+              - device: /dev/sdc
+                layout: true
           mounts:
-            - - LABEL=etcd_data
```
AshleyDumaine, line 45: Not sure about this change, I was following https://cluster-api.sigs.k8s.io/tasks/bootstrap/kubeadm-bootstrap/:
> KubeadmConfig.Mounts specifies a list of mount points to be setup.
> ```
> mounts:
> - - LABEL=etcd_disk
>   - /var/lib/etcddisk
>```

templates/flavors/kubeadm/full/kustomization.yaml
```diff
@@ -41,6 +41,47 @@ patches:
               - label: etcd_data
                 filesystem: ext4
                 device: /dev/sdc
+            partitions:
```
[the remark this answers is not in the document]
amold1: Yeah I thought so too but if we deploy a cluster without this change, the cloud-init config that is generated is faulty. We get
```
---
disk_setup: null
fs_setup:
-   device: /dev/sdc
    filesystem: ext4
    label: etcd_data
mounts:
-   - LABEL=etcd_data
    - /var/lib/etcd_data
``` and cloud-init will complain it is invalid. With the above change, we get the correct config
```
---
disk_setup:
    /dev/sdc:
        layout: true
fs_setup:
-   device: /dev/sdc
    filesystem: ext4
    label: etcd_data
mounts:
-   - LABEL=etcd_data
    - /var/lib/etcd_data
```
You can deploy a cluster without this change and ssh to the node and check this with cloud-init and it should give you something like
```
root@kubeadm-full-control-plane-8f6ff:~# cloud-init schema --system
Invalid cloud-config /var/lib/cloud/instances/59513589/cloud-config.txt
Error: Cloud config schema errors: disk_setup: None is not of type 'object'

Error: Invalid cloud-config schema: user-data

root@kubeadm-full-control-plane-8f6ff:~#
```

File: templates/flavors/kubeadm/full/kustomization.yaml (hunk 1 above)
Comment (amold1) [reply templates/flavors/kubeadm/full/kustomization.yaml:45]: Without this change, we noticed that the filesystem was not being mounted and the `var/lib/etcd_data` was on the root disk itself

File: templates/flavors/rke2/full-vpcless/kustomization.yaml [hunk 3]
```diff
@@ -59,5 +71,3 @@ patches:
                   device: /dev/sdc
               mounts:
                 - ["/dev/sdc", "/var/lib/etcd_data"]
-
```
AshleyDumaine, line 62: do we also need to change `diskSetup` to `disk_setup` above?

AshleyDumaine (approved):

PR #355, opened 2024-06-06
[feat]: add konnectivity flavor for kubeadm

rahulait:
This PR adds [konnectivity](https://kubernetes.io/docs/tasks/extend-kubernetes/setup-konnectivity/) kubeadm flavor. Traffic from apiserver to cluster flows over tunnels created by konnectivity-server and konnectivity-agent.

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests


templates/flavors/kubeadm/full-vpcless/kustomization.yaml
```diff
@@ -2,13 +2,80 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
+      - op: remove
+        path: /spec/vpcRef
+  - target:
+      group: controlplane.cluster.x-k8s.io
+      version: v1beta1
+      kind: KubeadmControlPlane
+    patch: |-
+      - op: remove
+        path: /spec/kubeadmConfigSpec/initConfiguration/skipPhases
+  - target:
```
AshleyDumaine, line 68: I think we want to remove this patch now to be in line with changes merged in https://github.com/linode/cluster-api-provider-linode/pull/358

rahulait: updated.

api/v1alpha1/linodecluster_conversion.go
```diff
@@ -36,10 +36,13 @@ func (src *LinodeCluster) ConvertTo(dstRaw conversion.Hub) error {
...
 	dst.Spec.Network = infrastructurev1alpha2.NetworkSpec{
-		LoadBalancerType:              src.Spec.Network.LoadBalancerType,
-		ApiserverLoadBalancerPort:     src.Spec.Network.LoadBalancerPort,
-		NodeBalancerID:                src.Spec.Network.NodeBalancerID,
-		ApiserverNodeBalancerConfigID: src.Spec.Network.NodeBalancerConfigID,
+		LoadBalancerType:                 src.Spec.Network.LoadBalancerType,
+		ApiserverLoadBalancerPort:        src.Spec.Network.LoadBalancerPort,
+		NodeBalancerID:                   src.Spec.Network.NodeBalancerID,
+		ApiserverNodeBalancerConfigID:    src.Spec.Network.NodeBalancerConfigID,
+		Konnectivity:                     false,
```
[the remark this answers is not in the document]
rahulait: updated by adding additionalPorts field.

templates/flavors/kubeadm/konnectivity/allow-konnectivity-port.yaml
```diff
@@ -0,0 +1,33 @@
...
+    spec:
+      description: "allow incoming connections to konnectivity-server from everywhere"
+      nodeSelector: {}
+      ingress:
+        - fromEntities:
+            - world
+          toPorts:
+            - ports:
+                - port: "6443"
+                - port: "8132"
```
rahulait, line 20: We have seen that sometimes konnectivity-cilium-policy gets applied first than the default cilium policies and port 6443 gets blocked (default behavior of cilium to block ports if any ccnp is present). There is no ordering of different clusterresourceset when applied. Due to this, cluster installs get stuck if konnectivity ccnp is applied as one cannot reach to cluster over 6443 port. Hence, here we are explicitly allowing 6443 here so that even if konnectivity policy gets applied first, we still have 6443 port open. I am open to discussions if we want to take a different approach.

rahulait: Updated to use patch instead.

templates/flavors/kubeadm/konnectivity/kustomization.yaml
```diff
@@ -0,0 +1,119 @@
...
+        name: ${CLUSTER_NAME}-control-plane
+      spec:
+        kubeadmConfigSpec:
+          preKubeadmCommands:
+            - /kubeadm-pre-init.sh ${KUBERNETES_VERSION}
+            - sed -i '/swap/d' /etc/fstab
+            - swapoff -a
+            - hostnamectl set-hostname '{{ ds.meta_data.label }}' && hostname -F /etc/hostname
+            - mkdir -p -m 755 /etc/kubernetes/konnectivity
+            - curl -s -L https://raw.githubusercontent.com/linode/konnectivity/876b00f976975833929723ad9e247d0c124124bd/config/egress-selector-configuration.yaml > /etc/kubernetes/konnectivity/egress-selector-configuration.yaml
```
AshleyDumaine, line 26: I know we already merged something like this for the kubeadm base flavor, but why do we want to lock this to a specific commit version instead of `main` or a tag?

rahulait: We want to make sure updates to these scripts don't break things for someone using an older release. So we are using git commit hash which is helping us to pin to a specific tested version (same what a tag would do).

eljohnson92 (approved):

templates/flavors/kubeadm/konnectivity/kustomization.yaml
eljohnson92: out of scope for this PR, but I think we should produce this is a release artifact and eventually reference that directly instead of using a raw commit file

AshleyDumaine (approved):
LGTM

PR #358, opened 2024-06-10
[improvement]: move vpc interface to eth0

rahulait:
This PR moves vpc interface from eth1 to eth0. This simplifies networking for clusters provisioned within VPC.

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests


templates/flavors/kubeadm/default/kustomization.yaml
```diff
@@ -4,7 +4,6 @@ resources:
   - ../../../infra
   - kubeadmConfigTemplate.yaml
   - kubeadmControlPlane.yaml
-  - allow-etcd-policy.yaml
```
eljohnson92, line 7: I think we have some docs noting that this is being set, could you also update to reflect that this is no longer added through a CCNP in our docs?

AshleyDumaine:
LGTM aside from doc change needed on docs/src/topics/vpc.md 👍

templates/flavors/kubeadm/default/kustomization.yaml
rahulait: Updated it. We do require this for kubeadm vpcless flavors as etcd there still uses public ip. Updated docs to reflect the same.

templates/flavors/kubeadm/vpcless/kustomization.yaml
```diff
@@ -63,14 +64,3 @@ patches:
     patch: |-
       - op: remove
         path: /spec/kubeadmConfigSpec/initConfiguration/skipPhases
-  - target:
```
AshleyDumaine, line 66: do we also want to remove this from the rke2 and k3s flavors for consistency?

rahulait: updated.

AshleyDumaine (approved):
LGTM

PR #365, opened 2024-06-12
[feat] add distributed tracing using OTEL

shanduur:
This PR introduces distributed tracing to the CAPL manager using OpenTelemetry (OTEL) Autoexport. It enhances observability by providing detailed insights into the internal workings of the CAPL manager, which helps in performance monitoring, debugging, and integration with observability backends like Zipkin and Jaeger. Additionally, a new chapter on tracing has been added to the documentation to guide users on configuring and using this feature.

Please review the new tracing chapter in the documentation for accuracy and completeness. Ensure that the instructions for installing the OpenTelemetry Collector and setting up environment variables are clear and easy to follow.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


Makefile
```diff
@@ -306,6 +307,7 @@ HUSKY          ?= $(LOCALBIN)/husky
 NILAWAY        ?= $(LOCALBIN)/nilaway
 GOVULNC        ?= $(LOCALBIN)/govulncheck
 MOCKGEN        ?= $(LOCALBIN)/mockgen
+GOWRAP         ?= $(LOCALBIN)/gowrap
```
eljohnson92, line 310: since this isn't supported by devbox can we use `CACHE_BIN` here instead of `LOCALBIN`?

cmd/main.go
```diff
@@ -49,6 +57,13 @@ var (
 	setupLog = ctrl.Log.WithName("setup")
 )
 
+const (
+	controllerName = "cluster-api-privider-linode.linode.com"
```
eljohnson92, line 61:
```suggestion
	controllerName = "cluster-api-provider-linode.linode.com"
```

hack/templates/opentelemetry.go.gotpl
```diff
@@ -0,0 +1,52 @@
...
+        attribute.String("message", err.Error()),
+      )
+    }
+    {{end}}
+    _span.End()
+  }()
+  {{$method.Pass (printf "_d.%s." $.Interface.Name) }}
+}
+  {{end}}
+{{end}}
```
eljohnson92, line 52: nit: missing newline here

Dockerfile
```diff
@@ -18,6 +18,7 @@ COPY api/ api/
 COPY clients/ clients/
 COPY controller/ controller/
 COPY cloud/ cloud/
+COPY observability/ observability/
```
eljohnson92, line 21: looks like the e2e tests are failing because this dir is also missing from the Tiltfile Docker build step, could you add this there as well?

shanduur: Those are still now failing... Do you have any idea what might be wrong here?

eljohnson92: I think the issue is you don't have the `LINODE_TOKEN` secret in your fork? could you try adding that in?

PR #383, opened 2024-06-28
[feat] Add ipv6 support for dns loadbalancing

amold1:
Adds ipv6 support for dns loadbalancing by taking the SLAAC IPv6 of each CP machine and adds it to the DNS record

### DIFF
--- a/cloud/services/domains.go
+++ b/cloud/services/domains.go
@@ -16,6 +16,8 @@ import (
 	rutil "github.com/linode/cluster-api-provider-linode/util/reconciler"
 )
 
+var ipTypeToRecordTypeMapper = map[string]linodego.DomainRecordType{"IPv4": "A", "IPv6": "AAAA"}
+
 // AddIPToDNS creates the A and TXT record for the machine
 func AddIPToDNS(ctx context.Context, mscope *scope.MachineScope) error {
 	dnsTTLSec := rutil.DefaultDNSTTLSec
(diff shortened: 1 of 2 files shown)


cloud/services/domains.go
```diff
@@ -97,37 +103,38 @@ func DeleteIPFromDNS(ctx context.Context, mscope *scope.MachineScope) error {
...
-		err := errors.New("no public IP address")
-		return "", err
+
+	if len(addresses.IPv4.Public) == 0 || addresses.IPv6 == nil {
+		err := errors.New("no public address")
+		return nil, err
 	}
 
-	return addresses.IPv4.Public[0].Address, nil
+	return map[string]string{"IPv4": addresses.IPv4.Public[0].Address, "IPv6": addresses.IPv6.SLAAC.Address}, nil
```
AshleyDumaine, line 125: We may want to check `addresses.IPV6.SLAAC` isn't nil

amold1: Added

cloud/services/domains.go
```diff
@@ -97,37 +103,40 @@ func DeleteIPFromDNS(ctx context.Context, mscope *scope.MachineScope) error {
 	return nil
 }
 
-// GetMachinePublicIP gets the machines public IP
-func GetMachinePublicIP(ctx context.Context, mscope *scope.MachineScope) (string, error) {
+// GetMachinePublicIPs gets the machines public IP
+func GetMachinePublicIPs(ctx context.Context, mscope *scope.MachineScope) (map[string]string, error) {
```
AshleyDumaine, line 107: Since we're only calling this from this file, this should probably be fine to unexport?

cloud/services/domains.go
```diff
@@ -256,7 +265,7 @@ func IsDomainRecordOwner(ctx context.Context, mscope *scope.MachineScope, hostna
 	return true, nil
 }
 
-func CreateMD5HashOfString(stringToConvert string) string {
+func CreateSHA256HashOfString(stringToConvert string) string {
```
AshleyDumaine, line 268: same here

AshleyDumaine (approved):

PR #386, opened 2024-07-01
[fix] make apiserver port configurable for DNS-based LB

AshleyDumaine:
**What this PR does / why we need it**: Allows for the API server port to be configurable in the case that DNS load-balancing is used instead of NodeBalancers

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


rahulait:
I think this will need a bigger change. We also have CCNPs which have 6443 hardcoded in them. Simlarly, cilium config for k3s-default also have 6443 in them.

AshleyDumaine:
> I think this will need a bigger change. We also have CCNPs which have 6443 hardcoded in them. Simlarly, cilium config for k3s-default also have 6443 in them.

I added a new env var for the apiserver port that defaults to 6443 if unset, it seems like before we were expecting users to manually change the `linodecluster.spec.network.apiserverLoadBalancerPort` field instead of templating it for them

cloud/services/loadbalancers.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	defaultApiserverLBPort    = 6443
+	DefaultApiserverLBPort    = 6443
 	defaultKonnectivityLBPort = 8132
```
srust, line 20: worth it just making both of these public and configurable while we're here?

srust (approved):
This is great, thank you

cloud/services/loadbalancers.go
AshleyDumaine: Sounds good to me

templates/infra/linodeCluster.yaml
```diff
@@ -11,3 +11,5 @@ spec:
     apiVersion: infrastructure.cluster.x-k8s.io/v1alpha1
     kind: LinodeVPC
     name: ${VPC_NAME:=${CLUSTER_NAME}}
+  network:
+    apiserverLoadBalancerPort: ${APISERVER_PORT:=6443}
```
rahulait, line 15: This is fine, not sure if we should set this by default in linodeCluster.yaml

rahulait (dismissed):
LGTM

amold1 (approved):

templates/infra/linodeCluster.yaml
AshleyDumaine: Yeah I think I'm going to remove it given the port doesn't seem to be (easily?) configurable for rke2 and k3s won't work without a hacky workaround to account for kustomize stripping the quotes from `KThreesServerConfig.spec.kthreesConfigSpec.serverConfig.httpsListenPort` when it shouldn't

amold1 (approved):

PR #390, opened 2024-07-03
[fix] allow linodemachine controller to recreate linode on delete

amold1:
Currently if a linode from a CAPL cluster is deleted, say, from the CM, the linodemachine controller is not able to recreate it. This also leaves DNS entries behind if the loadbalancer is set to DNS.
This fix allows for the new linode to come up and join the cluster and also deletes the dns entries for the deleted machine and adds the ones for the new linode.

**How do I test**:
Create a 3 CP node cluster with DNS and then note the node ips and dns entries(On CM or using nslookup). Check using kubeconfig that the nodes are part of the cluster.
Now delete a node and verify that the node comes back up with the same name but different IP. Make sure the old DNS entry is gone and replaced with a new one(On CM or using nslookup). Check cluster again

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests


controller/linodemachine_controller.go
```diff
@@ -429,6 +425,25 @@ func (r *LinodeMachineReconciler) addMachineToLB(
 	return nil
 }
 
+func (r *LinodeMachineReconciler) removeMachineFromLB(
+	ctx context.Context,
+	logger logr.Logger,
+	machineScope *scope.MachineScope,
+) error {
+	if machineScope.LinodeCluster.Spec.Network.LoadBalancerType != "dns" {
```
AshleyDumaine, line 433: Probably want to check explicitly that this is a NB type before calling DeleteNodeFromNB:
```suggestion
	if machineScope.LinodeCluster.Spec.Network.LoadBalancerType == "NodeBalancer" {
```
then on 438 check that it's "dns"

controller/linodemachine_controller.go
```diff
@@ -675,16 +693,10 @@ func (r *LinodeMachineReconciler) reconcileDelete(
...
-			logger.Error(err, "Failed to remove node from Node Balancer backend")
-			return err
-		}
-	} else {
-		if err := services.DeleteIPFromDNS(ctx, machineScope); err != nil {
-			logger.Error(err, "Failed to remove IP from DNS")
-			return err
-		}
+	if err := r.removeMachineFromLB(ctx, logger, machineScope); err != nil {
+		logger.Error(err, "Failed to remove machine to LB")
```
AshleyDumaine, line 697:
```suggestion
		logger.Error(err, "Failed to remove machine from LB")
```

cloud/services/domains.go
```diff
@@ -85,46 +82,43 @@ func DeleteIPFromDNS(ctx context.Context, mscope *scope.MachineScope) error {
...
-		return nil, errors.New("no public address")
-	}
-
-	if addresses.IPv6.SLAAC == nil {
-		return nil, errors.New("no SLAAC address")
+func getMachinePublicIPs(mscope *scope.MachineScope) ([]map[string]string, error) {
+	if mscope.LinodeMachine.Status.Addresses == nil {
+		return nil, fmt.Errorf("no addresses available on the LinodeMachine resource")
+	}
+	IPsToReturn := make([]map[string]string, 0, 2) //nolint:mnd // Max 2 IPs(1 IPv4 and 1 IPv6)
```
AshleyDumaine, line 108: This is probably worth keeping as a const up top

cloud/services/domains.go
```diff
@@ -85,46 +82,43 @@ func DeleteIPFromDNS(ctx context.Context, mscope *scope.MachineScope) error {
...
+	for _, IPs := range mscope.LinodeMachine.Status.Addresses {
+		publicIPs := make(map[string]string)
+		if IPs.Type == "ExternalIP" {
+			if net.ParseIP(IPs.Address).To4() != nil {
```
nesv, line 112: If you can, please use the `net/netip` package for handling IP addresses. It has a much less-error-prone API.

controller/linodemachine_controller.go
amold1: Done

controller/linodemachine_controller.go
amold1: Done... Also updated another place with the same grammatical error

cloud/services/domains.go
amold1: I saw issues when running tests for data race... multiple goroutines were trying to read and write the variable is it was not declared within the function

cloud/services/domains.go
amold1: Aye! Updated the code to use `net/netip`

cloud/services/domains.go
```diff
@@ -85,46 +82,47 @@ func DeleteIPFromDNS(ctx context.Context, mscope *scope.MachineScope) error {
...
-		return nil, errors.New("no public address")
-	}
-
-	if addresses.IPv6.SLAAC == nil {
-		return nil, errors.New("no SLAAC address")
+func getMachinePublicIPs(mscope *scope.MachineScope) ([]map[string]string, error) {
+	if mscope.LinodeMachine.Status.Addresses == nil {
+		return nil, fmt.Errorf("no addresses available on the LinodeMachine resource")
+	}
+	IPsToReturn := make([]map[string]string, 0, 2) //nolint:mnd // Max 2 IPs(1 IPv4 and 1 IPv6)
```
nesv, line 108: If the maximum number of IPs to return is 2, it would be better to use two variables, rather than a map.

cloud/services/domains.go
```diff
@@ -85,46 +82,47 @@ func DeleteIPFromDNS(ctx context.Context, mscope *scope.MachineScope) error {
...
-		return nil, err
-	}
-
-	if len(addresses.IPv4.Public) == 0 || addresses.IPv6 == nil {
-		return nil, errors.New("no public address")
-	}
-
-	if addresses.IPv6.SLAAC == nil {
-		return nil, errors.New("no SLAAC address")
+func getMachinePublicIPs(mscope *scope.MachineScope) ([]map[string]string, error) {
```
nesv, line 104: Why does the return value need to be a slice of maps?

cloud/services/domains.go
```diff
@@ -200,8 +198,7 @@ func DeleteDomainRecord(ctx context.Context, mscope *scope.MachineScope, hostnam
 
 	// If record is A type, verify ownership
 	if recordType != "TXT" {
-		txtRecordValueString := createSHA256HashOfString(mscope.LinodeMachine.Name)
-		isOwner, ownerErr := IsDomainRecordOwner(ctx, mscope, hostname, "owner:"+txtRecordValueString, domainID)
+		isOwner, ownerErr := IsDomainRecordOwner(ctx, mscope, hostname, target, domainID)
```
nesv, line 201: nit: just call it `err`. If it would shadow an existing `err` variable, then the lexical scope isn't tight enough.

```suggestion
		isOwner, err := IsDomainRecordOwner(ctx, mscope, hostname, target, domainID)
```

controller/linodemachine_controller.go
```diff
@@ -240,13 +240,10 @@ func (r *LinodeMachineReconciler) reconcile(
...
-		if machineScope.LinodeMachine.Spec.InstanceID != nil {
-			logger = logger.WithValues("ID", *machineScope.LinodeMachine.Spec.InstanceID)
-		}
-
 		res, linodeInstance, err = r.reconcileUpdate(ctx, logger, machineScope)
-
-		return
+		if linodeInstance != nil || err == nil {
+			return
+		}
```
nesv, line 246: Could you please add a comment to explain this block? What happens if `err != nil`? Why do we want to stop reconciliation here? Could we simply call `return r.reconcileUpdate(...)` instead?

controller/linodemachine_controller.go
```diff
@@ -429,6 +425,25 @@ func (r *LinodeMachineReconciler) addMachineToLB(
 	return nil
 }
 
+func (r *LinodeMachineReconciler) removeMachineFromLB(
+	ctx context.Context,
+	logger logr.Logger,
```
nesv, line 430: You don't need to pass a logger in as an argument. `github.com/go-logr/logr` has the [NewContext](https://pkg.go.dev/github.com/go-logr/logr#NewContext) and [FromContextOrDiscard](https://pkg.go.dev/github.com/go-logr/logr#FromContextOrDiscard) functions to store and retrieve loggers from a context.Context.

controller/linodemachine_controller.go
```diff
@@ -429,6 +425,25 @@ func (r *LinodeMachineReconciler) addMachineToLB(
...
+	machineScope *scope.MachineScope,
+) error {
+	if machineScope.LinodeCluster.Spec.Network.LoadBalancerType == "NodeBalancer" {
+		if err := services.DeleteNodeFromNB(ctx, logger, machineScope); err != nil {
+			logger.Error(err, "Failed to remove node from Node Balancer backend")
+			return err
+		}
+	} else if machineScope.LinodeCluster.Spec.Network.LoadBalancerType == "dns" {
```
nesv, line 438: `"NodeBalancer"` and `"dns"` should be constants.

controller/linodemachine_controller.go
```diff
@@ -622,13 +637,16 @@ func (r *LinodeMachineReconciler) reconcileUpdate(
...
 			machineScope.LinodeMachine.Status.InstanceState = nil
+			machineScope.LinodeMachine.Status.Conditions = nil
 
 			conditions.MarkFalse(machineScope.LinodeMachine, clusterv1.ReadyCondition, "missing", clusterv1.ConditionSeverityWarning, "instance not found")
 		}
-
+		if err = r.removeMachineFromLB(ctx, logger, machineScope); err != nil {
+			logger.Error(err, "Failed to remove machine from LB")
+			return res, nil, err
+		}
```
nesv, line 647: Keep the error scoped to the `if` block. It will make any future refactoring much easier.

```suggestion
		if err := r.removeMachineFromLB(ctx, logger, machineScope); err != nil {
			logger.Error(err, "Failed to remove machine from LB")
			return res, nil, err
		}
```

controller/linodemachine_controller.go
```diff
@@ -675,16 +693,10 @@ func (r *LinodeMachineReconciler) reconcileDelete(
...
-		}
-	} else {
-		if err := services.DeleteIPFromDNS(ctx, machineScope); err != nil {
-			logger.Error(err, "Failed to remove IP from DNS")
-			return err
-		}
+	if err := r.removeMachineFromLB(ctx, logger, machineScope); err != nil {
+		logger.Error(err, "Failed to remove machine from LB")
+
+		return err
```
nesv, line 699: Please only log or return the error. There is seldom reason to do both.

If you choose to return the error here, please wrap it like so:

```go
return fmt.Errorf("remove machine from loadbalancer: %w", err)
```

controller/linodemachine_controller_helpers.go
```diff
@@ -109,20 +109,30 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
 func (r *LinodeMachineReconciler) buildInstanceAddrs(ctx context.Context, machineScope *scope.MachineScope, instanceID int) ([]clusterv1.MachineAddress, error) {
 	addresses, err := machineScope.LinodeClient.GetInstanceIPAddresses(ctx, instanceID)
 	if err != nil {
-		return nil, fmt.Errorf("get instance ips: %w", err)
+		return nil, err
```
nesv, line 112: Why was the error wrapping removed?

controller/linodemachine_controller_helpers.go
```diff
@@ -109,20 +109,30 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
...
 	}
 
 	// get the default instance config
 	configs, err := machineScope.LinodeClient.ListInstanceConfigs(ctx, instanceID, &linodego.ListOptions{})
 	if err != nil || len(configs) == 0 {
-		return nil, fmt.Errorf("list instance configs: %w", err)
+		return nil, err
```
nesv, line 118: Why was the error wrapping removed?

controller/linodemachine_controller_helpers.go
```diff
@@ -109,20 +109,30 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
...
 	}
 
 	ips := []clusterv1.MachineAddress{}
-	// check if a node has public ip and store it
-	if len(addresses.IPv4.Public) != 0 {
-		ips = append(ips, clusterv1.MachineAddress{Address: addresses.IPv4.Public[0].Address, Type: clusterv1.MachineExternalIP})
+	// check if a node has public ipv4 ip and store it
+	if len(addresses.IPv4.Public) == 0 {
+		return nil, fmt.Errorf("No Public IPv4 address set on machine")
```
nesv, line 124: If you do not need a formatted error message, please use `errors.New` instead:

```suggestion
		return nil, errors.New("no public ipv4 addresses set on machine")
```

controller/linodemachine_controller_helpers.go
```diff
@@ -109,20 +109,30 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
...
+	}
+	ips = append(ips, clusterv1.MachineAddress{Address: addresses.IPv4.Public[0].Address, Type: clusterv1.MachineExternalIP})
+
+	// check if a node has public ipv6 ip and store it
+	if addresses.IPv6 == nil {
+		return nil, fmt.Errorf("No Public IPv6 address set on machine")
```
nesv, line 130:
```suggestion
		return nil, errors.New("no public ipv6 address set on machine")
```

controller/linodemachine_controller_helpers.go
nesv: This could probably also be hoisted up into being a sentinel error:

```go

var errNoPublicIPv4Addrs = errors.New("no public ipv4 addresses set")

File: controller/linodemachine_controller_helpers.go [hunk 17]
```diff
@@ -109,20 +109,30 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
...
+	}
+	if addresses.IPv6.SLAAC == nil {
+		return nil, fmt.Errorf("No Public SLAAC IPv6 address set on machine")
```
nesv, line 133:
```suggestion
		return nil, errors.New("no public slaac address set on machine")
```

SLAAC is only for IPv6.

controller/linodemachine_controller_helpers.go
```diff
@@ -109,20 +109,30 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
...
 	}
+	ips = append(ips, clusterv1.MachineAddress{Address: addresses.IPv6.SLAAC.Address, Type: clusterv1.MachineExternalIP})
```
nesv, line 135: nit:

```suggestion
	ips = append(ips, clusterv1.MachineAddress{
		Address: addresses.IPv6.SLAAC.Address,
		Type:    clusterv1.MachineExternalIP,
	})
```

amold1: Done

controller/linodemachine_controller_helpers.go
amold1: Done

controller/linodemachine_controller_helpers.go
amold1: Done

controller/linodemachine_controller_helpers.go
amold1: Added back

controller/linodemachine_controller_helpers.go
amold1: Added back

controller/linodemachine_controller.go
amold1: done

controller/linodemachine_controller.go
amold1: done

controller/linodemachine_controller.go
amold1: done

controller/linodemachine_controller.go
amold1: Added a comment

cloud/services/domains.go
amold1: done

cloud/services/domains.go
amold1: DNS record type needs to be set based on IP type. So we nede to link IPv4/IPv6 to A/AAAA. Maps allows us to do that with the DNS record type.
If we use variables as return values, we need to do a isIPv4 or not check again when we try to determine what record needs to be created
The slice allows us to return something like [{"ipv4": "10.10.10.10"}, {"ipv6": "fd00::"}] and then reference returned_value["ipv4"] to set record type

cloud/services/domains.go
amold1: see [comment](https://github.com/linode/cluster-api-provider-linode/pull/390#discussion_r1666789852)

cloud/services/domains.go
nesv: What I was thinking was something along the lines of this:

```go
func getMachinePublicIPs(mscope *scope.MachineScope) (ip4, ip6 netip.Addr, err error)
```

I don't think you would need to do another `ipIPv4` check again afterwards if you went this route. The individual values are named for the data they would hold, and could be checked with [IsValid](https://pkg.go.dev/net/netip#Addr.IsValid) to see if they are the zero value for a `netip.Addr`.

Looking at the function, it seems like it will only return, at most, one IPv4 address, and one IPv6 address. Having two maps with only one key set in each seems a bit overkill for what could be a single map:

```go
addrs := map[string]string{
    "ipv4": "10.10.10.10",
    "ipv6": "fd00::",
}
```

If these IP addresses are only ever going to be put into DNS records, would it make sense for `getMachinePublicIPs` to return a type (new or existing) that could be directly passed to `CreateUpdateDomainRecord`?

> **NOTE**: I know it's outside of the scope for this pull request, so maybe in another one, we could address that `CreateUpdateDomainRecord` takes 6 arguments (excluding the `context.Context`)! That might be better served by passing in a struct.

cloud/services/domains.go
nesv: > This is probably worth keeping as a const up top

@AshleyDumaine I'm not sure what you mean by this.


> I saw issues when running tests for data race... multiple goroutines were trying to read and write the variable if it was not declared within the function

@amold1 Yeah, if you want to keep the race detector happy when you have...well...a race condition, access should be guarded by a `sync.Mutex`, or `sync.RWMutex`.

controller/linodemachine_controller_helpers.go
```diff
@@ -50,6 +50,12 @@ import (
 // The decoded user_data must not exceed 16384 bytes per the Linode API
 const maxBootstrapDataBytes = 16384
 
+var (
+	errNoPublicIPv4Addrs      = errors.New("no public ipv4 addresses set")
+	errNoPublicIPv6Addrs      = errors.New("No public IPv6 address set")
+	errNoPublicIPv6SLAACAddrs = errors.New("No public SLAAC address set")
```
nesv, line 56: Don't capitalize error strings. There is a linter that should be checking this.

```suggestion
	errNoPublicIPv6Addrs      = errors.New("no public IPv6 address set")
	errNoPublicIPv6SLAACAddrs = errors.New("no public SLAAC address set")
```

nesv (commented):
Thank you for making those changes, @amold1! This is looking much better. :smile:

controller/linodemachine_controller_helpers.go
amold1: done

cloud/services/domains.go
amold1: I've simplified the code using a struct now. Still using a slice, but a slice of the struct, instead of a map

cloud/services/domains.go
```diff
@@ -41,90 +45,53 @@ func AddIPToDNS(ctx context.Context, mscope *scope.MachineScope) error {
...
+		return nil, fmt.Errorf("no addresses available on the LinodeMachine resource")
 	}
 	domainHostname := mscope.LinodeCluster.ObjectMeta.Name + "-" + mscope.LinodeCluster.Spec.Network.DNSUniqueIdentifier
 
-	// Delete A record
-	for ipType, publicIP := range publicIPs {
-		if err := DeleteDomainRecord(ctx, mscope, domainHostname, publicIP, dnsTTLSec, domainID, ipTypeToRecordTypeMapper[ipType]); err != nil {
-			return err
+	for _, IPs := range mscope.LinodeMachine.Status.Addresses {
+		recordType := "A"
```
AshleyDumaine, line 79: nit: we should probably use https://pkg.go.dev/github.com/linode/linodego#DomainRecordType, e.g.
```suggestion
		recordType := linodego.RecordTypeA
```

cloud/services/domains.go
```diff
@@ -41,90 +45,53 @@ func AddIPToDNS(ctx context.Context, mscope *scope.MachineScope) error {
...
+		if IPs.Type != "ExternalIP" {
```
AshleyDumaine, line 80:
```suggestion
		if IPs.Type != v1beta1.MachineExternalIP {
```

controller/linodemachine_controller.go
AshleyDumaine: While we're on it, it might be worth changing this to be "DNS"

nesv (approved):

cloud/services/domains.go
```diff
@@ -199,11 +167,10 @@ func DeleteDomainRecord(ctx context.Context, mscope *scope.MachineScope, hostnam
 	}
 
 	// If record is A type, verify ownership
-	if recordType != "TXT" {
-		txtRecordValueString := createSHA256HashOfString(mscope.LinodeMachine.Name)
-		isOwner, ownerErr := IsDomainRecordOwner(ctx, mscope, hostname, "owner:"+txtRecordValueString, domainID)
-		if ownerErr != nil {
-			return ownerErr
+	if dnsEntry.DNSRecordType != "TXT" {
```
AshleyDumaine, line 170:
```suggestion
	if dnsEntry.DNSRecordType != linodego.RecordTypeTXT {
```
Although I wonder if in line with the comment we should be checking only A records

cloud/services/domains.go
```diff
@@ -159,31 +127,31 @@ func CreateUpdateDomainRecord(ctx context.Context, mscope *scope.MachineScope, h
...
+		if err := CreateDomainRecord(ctx, mscope, domainID, dnsEntry); err != nil {
 			return err
 		}
 		return nil
 	}
 
 	// If record exists, update it
-	if len(domainRecords) != 0 && recordType != "TXT" {
-		isOwner, err := IsDomainRecordOwner(ctx, mscope, hostname, target, domainID)
+	if len(domainRecords) != 0 && dnsEntry.DNSRecordType != "TXT" {
```
AshleyDumaine, line 137:
```suggestion
	if len(domainRecords) != 0 && dnsEntry.DNSRecordType != inodego.RecordTypeTXT {
```

cloud/services/domains.go
```diff
@@ -41,90 +46,53 @@ func AddIPToDNS(ctx context.Context, mscope *scope.MachineScope) error {
...
+		addr, err := netip.ParseAddr(IPs.Address)
+		if err != nil {
+			return nil, fmt.Errorf("not a valid IP %w", err)
+		}
+		if !addr.Is4() {
+			recordType = linodego.RecordTypeAAAA
+		}
+		d.options = append(d.options, DNSOptions{domainHostname, IPs.Address, recordType, dnsTTLSec})
 	}
+	d.options = append(d.options, DNSOptions{domainHostname, mscope.LinodeMachine.Name, linodego.DomainRecordType("TXT"), dnsTTLSec})
```
AshleyDumaine, line 93:
```suggestion
	d.options = append(d.options, DNSOptions{domainHostname, mscope.LinodeMachine.Name, inodego.RecordTypeTXT, dnsTTLSec})
```

amold1: done

cloud/services/domains.go
amold1: done

cloud/services/domains.go
amold1: done and updated the comment

controller/linodemachine_controller.go
amold1: I have an upcoming PR to add webhooks for the DNS optional fields. I will make the uppercase change there if that's ok

AshleyDumaine (approved):
Changes work for me, I see a linode get recreated if I delete it on cloud manager

PR #395, opened 2024-07-05
[fix] add support for filling in controlPlaneEndpoint if only network info is specified

AshleyDumaine:
**What this PR does / why we need it**: Currently if a `LinodeCluster` resource is created and the nodebalancer ID/nodebalancer Config ID is specified but not the control-plane endpoint CAPL cannot handle this case, e.g.:
```
spec:
  controlPlaneEndpoint:
    host: ""
    port: 0
  network:
    nodeBalancerConfigID: 12345
    nodeBalancerID: 67890
  region: us-ord 
```
This instead gets the NB and NB config ID if specified and uses those to fill out the controlPlaneEndpoint.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/controller/linodecluster_controller.go
+++ b/controller/linodecluster_controller.go
@@ -142,10 +142,10 @@ func (r *LinodeClusterReconciler) reconcile(
 	}
 
 	// Create
-	if (clusterScope.LinodeCluster.Spec.ControlPlaneEndpoint.Host == "") || (clusterScope.LinodeCluster.Spec.Network.LoadBalancerType == "dns") {
+	if clusterScope.LinodeCluster.Spec.ControlPlaneEndpoint.Host == "" {
 		if err := r.reconcileCreate(ctx, logger, clusterScope); err != nil {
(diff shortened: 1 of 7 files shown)


controller/linodecluster_controller.go
```diff
@@ -142,10 +142,10 @@ func (r *LinodeClusterReconciler) reconcile(
 	}
 
 	// Create
-	if (clusterScope.LinodeCluster.Spec.ControlPlaneEndpoint.Host == "") || (clusterScope.LinodeCluster.Spec.Network.LoadBalancerType == "dns") {
+	if clusterScope.LinodeCluster.Spec.ControlPlaneEndpoint.Host == "" {
```
rahulait, line 145: I wonder why we needed this if loop before for dns which did reconcileCreate and why its not needed now.

eljohnson92: This is new functionality(we couldn't recover from the situation this is fixing before) so I would believe it was needed but we just hadn't tried this scenario yet

eljohnson92 (approved):

PR #404, opened 2024-07-09
fix: handle 429 and 002 errors during linodemachine creation

rahulait:
If too many requests are made to linodeAPI in a short span of time, it returns 429 error. This PR handles 429 errors and makes sure we retry creating machine after some time. Similarly, if there is client timeout or unexpected EOF errors, we get 002 as returncode. This PR handles those errors as well.
Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/controller/linodemachine_controller.go
+++ b/controller/linodemachine_controller.go
@@ -55,6 +55,7 @@ import (
 
 const (
 	linodeBusyCode        = 400
+	linodeTooManyRequests = 429
 	defaultDiskFilesystem = string(linodego.FilesystemExt4)
 
 	// conditions for preflight instance creation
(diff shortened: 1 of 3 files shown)


controller/linodemachine_controller.go
```diff
@@ -308,6 +310,10 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 
 		linodeInstance, err = machineScope.LinodeClient.CreateInstance(ctx, *createOpts)
 		if err != nil {
+			if linodego.ErrHasStatus(err, linodeTooManyRequests) || linodego.ErrHasStatus(err, linodeEOFOrTimeout) {
+				logger.Error(err, "Failed to create Linode instance due to API error, requeing")
+				return ctrl.Result{RequeueAfter: reconciler.DefaultMachineControllerWaitForRunningDelay}, nil
```
rahulait, line 315: Question: should I use a different delay here? Something like `DefaultMachineControllerRetryDelay`?

controller/linodemachine_controller.go
```diff
@@ -308,6 +310,10 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 
 		linodeInstance, err = machineScope.LinodeClient.CreateInstance(ctx, *createOpts)
 		if err != nil {
+			if linodego.ErrHasStatus(err, linodeTooManyRequests) || linodego.ErrHasStatus(err, linodeEOFOrTimeout) {
```
rahulait, line 313: Also check if `linodego.ErrorFromError` works here or not instead of using `linodeEOFOrTimeout`

controller/linodemachine_controller.go
AshleyDumaine: Yeah, my vote would be for `DefaultMachineControllerRetryDelay` since this is different from waiting for the linode to be running. For something like a 429 this makes me think we might want an exponential backoff but just a separate variable is fine for now.

rahulait: updated.

AshleyDumaine (approved):

eljohnson92 (approved):

PR #414, opened 2024-07-15
[feat] linode placement groups - boilerplate + controllers + e2e tests

tchinmai7:
Adds controllers and boilerplate for linode placement groups - introduces the types, web-hooks, controllers and e2e tests.

Fixes #

This doesn't have any integration with `LinodeMachine` or `LinodeCluster` yet - will put up a follow up PR for implementing that. For now, this can create and manage a linode placement group.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests


AshleyDumaine:
@tchinmai7 This generally looks fine though I'm curious how this new CRD was generated. Did you use kubebuilder? I don't see the PROJECT file updated to account for the new CRD.

tchinmai7:
> @tchinmai7 This generally looks fine though I'm curious how this new CRD was generated. Did you use kubebuilder? I don't see the PROJECT file updated to account for the new CRD.

I had manually added the types and ran generate - I'll fix the project file and other stuff to be more `kubebuilder` like

api/v1alpha1/linodeplacementgroup_types.go
```diff
@@ -0,0 +1,156 @@
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// +kubebuilder:default=true
+	// +optional
+	IsStrict bool `json:"isStrict"`
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// +kubebuilder:default="anti_affinity:local"
+	// +kubebuilder:validation:Enum="anti_affinity:local"
+	// +optional
+	AffinityType string `json:"affinity_type"`
```
AshleyDumaine, line 46: This should probably be camel case like the other fields

AshleyDumaine: Also, we should probably do `+kubebuilder:validation:Enum="antiAffinity"`, dropping the ":local"

controller/linodeplacementgroup_controller.go
```diff
@@ -0,0 +1,342 @@
...
+
+	// Add the finalizer if not already there
+	err = pgScope.AddFinalizer(ctx)
+	if err != nil {
+		logger.Error(err, "Failed to add finalizer")
+
+		return
+	}
+
+	// Update
```
AshleyDumaine, line 153: Since all the relevant fields are immutable I think we can skip handling update for now?

api/v1alpha1/linodeplacementgroup_types.go
tchinmai7: The API expects it to be set to `anti_affinity:local` - I could switch up the enum - but that would mean more changes when making the translation from spec->createOptions

controller/linodeplacementgroup_controller.go
tchinmai7: Yeah - good point - will remove it

api/v1alpha1/linodeplacementgroup_types.go
eljohnson92: want to make sure we separate the discussion on the enum vs field name. I agree we should camel case the json type here

tchinmai7: done - switched the json type

AshleyDumaine (dismissed):
Works for me, I was able to create and delete a LPG successfully with the new CRD after updating my mgmt cluster.

eljohnson92:
since we are in the middle of migrating from v1alpha1 to v1alpha2, could we potentially just add this directly to v1alpha2?

tchinmai7:
> since we are in the middle of migrating from v1alpha1 to v1alpha2, could we potentially just add this directly to v1alpha2?

sure - I can do that

PROJECT
```diff
@@ -83,4 +83,16 @@ resources:
...
     conversion: true
     webhookVersion: v1
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: cluster.x-k8s.io
+  group: infrastructure
+  kind: LinodePlacementGroup
+  path: github.com/linode/cluster-api-provider-linode/api/v1alpha1
```
eljohnson92, line 93: after the v1alpha2 move we want to update these as well

controller/linodeplacementgroup_controller.go
```diff
@@ -0,0 +1,321 @@
...
+				logger.Info("re-queuing Placement Group deletion")
+
+				return ctrl.Result{RequeueAfter: reconciler.DefaultPGControllerReconcilerDelay}, nil
+			}
+
+			return ctrl.Result{}, err
+		}
+
+		if pg != nil {
+			if len(pg.Members) > 0 {
```
eljohnson92, line 230: could we potentially move this into it's own function to remove the `nestif` exception?

eljohnson92 (approved):

PR #418, opened 2024-07-19
[feat] Add PlacementGroupRef to linodeMachine

tchinmai7:
1. Adds a placementGroupRef to the linodeMachine controller
2. bump linodego to 1.38
3. fix placementgroup api breaking change

Fixes #

1. To test this, add the following to the generated manifest ( the `placementGroupRef` and `LinodePlacementGroup`)
```
---
kind: LinodeMachineTemplate
metadata:
  name: schinmai-test-control-plane
  namespace: default
spec:
  template:
    spec:
      authorizedKeys: null
      image: linode/ubuntu22.04
      interfaces:
      - purpose: public
      placementGroupRef:
        kind: LinodePlacementGroup
        name: schinmai-test
      region: us-ord
      type: g6-standard-2
---
kind: LinodePlacementGroup
metadata:
  name: schinmai-test
  namespace: default
spec:
  region: us-ord
---
```

**TODOs**:

- [ ] squashed commits
- [x] includes documentation
- [x] adds unit tests


controller/linodemachine_controller_helpers.go
```diff
@@ -109,6 +109,17 @@ func (r *LinodeMachineReconciler) newCreateConfig(ctx context.Context, machineSc
 		createConfig.Interfaces = slices.Insert(createConfig.Interfaces, 0, *iface)
 	}
 
+	if machineScope.LinodeMachine.Spec.PlacementGroupRef != nil {
```
AshleyDumaine, line 112: I think we may want to have the PlacementGroupRef be on the LinodeCluster instead like with VPCRef.

tchinmai7: so we only want the PG on the control-plane nodes, easier to do it on the machine isn't it?

AshleyDumaine: It's not about it being easier, it's that it will be a common reference between multiple machines in the cluster like the VPCRef is

eljohnson92: with the current limitations you cannot put an entire cluster in a placementGroup unless it is very small, it feels more suited to individual machines or groups of machines such as in a machineDeployment

AshleyDumaine: We only use MachineDeployment for the workers, looks like we'll want to keep it on LinodeMachine so we can use it on LinodeMachineTemplate for the control-plane

eljohnson92: yeah, and we also can't update the spec for MachineDeployment

AshleyDumaine (dismissed):

eljohnson92 (approved):

PR #419, opened 2024-07-21
[feat] Add support for using Akamai EDGE DNS as DNS Provider for API Server Loadbalancing

amold1:
Add support for using Akamai EDGE DNS as DNS Provider for API Server Loadbalancing

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests


cloud/scope/common.go
```diff
@@ -23,6 +27,7 @@ import (
 const (
 	// defaultClientTimeout is the default timeout for a client Linode API call
 	defaultClientTimeout = time.Second * 10
+	maxBody              = 131072
```
AshleyDumaine, line 30: should have a comment here like above for what this does

amold1: Done

cloud/services/domains.go
```diff
@@ -62,6 +73,82 @@ func EnsureDNSEntries(ctx context.Context, mscope *scope.MachineScope, operation
...
+		return err
+	}
+	return nil
+}
+
+func updateAkamaiDNSEntry(ctx context.Context, mscope *scope.MachineScope, recordBody *dns.RecordBody) error {
+	if err := mscope.AkamaiDomainsClient.UpdateRecord(ctx, recordBody, mscope.LinodeCluster.Spec.Network.DNSRootDomain); err != nil {
+		return err
+	}
+	return nil
```
AshleyDumaine, line 142: Shouldn't need a conditional here
```suggestion
	return mscope.AkamaiDomainsClient.UpdateRecord(ctx, recordBody, mscope.LinodeCluster.Spec.Network.DNSRootDomain)
```

cloud/services/domains.go
```diff
@@ -62,6 +73,82 @@ func EnsureDNSEntries(ctx context.Context, mscope *scope.MachineScope, operation
...
+}
+
+func deleteAkamaiDNSEntry(ctx context.Context, mscope *scope.MachineScope, recordBody *dns.RecordBody) error {
+	if err := mscope.AkamaiDomainsClient.DeleteRecord(ctx, recordBody, mscope.LinodeCluster.Spec.Network.DNSRootDomain); err != nil {
+		return err
+	}
+	return nil
```
AshleyDumaine, line 149: same here
```suggestion
	return mscope.AkamaiDomainsClient.DeleteRecord(ctx, recordBody, mscope.LinodeCluster.Spec.Network.DNSRootDomain)
```

cloud/services/domains.go
```diff
@@ -62,6 +73,82 @@ func EnsureDNSEntries(ctx context.Context, mscope *scope.MachineScope, operation
...
+	recordBody := &dns.RecordBody{
+		Name:       fqdn,
+		RecordType: string(dnsOptions.DNSRecordType),
+		TTL:        dnsOptions.DNSTTLSec,
+		Target:     []string{dnsOptions.Target},
+	}
+	if err := mscope.AkamaiDomainsClient.CreateRecord(ctx, recordBody, mscope.LinodeCluster.Spec.Network.DNSRootDomain); err != nil {
+		return err
+	}
+	return nil
```
AshleyDumaine, line 135:
```suggestion
	return mscope.AkamaiDomainsClient.CreateRecord(
		ctx,
		&dns.RecordBody{
			Name:       fqdn,
			RecordType: string(dnsOptions.DNSRecordType),
			TTL:        dnsOptions.DNSTTLSec,
			Target:     []string{dnsOptions.Target},
		}, mscope.LinodeCluster.Spec.Network.DNSRootDomain)
```

amold1: done

cloud/services/domains.go
amold1: done

cloud/services/domains.go
amold1: done

cloud/services/domains.go
```diff
@@ -62,6 +73,74 @@ func EnsureDNSEntries(ctx context.Context, mscope *scope.MachineScope, operation
 	return nil
 }
 
+// EnsureAkamaiDNSEntries ensures the domainrecord on Akamai EDGE DNS is created, updated, or deleted based on operation passed
+func EnsureAkamaiDNSEntries(ctx context.Context, mscope *scope.MachineScope, operation string, dnsEntries []DNSOptions) error {
+	// SetUp Client for Akamai EDGE DNS
+
+	fqdn := mscope.LinodeCluster.ObjectMeta.Name + "-" + mscope.LinodeCluster.Spec.Network.DNSUniqueIdentifier + "." + mscope.LinodeCluster.Spec.Network.DNSRootDomain
+
```
AshleyDumaine, line 81: I think it might be worth doing something like
`akaDNSClient := mscope.AkamaiDomainsClient`
and then just calling the Record functions on that without the extra functions below.

config/default/akamai-edgerc-secret.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: akamai-edgerc-secret
+stringData:
+  AKAMAI_HOST: ${AKAMAI_HOST}
```
eljohnson92, line 7: I think all of these have to have defaults so it is not a required field with clusterctl initialization

cloud/services/domains.go
```diff
@@ -62,6 +73,74 @@ func EnsureDNSEntries(ctx context.Context, mscope *scope.MachineScope, operation
...
+			recordBody.Target = append(recordBody.Target, dnsEntry.Target)
+			if err := updateAkamaiDNSEntry(ctx, mscope, recordBody); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func removeElement(stringList []string, elemToRemove string) []string {
```
AshleyDumaine, line 115: I think we might be able to instead use [slices.Delete](https://pkg.go.dev/golang.org/x/exp/slices#Delete) like how we already use [slices.Insert in the linodemachine_controller_helpers](https://github.com/linode/cluster-api-provider-linode/blob/main/controller/linodemachine_controller_helpers.go#L109)

api/v1alpha2/linodecluster_types.go
```diff
@@ -111,7 +111,7 @@ type NetworkSpec struct {
 	// Ignored if the LoadBalancerType is set to anything other than dns
 	// If not set, defaults linode dns
 	// +optional
-	DNSProvider int `json:"dnsProvider,omitempty"`
+	DNSProvider string `json:"dnsProvider,omitempty"`
```
tchinmai7, line 114: do want this to be an enum? - and does switching types need a CRD migration

cloud/services/domains.go
```diff
@@ -62,6 +74,74 @@ func EnsureDNSEntries(ctx context.Context, mscope *scope.MachineScope, operation
...
 
+// EnsureAkamaiDNSEntries ensures the domainrecord on Akamai EDGE DNS is created, updated, or deleted based on operation passed
+func EnsureAkamaiDNSEntries(ctx context.Context, mscope *scope.MachineScope, operation string, dnsEntries []DNSOptions) error {
+	// SetUp Client for Akamai EDGE DNS
+
+	fqdn := mscope.LinodeCluster.ObjectMeta.Name + "-" + mscope.LinodeCluster.Spec.Network.DNSUniqueIdentifier + "." + mscope.LinodeCluster.Spec.Network.DNSRootDomain
+
+	for _, dnsEntry := range dnsEntries {
+		recordBody, err := mscope.AkamaiDomainsClient.GetRecord(ctx, mscope.LinodeCluster.Spec.Network.DNSRootDomain, fqdn, string(dnsEntry.DNSRecordType))
+		if err != nil {
```
AshleyDumaine, line 85: do we know this will only error if the record doesn't exist?

cloud/services/domains.go
```diff
@@ -62,6 +74,74 @@ func EnsureDNSEntries(ctx context.Context, mscope *scope.MachineScope, operation
...
+			if operation == "create" {
+				if err := createAkamaiDNSEntry(ctx, mscope, fqdn, dnsEntry); err != nil {
+					return err
+				}
+			}
+			continue
+		}
+		if operation == "delete" {
+			switch {
+			case len(recordBody.Target) > 1:
```
AshleyDumaine, line 95: is this switch to avoid the nestif on the linter?

amold1: unfortunately, yes

cloud/services/domains.go
amold1: I've added an extra check now - This is what the API sends when the record is not found
```
2024/07/22 16:30:27 Title: Not Found; Type: https://problems.luna.akamaiapis.net/authoritative-dns/notFound; Detail: Unable to find record set with name nonexistent-domain.akafn.com and type A
```

api/v1alpha2/linodecluster_types.go
amold1: unsure about the need for migration but I've added enum kubebuilder check to restrict this to linode/akamai

config/default/akamai-edgerc-secret.yaml
amold1: I've added a ```""``` default now

cloud/services/domains.go
amold1: Done

cloud/services/domains.go
AshleyDumaine: ah gotcha. If there's an error other than Not Found, should this bail out with the error so we can retry later (e.g. if the API is down?)

amold1: Yeah you are right. We now bail out on any error other than "Not Found"

eljohnson92 (approved):

PR #422, opened 2024-07-23
[fix] account for retryable errors

AshleyDumaine:
**What this PR does / why we need it**: This introduces a new utility function that reconcilers can use to determine if they should requeue reconciliation to try again later

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/controller/linodemachine_controller.go
+++ b/controller/linodemachine_controller.go
@@ -300,6 +300,9 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 		createOpts, err := r.newCreateConfig(ctx, machineScope, tags, logger)
 		if err != nil {
 			logger.Error(err, "Failed to create Linode machine InstanceCreateOptions")
+			if util.IsTransientError(err) {
+				return ctrl.Result{RequeueAfter: reconciler.DefaultMachineControllerRetryDelay}, nil
+			}
 
 			return ctrl.Result{}, err
 		}
--- a/util/helpers.go
+++ b/util/helpers.go
@@ -2,6 +2,9 @@ package util
 
 import (
 	"errors"
+	"io"
+	"net/http"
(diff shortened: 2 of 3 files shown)


amold1 (commented):
LGTM... Is this the only place where we need this kind of check?

AshleyDumaine:
> LGTM... Is this the only place where we need this kind of check?

Based on the scale testing we ran it is. We can add this in other places if needed but the main issues we saw were on the machine controller specifically.

amold1:
> > LGTM... Is this the only place where we need this kind of check?
> 
> Based on the scale testing we ran it is. We can add this in other places if needed but the main issues we saw were on the machine controller specifically.

Should we use the new function in places like [these](https://github.com/linode/cluster-api-provider-linode/blob/reconcile-machine-fix/controller/linodemachine_controller.go#L497) in the `linodemachine_controller.go`?

AshleyDumaine:
> > > LGTM... Is this the only place where we need this kind of check?
> > 
> > 
> > Based on the scale testing we ran it is. We can add this in other places if needed but the main issues we saw were on the machine controller specifically.
> 
> Should we use the new function in places like [these](https://github.com/linode/cluster-api-provider-linode/blob/reconcile-machine-fix/controller/linodemachine_controller.go#L497) in the `linodemachine_controller.go`?

That already gets requeued if there's an error on https://github.com/linode/cluster-api-provider-linode/blob/reconcile-machine-fix/controller/linodemachine_controller.go#L352. Any errors encountered would be from the Linode API whereas in this change the error could be either from the Linode API or a k8s error

amold1 (approved):

eljohnson92 (approved):

PR #423, opened 2024-07-24
[feat] add support for disk encryption if the capability is present

AshleyDumaine:
**What this PR does / why we need it**: LDE was released as of the [latest linodego release](https://github.com/linode/linodego/releases/tag/v1.38.0). This allows for that feature to be used (if available via a webhook check) for LinodeMachines. This is an optional field (thus we may want to remove "disabled" from the enum? I included it for consistency with https://github.com/linode/linodego/blob/v1.38.0/instances.go#L156)

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation <-- still need this
- [x] adds unit tests


api/v1alpha2/linodemachine_types.go
```diff
@@ -74,6 +74,10 @@ type LinodeMachineSpec struct {
 	// DataDisks is a map of any additional disks to add to an instance,
 	// The sum of these disks + the OSDisk must not be more than allowed on a linodes plan
 	DataDisks map[string]*InstanceDisk `json:"dataDisks,omitempty"`
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// +kubebuilder:validation:Enum=enabled;disabled
+	// DiskEncryption determines if the disks of the instance should be encrypted.
+	DiskEncryption string `json:"diskEncryption,omitempty"`
```
tchinmai7, line 80: should we set a default?

eljohnson92: The platform default is to have LDE enabled so I think leaving this empty should result in LDE=true as the platform default

api/v1alpha2/linodemachine_webhook.go
```diff
@@ -113,8 +113,14 @@ func (r *LinodeMachine) validateLinodeMachineSpec(ctx context.Context, client Li
 	// TODO: instrument with tracing, might need refactor to preserve readibility
 	var errs field.ErrorList
 
-	if err := validateRegion(ctx, client, r.Spec.Region, field.NewPath("spec").Child("region")); err != nil {
-		errs = append(errs, err)
+	if r.Spec.DiskEncryption == "enabled" {
```
eljohnson92, line 116: given LDE will be enabled by default platform wide I think we can remove this check now?

AshleyDumaine: Is LDE available in all DCs yet? If not we should leave this in

eljohnson92: since LDE is enabled by default, the use case for this flag is not actually to turn LDE on, but rather to turn it off. I think that would make this check a noop in almost all situations.

eljohnson92 (approved):

PR #427, opened 2024-07-24
[feat] add a v1alpha2 version of LinodeObjectStorageBucket

unnatiagg:
This PR adds LinodeObjectStorageBucket  to the v1alpha2 version
Fixes #
This is the Step 2: for the two fold update mentioned in [PR](https://github.com/linode/cluster-api-provider-linode/pull/417)

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


api/v1alpha1/linodeobjectstoragebucket_conversion.go
```diff
@@ -0,0 +1,86 @@
...
+	src, ok := srcRaw.(*infrastructurev1alpha2.LinodeObjectStorageBucket)
+	if !ok {
+		return errors.New("failed to convert LinodeObjectStorageBucket version from v1alpha2 to v1alpha1")
+	}
+
+	// ObjectMeta
+	dst.ObjectMeta = src.ObjectMeta
+
+	// Spec
+	dst.Spec.Cluster = src.Spec.Region + "-1"
```
bcm820, line 70: Is it guaranteed that cluster will be `{region}-1`?

Of course, I can't imagine a world where anyone would be trying to convert to `v1alpha1`. Is the reverse conversion from `v1alpha2` to `v1alpha1` actually necessary?

api/v1alpha2/linodeobjectstoragebucket_types.go
```diff
@@ -0,0 +1,139 @@
...
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Region is the ID of the Object Storage region for the bucket.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	Region string `json:"region"`
+
+	// Cluster is the ID of the Object Storage cluster for the bucket.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// +optional
+	Cluster string `json:"cluster,omitempty"`
```
bcm820, line 43: Shouldn't we remove `cluster`?

api/v1alpha1/linodeobjectstoragebucket_conversion.go
eljohnson92: I don't think it is necessarily true that it will always be `{region}-1` sounds like this might make downgrading pretty challenging here though

config/crd/bases/infrastructure.cluster.x-k8s.io_linodeobjectstoragebuckets.yaml
```diff
@@ -179,6 +179,175 @@ spec:
             type: object
         type: object
     served: true
+    storage: false
```
tchinmai7, line 182: you'd need to update the reference in `config/crd/patches/capicontract_in_linodeobjectstoragebuckets.yaml` to `v1alpha2` as well

unnatiagg: Hey @tchinmai7 , I have actually made that change in my local version later. Will be pushing it out in a bit.

api/v1alpha1/linodeobjectstoragebucket_conversion.go
tchinmai7: ^ while it has been true so far, the api now responds with a `hostname` for a bucket - maybe we can infer the clusterID from it - not always guaranteed to work, but better than `-1`

bcm820: Agreed. For successfully provisioned v1alpha1 buckets, the hostname is stored as `status.hostname`. If no `status.hostname` has been set (i.e. `status.ready` is false), then it's probably safe to default to `{region}-1` since it'd be a no-op.

api/v1alpha2/linodeobjectstoragebucket_types.go
unnatiagg: I have removed this field from v1alpha2.

api/v1alpha1/linodeobjectstoragebucket_conversion.go
unnatiagg: going forward with {region}-1, while converting from v1alpha2 to v1alpha1.

eljohnson92: will v1alpha1 work without that conversion?

api/v1alpha1/conversion.go
```diff
@@ -43,3 +45,30 @@ func Convert_v1alpha2_LinodeMachineSpec_To_v1alpha1_LinodeMachineSpec(in *infras
...
+	// WARNING: in.Region requires manual conversion: does not exist in peer-type
+	out.Cluster = in.Region
+	out.CredentialsRef = in.CredentialsRef
+	out.KeyGeneration = in.KeyGeneration
+	out.SecretType = in.SecretType
+	return nil
+}
+
+func Convert_v1alpha2_LinodeObjectStorageBucket_To_v1alpha1_LinodeObjectStorageBucket(in *infrastructurev1alpha2.LinodeObjectStorageBucket, out *LinodeObjectStorageBucket, scope conversion.Scope) error {
+	if in.Status.Ready {
```
eljohnson92, line 68: can we just check if `in.Status.Hostname` is not `""` instead of checking Ready?

PR #432, opened 2024-07-29
[fix] add more checks for retrying on transient error during machine creation, bump default retry for machine controller to 30 seconds

AshleyDumaine:
**What this PR does / why we need it**: During scale testing, we ran into 429 errors like below:
```
2024-06-28T14:49:10.187107427Z stderr F 2024-06-28T14:49:10Z    DEBUG   events  [429] Too many requests {"type": "Warning", "object": {"kind":"LinodeMachine","namespace":"default","name":"kubeadm4-md-0-xhh88-4j5tq","uid":"ebef4ebf-2659-4024-9524-186e8f5f7024","apiVersion":"infrastructure.cluster.x-k8s.io/v1alpha1","resourceVersion":"1138536"}, "reason": "CreateError"}
2024-06-28T14:49:10.19961153Z stderr F 2024-06-28T14:49:10Z     ERROR   Reconciler error        {"controller": "linodemachine", "controllerGroup": "infrastructure.cluster.x-k8s.io", "controllerKind": "LinodeMachine", "LinodeMachine": {"name":"kubeadm4-md-0-xhh88-4j5tq","namespace":"default"}, "namespace": "default", "name": "kubeadm4-md-0-xhh88-4j5tq", "reconcileID": "d7a20060-1f4b-4360-b852-14fdd1675bba", "error": "[429] Too many requests"}
```
This change adds extra checks for if reconciliation of a machine can be retried due to too many requests and extends the retry delay to 30 seconds from the current 10 seconds which may be too aggressive for when 429s are encountered.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


util/reconciler/defaults.go
```diff
@@ -37,7 +37,7 @@ const (
 	// DefaultMachineControllerWaitForRunningTimeout is the default timeout if instance is not running.
 	DefaultMachineControllerWaitForRunningTimeout = 20 * time.Minute
 	// DefaultMachineControllerRetryDelay is the default requeue delay if there is an error.
-	DefaultMachineControllerRetryDelay = 10 * time.Second
+	DefaultMachineControllerRetryDelay = 30 * time.Second
```
eljohnson92, line 40: if this bump is specifically for 429s could we maybe create a retry case just for too many requests so other transient errors can be retried more quickly?

AshleyDumaine: The others are 500 errors, I'm not sure we want to retry every 10 seconds in those cases either but can.

eljohnson92: if the errors are actually transient a quick retry should generally be appropriate, if you think we need to reevaluate retries for that status then I think we can do that as well

eljohnson92 (approved):

PR #433, opened 2024-07-30
use linodemachine status fields to delete dns records

amold1:
**Summary:**
If the entire NS, where a CAPL cluster is deployed, is deleted, DNS entries don't get cleaned up. This PR fixes that issue.

**Details:**
When a CAPL cluster is deployed in a namespace and the entire namespace is deleted, DNS entries are not cleaned up. This happens because when the NS is deleted, all resources are marked for deletion and since LinodeCluster is not directly linked to LinodeMachines, it gets deleted. This causes the linodemachine controller to fail as it relies on the LinodeCluster network spec for DNS clean ups.

**Solution:**
Store the LinodeCluster network spec to the LinodeMachine resource so that the controller can function independently

**Sample LinodeCluster Spec and LinodeMachine Status:**
```
$ k -n test-ns get linodemachine test-cluster-control-plane-g2mwb -o yaml | yq '.status.clusterNetworkSpec'
apiserverLoadBalancerPort: 6443
dnsProvider: linode
dnsRootDomain: adeodhar.net
dnsUniqueIdentifier: a1b2c3
loadBalancerType: dns
$
```
```
$ k -n test-ns get linodecluster test-cluster -o yaml | yq '.spec.network'
apiserverLoadBalancerPort: 6443
dnsProvider: linode
dnsRootDomain: adeodhar.net
dnsUniqueIdentifier: a1b2c3
loadBalancerType: dns
$
```


cloud/services/domains.go
```diff
@@ -37,14 +40,19 @@ func EnsureDNSEntries(ctx context.Context, mscope *scope.MachineScope, operation
 		return nil
 	}
 
+	// Verify that the ClusterNetworkSpec is set
+	if cmp.Equal(mscope.LinodeMachine.Status.ClusterNetworkSpec, v1alpha2.NetworkSpec{}) {
```
AshleyDumaine, line 44: [cmp should only be used for tests since it can panic if it can't compare values](https://pkg.go.dev/github.com/google/go-cmp/cmp)

amold1: fixed with [this commit](https://github.com/linode/cluster-api-provider-linode/pull/433/commits/c44ec02ba3d909b107d85cb13bacd871cdb11277)

AshleyDumaine: awesome, thanks! 🙏

AshleyDumaine:
While this change does look good, I'm wondering if instead of this approach we should instead modify how we're removing the finalizer on the `LinodeCluster` so it doesn't prematurely clean up in the case of dns-based LB instead of NB-based. 

We currently check that the `LinodeCluster.Spec.Network.NodeBalancerID` is nil to determine if the `LinodeCluster` can have its finalizer removed, but maybe we can have some field on the `LinodeCluster` (maybe in the `Status`) that must be set to nil if using dns-based LB before the finalizer can be removed?

I'm not sure we want to add this status field to the LinodeMachines if it's only the control plane ones that need it to remove themselves from DNS.

amold1:
> While this change does look good, I'm wondering if instead of this approach we should instead modify how we're removing the finalizer on the `LinodeCluster` so it doesn't prematurely clean up in the case of dns-based LB instead of NB-based.
> 
> We currently check that the `LinodeCluster.Spec.Network.NodeBalancerID` is nil to determine if the `LinodeCluster` can have its finalizer removed, but maybe we can have some field on the `LinodeCluster` (maybe in the `Status`) that must be set to nil if using dns-based LB before the finalizer can be removed?
> 
> I'm not sure we want to add this status field to the `LinodeMachines` if it's only the control plane ones that need it to remove themselves from DNS.

The field on the LinodeCluster Status to be set/unset to indicate that dns entries have been cleaned up  will need to be done by the LinodeCluster itself.
The machine controller will not know if all machines have been cleaned up. And the LinodeCluster doesnt have a "map" of the LinodeMachines associated to it to know that information either

PR #442, opened 2024-08-06
[improvement] add override so users can set full dns name for controlplane endpoint

amold1:
LKE would like an enhancement that would allow users to override CAPLs domain name format in lieu of their own full domain.

Example: Currently we have these settings-
```
kind: LinodeCluster
metadata:
    name: testname
spec:
    network:
        dnsRootDomain: example.com
        dnsUniqueIdentifier: abc123
```
and CAPL creates a domain name with the format
```
spec:
    controlPlaneEndpoint:
        host: testname-abc123.example.com
        port: 6443
```

This PR adds an override feature so users can set the hostname
```
kind: LinodeCluster
metadata:
    name: testname
spec:
    network:
        dnsEndpointOverride: testname-admin.example.com
```
that will create
```
spec:
    controlPlaneEndpoint:
        host: testname-admin.example.com
        port: 6443
```


api/v1alpha2/linodecluster_types.go
```diff
@@ -128,6 +128,9 @@ type NetworkSpec struct {
 	// If not set, defaults to 30
 	// +optional
 	DNSTTLSec int `json:"dnsTTLsec,omitempty"`
+	// DNSEndpointOverride is used to override CAPL's construction of the controlplane endpoint
+	// +optional
+	DNSEndpointOverride string `json:"dnsEndpointOverride,omitempty"`
```
eljohnson92, line 133: Could we maybe rename this to just `DNSEndpoint` with a description of it overriding the other settings? also might be worth in this message saying specifically it ignores roodDomain+unique identifier

cloud/services/domains.go
```diff
@@ -150,7 +153,18 @@ func (d *DNSEntries) getDNSEntriesToEnsure(mscope *scope.MachineScope) ([]DNSOpt
 	if mscope.LinodeMachine.Status.Addresses == nil {
 		return nil, fmt.Errorf("no addresses available on the LinodeMachine resource")
 	}
-	domainHostname := mscope.LinodeCluster.ObjectMeta.Name + "-" + mscope.LinodeCluster.Spec.Network.DNSUniqueIdentifier
+	clusterMetadata := mscope.LinodeCluster.ObjectMeta
+	clusterSpec := mscope.LinodeCluster.Spec
+	var domainHostname string
+	if clusterSpec.Network.DNSEndpointOverride != "" {
+		domainHostname = strings.ReplaceAll(clusterSpec.Network.DNSEndpointOverride, clusterSpec.Network.DNSRootDomain, "")
```
scmeggitt, line 160: is the expectation that the DNSEndpointOverride would contain the DNSRootDomain? It is fine if that is true, just want to make sure I understand what domainHostname is being created.

cloud/services/domains.go
```diff
@@ -79,8 +79,11 @@ func EnsureAkamaiDNSEntries(ctx context.Context, mscope *scope.MachineScope, ope
 	linodeCluster := mscope.LinodeCluster
 	linodeClusterNetworkSpec := linodeCluster.Spec.Network
 	rootDomain := linodeClusterNetworkSpec.DNSRootDomain
-	fqdn := linodeCluster.Name + "-" + linodeClusterNetworkSpec.DNSUniqueIdentifier + "." + rootDomain
 	akaDNSClient := mscope.AkamaiDomainsClient
+	fqdn := linodeCluster.Name + "-" + linodeClusterNetworkSpec.DNSUniqueIdentifier + "." + rootDomain
+	if linodeClusterNetworkSpec.DNSEndpointOverride != "" {
+		fqdn = linodeClusterNetworkSpec.DNSEndpointOverride
```
scmeggitt, line 85: Should this contain the rootDomain?

amold1: post our discussion, i have updated the override field to only override the subdomain

cloud/services/domains.go
amold1: post our discussion, i have updated the override field to only override the subdomain

api/v1alpha2/linodecluster_types.go
amold1: renamed and updated the comment

controller/linodecluster_controller.go
```diff
@@ -232,13 +232,25 @@ func (r *LinodeClusterReconciler) handleNBCreate(ctx context.Context, logger log
 }
 
 func (r *LinodeClusterReconciler) handleDNS(clusterScope *scope.ClusterScope) {
-	domainName := clusterScope.LinodeCluster.ObjectMeta.Name + "-" + clusterScope.LinodeCluster.Spec.Network.DNSUniqueIdentifier + "." + clusterScope.LinodeCluster.Spec.Network.DNSRootDomain
+	subDomain := ""
+	clusterSpec := clusterScope.LinodeCluster.Spec
+	if clusterScope.LinodeCluster.Spec.Network.DNSSubDomainOverride != "" {
+		subDomain = clusterScope.LinodeCluster.Spec.Network.DNSSubDomainOverride
```
eljohnson92, line 238: can we just initialize the subdomain to what is in the `else` part of this and override it with the override if it exists?

eljohnson92 (approved):

scmeggitt (approved):

PR #443, opened 2024-08-07
[feat] Adding new fields in LinodeObjectStorageBucket and decoupling ObjKey Functionality

unnatiagg:
To keep CAPL API in line with Linodego APIs, Adding two fields to the LinodeObjectStorageBucket

1. acl: string; 
2. corsEnabled: bool (defaults to true, per API docs)

Fixes #
With the introduction of LinodeObjectStorageKey in CAPL, this PR decouples all the access key related code from the 
LinodeObjectStorageBucket controllers, services, and testcases.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


unnatiagg:
Lint gives the following suggestion:
controller/linodeobjectstoragebucket_controller.go:166:63: `(*LinodeObjectStorageBucketReconciler).reconcileDelete` - `ctx` is unused (unparam)
`func (r *LinodeObjectStorageBucketReconciler) reconcileDelete(ctx context.Context, bScope *scope.ObjectStorageBucketScope) error {}`

Should we remove the context param for now, with an additional comment on top of the function?

unnatiagg:
The minimal-linodeObjectStorageBucket test lifecycle before tested the following:
1. Create a Bucket, Secret is created with it by default. Check the creation of the bucket and the secret.
2. Patch the Bucket object for key rotation. Assert the key generation
3. Delete the secret. Check for the restored secret.
4. Delete the Bucket. Check if the bucket and secret are deleted together.

Given the decoupling, the minimal-linodeObjectStorageBucket test lifecycle now:
1. Create a Bucket. Check the creation of the bucket.
2. Delete the Bucket. Check if the bucket and secret are deleted together.

In a new PR after this is merged, I would like to create a e2e test, where a bucket and (read-write,read-only) access keys are created separately.
1. Checking their resource association with each other.
2. Restore logic is tested
3. Patch logic is tested

cloud/services/object_storage_buckets.go
```diff
@@ -31,7 +29,7 @@ func EnsureObjectStorageBucket(ctx context.Context, bScope *scope.ObjectStorageB
 	opts := linodego.ObjectStorageBucketCreateOptions{
 		Region: bScope.Bucket.Spec.Region,
 		Label:  bScope.Bucket.Name,
-		ACL:    linodego.ACLPrivate,
+		ACL:    linodego.ObjectStorageACL(bScope.Bucket.Spec.ACL),
```
komer3, line 32: Question: Where is the cors var being set? I'm assuming we have to set before calling Linode API for obj creation?

komer3:
> Lint gives the following suggestion: controller/linodeobjectstoragebucket_controller.go:166:63: `(*LinodeObjectStorageBucketReconciler).reconcileDelete` - `ctx` is unused (unparam) `func (r *LinodeObjectStorageBucketReconciler) reconcileDelete(ctx context.Context, bScope *scope.ObjectStorageBucketScope) error {}`
> 
> Should we remove the context param for now, with an additional comment on top of the function?

During the standup today, I believe we decided to remove unused context and just add it back later if needed

e2e/linodeobjectstoragebucket-controller/minimal-linodeobjectstoragebucket/chainsaw-test.yaml
komer3, line 1: Could we set the acl in the LinodeObj resource to something other than default one to check if the bucket that is created has the right acl (check using linode api - unsure if the api response has this info)

cloud/services/object_storage_buckets.go
unnatiagg: Hey, I had a doubt about his, these changes have not been pushed yet.
Linodego API defines CorsEnabled *bool , so for CAPL in LinodeObjectStorageBucketSpec should the field be defined as 
`CorsEnabled bool` and coverted as following `opts := linodego.ObjectStorageBucketCreateOptions{
		Region:      bScope.Bucket.Spec.Region,
		Label:       bScope.Bucket.Name,
		ACL:         linodego.ObjectStorageACL(bScope.Bucket.Spec.ACL),
		CorsEnabled: util.Pointer(bScope.Bucket.Spec.CorsEnabled),
	}`

OR

directly defined as `CorsEnabled *bool`?

komer3: I would make `CorsEnabled` a bool pointer personally. Like you have in your second option!

I'm not sure if there is a preferred way in the team. Might be a good question to ask and get opinions.

eljohnson92: I agree with @komer3 your second solution looks like the correct one

api/v1alpha1/conversion.go
```diff
@@ -50,16 +51,26 @@ func Convert_v1alpha1_LinodeObjectStorageBucketSpec_To_v1alpha2_LinodeObjectStor
...
+	// WARNING: in.AccessKeyRefs requires manual conversion: does not exist in peer-type
+	return nil
+}
 
 func Convert_v1alpha2_LinodeObjectStorageBucketSpec_To_v1alpha1_LinodeObjectStorageBucketSpec(in *infrastructurev1alpha2.LinodeObjectStorageBucketSpec, out *LinodeObjectStorageBucketSpec, s conversion.Scope) error {
 	// WARNING: in.Region requires manual conversion: does not exist in peer-type
 	out.Cluster = in.Region
 	out.CredentialsRef = in.CredentialsRef
-	out.KeyGeneration = in.KeyGeneration
+	out.KeyGeneration = ptr.To(1)
```
bcm820, line 73: Would `0` make more sense here? Or rather, if the Object controller is no longer generating keys, is it safe to leave this as `nil`?

api/v1alpha2/linodeobjectstoragebucket_types.go
```diff
@@ -37,16 +44,22 @@ type LinodeObjectStorageBucketSpec struct {
...
+	// ACL sets The Access Control Level of the bucket using a canned ACL string
+	// +optional
+	// +kubebuilder:default=private
+	// +kubebuilder:validation:Enum=private;public-read;authenticated-read;public-read-write
+	ACL ObjectStorageACL `json:"acl,omitempty"`
+
+	// corsEnabled enables for all origins in the bucket .If set to false, CORS is disabled for all origins in the bucket
+	// +optional
+	// +kubebuilder:default=true
+	CorsEnabled *bool `json:"corsEnabled,omitempty"`
```
bcm820, line 56: Does it need to be a pointer?

controller/linodeobjectstoragebucket_controller.go
```diff
@@ -233,16 +162,10 @@ func (r *LinodeObjectStorageBucketReconciler) reconcileApply(ctx context.Context
 	return nil
 }
 
-func (r *LinodeObjectStorageBucketReconciler) reconcileDelete(ctx context.Context, bScope *scope.ObjectStorageBucketScope) error {
+// ctx has been removed from the param list
```
bcm820, line 165: nit: you can remove this comment

controller/linodeobjectstoragebucket_controller.go
```diff
@@ -233,16 +162,10 @@ func (r *LinodeObjectStorageBucketReconciler) reconcileApply(ctx context.Context
...
+func (r *LinodeObjectStorageBucketReconciler) reconcileDelete(bScope *scope.ObjectStorageBucketScope) error {
```
bcm820, line 166: We should be able to get rid of this function entirely now since we only added the finalizer to make sure we handled key revocation. This also means that we can get rid of the finalizer for OBJ buckets.

api/v1alpha2/linodeobjectstoragebucket_types.go
bcm820: Ah, I see. I think we don't need to make ours a pointer. We can just pass the pointer to the real value when doing create/update opts.

eljohnson92 (approved):

komer3 (approved):

bcm820 (approved):

PR #451, opened 2024-08-09
[feat] Add LinodeFirewall controller

AshleyDumaine:
**What this PR does / why we need it**: It should be possible to create a Cloud Firewall rather than just relying on cilium hostNetwork policies for a cluster. This adds the ability to create Cloud Firewalls but does not yet tie them to Machines or Clusters

Fixes #

Example Firewall:
```
kind: LinodeFirewall
metadata:
  name: test-fw
spec:
  enabled: true
  inboundPolicy: DROP
  inboundRules:
    - action: ACCEPT
      label: ssh
      ports: "22"
      protocol: "TCP"
      addresses:
        ipv4:
          - "192.168.1.1/32"
```
The inboundRules and outboundRules follow https://pkg.go.dev/github.com/linode/linodego#FirewallRule 

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests


.github/workflows/e2e-test.yaml
```diff
@@ -29,6 +29,7 @@ on:
           - linodeobj
           - linodevpc
           - linodeplacementgroup
+          - linodefirewall
```
komer3, line 32: Not sure if you added this test to `.github/filters.yml`. We need to add it there to trigger this e2e test!

.golangci.yml
```diff
@@ -131,7 +131,7 @@ linters:
     - decorder
     # - depguard
     - dogsled
-    - dupl
+    # - dupl
```
komer3, line 134: question: any reason you commented this out?

api/v1alpha2/linodefirewall_types.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// +optional
+	// +kubebuilder:default=false
+	Enabled bool `json:"enabled,omitempty"`
+
+	// +kubebuilder:validation:MinLength=3
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// +optional
+	Label string `json:"label,omitempty"`
```
komer3, line 45: question: what do we use this label for? Custom naming for the firewall?

cloud/scope/firewall.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Override the controller credentials with ones from the Firewall's Secret reference (if supplied).
+	if params.LinodeFirewall.Spec.CredentialsRef != nil {
+		// TODO: This key is hard-coded (for now) to match the externally-managed `manager-credentials` Secret.
+		apiToken, err := getCredentialDataFromRef(ctx, params.Client, *params.LinodeFirewall.Spec.CredentialsRef, params.LinodeFirewall.GetNamespace(), "apiToken")
+		if err != nil {
+			return nil, fmt.Errorf("credentials from secret ref: %w", err)
+		}
+		linodeClientConfig.Token = string(apiToken)
+	}
```
komer3, line 69: Should we have an else statement here incase the credential ref is not set for whatever reason?

.golangci.yml
AshleyDumaine: Yeah, we would have needed a lot of nodupl like we did for the PG controller. We're duplicating a lot of code between PG, FW, and VPC controllers. A separate PR would be needed for using generics

cloud/scope/firewall.go
komer3, line 69: Should we have an else statement here incase the credential ref is not set for whatever reason?

.github/workflows/e2e-test.yaml
AshleyDumaine: I only see flavor tests there, none of the flavors include a firewall yet

api/v1alpha2/linodefirewall_types.go
AshleyDumaine: I think that's leftover from March and can probably be removed. IIRC we make the Linode label based on the k8s resource name

.github/workflows/e2e-test.yaml
komer3: oh yeah we don't have firewall flavor yet. I though you added it. Sorry about that. 

e2e test you added runs under quick test so all good!

config/crd/kustomization.yaml
```diff
@@ -36,6 +37,7 @@ patches:
 - path: patches/webhook_in_linodeobjectstoragebuckets.yaml
 - path: patches/webhook_in_linodeclustertemplates.yaml
 - path: patches/webhook_in_linodemachinetemplates.yaml
+#- path: patches/webhook_in_linodefirewalls.yaml
```
komer3, line 40: I'm guessing webhooks are not part of this PR implemention?

controller/linodefirewall_controller.go
```diff
@@ -0,0 +1,273 @@
...
+			string(failureReason),
+			err.Error(),
+			reconciler.DefaultTimeout(r.ReconcileTimeout, reconciler.DefaultFWControllerReconcileTimeout),
+		)
+		r.Recorder.Event(fwScope.LinodeFirewall, corev1.EventTypeWarning, string(failureReason), err.Error())
+
+		switch {
+		case errors.Is(err, errTooManyIPs):
+			// Cannot reconcile firewall with too many ips, wait for an update to the spec
+			return ctrl.Result{}, nil
```
komer3, line 180: question: I'm curious, how many ips is too many?

cloud/scope/firewall.go
AshleyDumaine: the credential ref is optional, we shouldn't need to do that. This is what the VPC and PG scopes are also doing.

komer3 (commented):
Not sure why the e2e test for firewall is failing. But otherwise looks good!

komer3 (dismissed):

controller/linodefirewall_controller.go
```diff
@@ -0,0 +1,273 @@
...
+		return fmt.Errorf("failed to create mapper for LinodeFirewalls: %w", err)
+	}
+	err = ctrl.NewControllerManagedBy(mgr).
+		For(&infrav1alpha2.LinodeFirewall{}).
+		WithOptions(options).
+		WithEventFilter(
+			predicate.And(
+				predicates.ResourceNotPausedAndHasFilterLabel(mgr.GetLogger(), r.WatchFilterValue),
+				predicate.GenerationChangedPredicate{},
+				predicate.Funcs{DeleteFunc: func(e event.DeleteEvent) bool { return false }},
```
eljohnson92, line 249: I think we've landed on this predicate no longer being necessary?

AshleyDumaine: yep

eljohnson92 (approved):

PR #456, opened 2024-08-15
[feat] LOBJ controller updated for mutable fields LinodeObjectStorage Finalizer removed

unnatiagg:
This PR enables the LOBJ controller to update mutable fields like ACL, CorsEnabled on the actual cloud Manager. 
The LinodeObjectStorage Finalizer has been removed with the entirety of Reconciling delete functionality. 

Test cases have been added for Updating the Fields and Reconciling Delete testcases have been filtered. 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


cloud/services/object_storage_buckets.go
```diff
@@ -30,7 +46,7 @@ func EnsureObjectStorageBucket(ctx context.Context, bScope *scope.ObjectStorageB
 		Region:      bScope.Bucket.Spec.Region,
 		Label:       bScope.Bucket.Name,
 		ACL:         linodego.ObjectStorageACL(bScope.Bucket.Spec.ACL),
-		CorsEnabled: bScope.Bucket.Spec.CorsEnabled,
+		CorsEnabled: util.Pointer(bScope.Bucket.Spec.CorsEnabled),
```
eljohnson92, line 49: we shouldn't need to use this util here
```suggestion
		CorsEnabled: &bScope.Bucket.Spec.CorsEnabled,
```

cloud/services/object_storage_buckets.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/linode/cluster-api-provider-linode/util"
 )
 
-func EnsureObjectStorageBucket(ctx context.Context, bScope *scope.ObjectStorageBucketScope) (*linodego.ObjectStorageBucket, error) {
+func EnsureAndUpdateObjectStorageBucket(ctx context.Context, bScope *scope.ObjectStorageBucketScope) (*linodego.ObjectStorageBucket, error) {
```
eljohnson92, line 14: I think there is an opportunity to simplify this nested if logic here, could we do a small refactor here so it is something like
```
if bucket == nil {
 return create bucket
}
if bucketAccess == bscope.Access { return bucket, nil }
update bucketAccess
return bucket,nil
```

cloud/services/object_storage_buckets.go
```diff
@@ -21,7 +21,23 @@ func EnsureObjectStorageBucket(ctx context.Context, bScope *scope.ObjectStorageB
 		return nil, fmt.Errorf("failed to get bucket from region %s: %w", bScope.Bucket.Spec.Region, err)
 	}
 	if bucket != nil {
-		bScope.Logger.Info("Bucket exists")
+		bucketAccess, err := bScope.LinodeClient.GetObjectStorageBucketAccess(
```
komer3, line 24: Nit: I think adding a log here to say we have an existing bucket and we are updating/updated bucket access would be beneficial

cloud/services/object_storage_buckets.go
```diff
@@ -20,24 +20,45 @@ func EnsureObjectStorageBucket(ctx context.Context, bScope *scope.ObjectStorageB
...
+		bucket.Region,
+		bucket.Label,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get bucket access details for %s: %w", bScope.Bucket.Name, err)
 	}
 
-	if bucket, err = bScope.LinodeClient.CreateObjectStorageBucket(ctx, opts); err != nil {
-		return nil, fmt.Errorf("failed to create bucket: %w", err)
+	if (bucketAccess.ACL == linodego.ObjectStorageACL(bScope.Bucket.Spec.ACL)) && bucketAccess.CorsEnabled == bScope.Bucket.Spec.CorsEnabled {
```
bcm820, line 49: nit: Parens are not needed around the first conditional

controller/linodeobjectstoragebucket_controller.go
```diff
@@ -117,15 +115,6 @@ func (r *LinodeObjectStorageBucketReconciler) reconcile(ctx context.Context, bSc
...
-
-	if !bScope.Bucket.DeletionTimestamp.IsZero() {
-		return res, r.reconcileDelete(bScope)
-	}
-
-	if err := bScope.AddFinalizer(ctx); err != nil {
-		return res, err
-	}
-
 	if err := r.reconcileApply(ctx, bScope); err != nil {
```
bcm820, line 118: We should still check if `!bScope.Bucket.DeletionTimestamp.IsZero()` and return `res, nil` to avoid invoking `r.reconcileApply` when the resource is deleted.

bcm820: Alternatively, we could implement Predicate's Delete function to always return false to ignore all deletions. Down at Line 174-
```go
		WithEventFilter(predicate.And(
			predicate.Funcs{DeleteFunc: func(event.DeleteEvent) bool { return false }},
			predicates.ResourceHasFilterLabel(mgr.GetLogger(), r.WatchFilterValue),
			predicate.GenerationChangedPredicate{},
		)).
```

unnatiagg: @bcm820 I checked a few times, even without your suggested predicate changes, the reconcile loop is not triggered at all on Deletion. I checked a few times and added logger statements while debugging as well. My understanding was, previously it was called due to an update event on the Deletiontimestamp and then reconciled. Is my understanding correct? Without finalizers, isn't the resource sent for deletion immediately?

unnatiagg: I think Brian the Reconcile is triggered but the error gets caught here `if err := r.TracedClient().Get(ctx, req.NamespacedName, objectStorageBucket); err != nil` and will never reach the reconcile apply. Found this as part of an article. **But after an object is deleted and the reconciler is called because of that delete, the call r.Get(ctx, req.NamespacedName, &obj) will fail with a not found err because the object is immediately deleted internally.** Do we still need the predicate to not initiate a reconciliation event during Deletion?

bcm820: Ah right, it doesn't find the object because it gets deleted right away. That's a pretty safe guarantee, and so filtering out deletes might be an over-optimization. I'd say let's not worry about the predicate at this time.

eljohnson92 (approved):

komer3 (approved):

bcm820 (approved):

PR #462, opened 2024-08-16
[improvement] update install script to support newer containerd versions

eljohnson92:
This PR moves the apt install of containerd to a `curl` based install that allows us to control the containerd version better. This is needed for ubuntu 24.04 support which requires a newer containerd version to support new AppArmor defaults
Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/scripts/pre-kubeadminit.sh
+++ b/scripts/pre-kubeadminit.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -euo pipefail
-
+CONTAINERD_VERSION='1.7.20'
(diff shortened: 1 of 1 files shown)


rahulait (approved):
LGTM. Hoping this keeps working for ubuntu 22.04 as well. Current GHA runs might not test this change as we download a specific commit version of this pre script, so we might have to manually test this with older versions. Overall, the changes look good to me.

eljohnson92:
I've confirmed this is working with debian12,ubuntu 24.04 and ubuntu 22.04

PR #465, opened 2024-08-19
[deprecation][feature] deprecate Firewall ID and add FirewallRef

AshleyDumaine:
**What this PR does / why we need it**: With the addition of LinodeFirewalls in #451 we should use FirewallRefs instead similar to PGs and VPCs

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests


api/v1alpha2/linodemachine_types.go
```diff
@@ -66,8 +66,6 @@ type LinodeMachineSpec struct {
 	PrivateIP *bool `json:"privateIP,omitempty"`
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	Tags []string `json:"tags,omitempty"`
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
-	FirewallID int `json:"firewallID,omitempty"`
```
eljohnson92, line 70: do we want to delete this or deprecate it?

AshleyDumaine: I can add this back and FirewallRef's ID can take precedence if it's set

eljohnson92: I would be fine with firewallRef taking precedence if it is set over ID since we want to deprecate ID anyway. I think we also need a deprecation annotation on firewallID like
```
// +kubebuilder:deprecatedversionwarning=
```

eljohnson92 (approved):

PR #470, opened 2024-08-19
[cleanup] move non-reconcile functions into controller helpers

AshleyDumaine:
**What this PR does / why we need it**: General cleanup for controllers controller which should have helper functions in the appropriate helpers file for consistency

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controller/linodecluster_controller_helpers.go
+++ b/controller/linodecluster_controller_helpers.go
@@ -10,9 +10,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/linode/linodego"
 	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/cluster-api/api/v1beta1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
-	cerrs "sigs.k8s.io/cluster-api/errors"
 	kutil "sigs.k8s.io/cluster-api/util"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -25,7 +23,7 @@ import (
 	"github.com/linode/cluster-api-provider-linode/util/reconciler"
 )
 
-func (r *LinodeClusterReconciler) addMachineToLB(ctx context.Context, clusterScope *scope.ClusterScope) error {
+func addMachineToLB(ctx context.Context, clusterScope *scope.ClusterScope) error {
 	logger := logr.FromContextOrDiscard(ctx)
 	if clusterScope.LinodeCluster.Spec.Network.LoadBalancerType == "dns" {
 		if err := services.EnsureDNSEntries(ctx, clusterScope, "create"); err != nil {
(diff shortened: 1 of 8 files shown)


komer3 (dismissed):
LGTM!

controller/linodecluster_controller_helpers.go
```diff
@@ -40,3 +51,110 @@ func (r *LinodeClusterReconciler) removeMachineFromLB(ctx context.Context, logge
 	}
 	return nil
 }
+
```
amold1, line 54: can we make this a "common" function instead on a cluster reconciler method?
The only thing it uses from the reconciler is the TracedClient. If we just pass the client, other reconcilers could use the same function like [this one](https://github.com/linode/cluster-api-provider-linode/blob/main/controller/linodemachine_controller_helpers.go#L191)

controller/linodecluster_controller_helpers.go
```diff
@@ -40,3 +51,110 @@ func (r *LinodeClusterReconciler) removeMachineFromLB(ctx context.Context, logge
...
+				Namespace: linodeCluster.Namespace,
+				Name:      linodeCluster.Name,
+			},
+		})
+
+		return result
+	}
+}
+
+func (r *LinodeClusterReconciler) handleDNS(clusterScope *scope.ClusterScope) {
```
amold1, line 98: looks like all these functions could do without being a reconciler method

controller/linodemachine_controller.go
```diff
@@ -125,25 +118,38 @@ func (r *LinodeMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	cluster, err := kutil.GetClusterFromMetadata(ctx, r.TracedClient(), machine.ObjectMeta)
 	if err != nil || cluster == nil {
 		return ctrl.Result{}, err
 	}
 
+	linodeClusterKey := client.ObjectKey{
+		Namespace: linodeMachine.Namespace,
+		Name:      cluster.Spec.InfrastructureRef.Name,
+	}
+	linodeCluster := &infrav1alpha2.LinodeCluster{}
```
amold1, line 136: I like this approach. But now that we are populating the field here, we should remove the `Get()` calls in the reconcile method

amold1 (commented):
we should probably create a generic/common helpers file and put the non-reconciler specific functions in there

AshleyDumaine:
> we should probably create a generic/common helpers file and put the non-reconciler specific functions in there

since we pass machineScope around so much, it looks like nothing actually needs the reconciler (`r.Get` can be replaced with `machineScope.Client.Get`). I was able to remove it on everything.

controller/linodecluster_controller_helpers.go
AshleyDumaine: updated to pass the traced client

amold1 (dismissed):

amold1:
LGTM

controller/linodecluster_controller_helpers.go
AshleyDumaine: updated

amold1 (dismissed):

eljohnson92 (approved):

eljohnson92 (approved):

PR #471, opened 2024-08-20
[fix] fix etcd-backup-restore based flavors

amold1:
CAPL clusterrs with etcd-backup-restore and related flavors fail to created. This is down to some recent OBJ changes for key names and namespaces within the secrets.

### DIFF
--- a/cloud/scope/object_storage_key.go
+++ b/cloud/scope/object_storage_key.go
@@ -107,6 +107,7 @@ const (
 apiVersion: v1
 metadata:
   name: %s
+  namespace: kube-system
 stringData:
   bucket_name: %s
   bucket_region: %s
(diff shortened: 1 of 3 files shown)


unnatiagg:
LGTM. Just a question @bcm820 for my understanding, the keys and bucket can be in different namespaces? From the PR I see, the secrets of type=addons.cluster.x-k8s.io/resource-set should always be created in namespace=kube-system. Was there any specific reason why it was removed in the first place?

bcm820 (approved):

amold1:
> LGTM. Just a question @bcm820 for my understanding, the keys and bucket can be in different namespaces? From the PR I see, the secrets of type=addons.cluster.x-k8s.io/resource-set should always be created in namespace=kube-system. Was there any specific reason why it was removed in the first place?

That is hard-coded because that secret is only meant for backup-restore and the [backup-restore pod is hardcoded to be in the kube-system namespace](https://github.com/linode/cluster-api-provider-linode/blob/main/templates/addons/etcd-backup-restore/etcd-backup-restore.yaml#L13)

PR #476, opened 2024-08-22
[feat] Support key secret templating

bcm820:
Allows for customizing how the data in the generated secret (from LinodeObjectStorageKey) is formatted.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests


cloud/scope/object_storage_key.go
```diff
@@ -99,27 +99,7 @@ func (s *ObjectStorageKeyScope) AddFinalizer(ctx context.Context) error {
 	return nil
 }
 
-const (
-	accessKeySecretNameTemplate = "%s-obj-key"
-
-	ClusterResourceSetSecretFilename = "etcd-backup.yaml"
-	BucketKeySecret                  = `kind: Secret
```
eljohnson92, line 106: what are your thoughts on keeping this as a default in case type is resource-set so we don't have to add this into the templates?

api/v1alpha2/linodeobjectstoragekey_types.go
```diff
@@ -52,8 +52,17 @@ type LinodeObjectStorageKeySpec struct {
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	// +optional
 	SecretType corev1.SecretType `json:"secretType,omitempty"`
+
+	// SecretDataFormat instructs the controller how to format the data stored in the secret containing access key details.
+	// It supports Go template syntax and interpolating the following values: .AccessKey, .SecretKey.
+	// If no format is supplied then a generic one is used containing the values specified.
+	// When SecretType is set to addons.cluster.x-k8s.io/resource-set, a .BucketEndpoint value is also available pointing to the location of the first bucket specified in BucketAccess.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// +optional
```
AshleyDumaine, line 64: Do we want to remove this optional marker if it looks like we error if this being unset in https://github.com/linode/cluster-api-provider-linode/pull/476/files#diff-cd057278b04faf08bb16208c91ddf55e5777d6c317e376fae1e84e071d9e3199R123?

cloud/scope/object_storage_key.go
bcm820: It makes more sense to me to keep it generic. There could be other times when secrets are generated in workload clusters for different use cases besides the one we define for etcd-backup-restore. Since we own the cluster resource-set defined in the etcd-backup-restore addon, it's easy for us to just define it there.

eljohnson92: very fair, sounds good to me!

api/v1alpha2/linodeobjectstoragekey_types.go
bcm820: We still want this to optional whenever the secret type is opaque. We only require templates whenever we are populating a cluster resource-set secret type.

amold1: should we add a webhook to error out whenever this is required but not set?

cloud/scope/object_storage_key.go
```diff
@@ -128,13 +108,21 @@ func (s *ObjectStorageKeyScope) GenerateKeySecret(ctx context.Context, key *lino
...
+
+	tmplData := map[string]string{
+		"AccessKey": key.AccessKey,
+		"SecretKey": key.SecretKey,
+	}
 
 	// If the desired secret is of ClusterResourceSet type, encapsulate the secret.
 	// Bucket details are retrieved from the first referenced LinodeObjectStorageBucket in the access key.
 	if s.Key.Spec.SecretType == clusteraddonsv1.ClusterResourceSetSecretType {
+		if len(s.Key.Spec.SecretDataFormat) == 0 {
```
amold1, line 122: yeah i think a webhook would be better here

controller/linodeobjectstoragekey_controller.go
```diff
@@ -173,16 +173,26 @@ func (r *LinodeObjectStorageKeyReconciler) reconcileApply(ctx context.Context, k
...
+			} else {
+				keyScope.Logger.Error(err, "Failed check for access key secret")
+				r.setFailure(keyScope, fmt.Errorf("failed check for access key secret: %w", err))
+
+				return err
+			}
 		}
 
-		if ok {
+		if shouldRestoreSecret {
```
amold1, line 195: should we stick line 195-207 on line 186 instead of set a var and using if?

amold1 (approved):

PR #486, opened 2024-08-26
[feat] Generate OBJKey Secret in other namespaces

bcm820:
Allows for customizing where a generated secret is created from a `LinodeObjectStorageKey`. I also updated the spec to have a `generatedSecret` field which contains four sub-fields: `name`, `namespace`, `type` (replacing `secretType`), and `format` (replacing `secretDataFormat`). I have deprecated the previous fields, and have introduced a mutating webhook for supporting the old fields whenever the new fields are not set.

Also adds tests and documentation for `LinodeObjectStorageKey`.

Fixes #

**TODOs**:

- [ ] squashed commits
- [x] includes documentation
- [x] adds unit tests


api/v1alpha2/linodeobjectstoragekey_types.go
```diff
@@ -90,10 +112,6 @@ type LinodeObjectStorageKeyStatus struct {
 	// +optional
 	LastKeyGeneration *int `json:"lastKeyGeneration,omitempty"`
 
-	// SecretName specifies the name of the Secret containing access key data.
-	// +optional
```
unnatiagg, line 94: Do we no longer need to set the GeneratedSecret.Name in the status after generating a secret for the key? Do we think this might be useful later?

controller/linodeobjectstoragekey_controller.go
unnatiagg, line 1: Would case keyScope.Key.Status.AccessKeyRef != nil sufficient enough to conclude that the secret was successfully generated and therefore retrieved?

bcm820: Yes, it should be sufficient, even when secret creation fails after a key is provisioned or rotated. As long as we store a pointer to an OBJ key, we should ensure its secret exists (re-generating it if it has been deleted).

api/v1alpha2/linodeobjectstoragekey_types.go
bcm820: Yes, we only want a single source of truth for referring to the secret. We manage the fields in `spec.generatedSecret.name` and `.namespace` so we don't need to replicate those fields in status.

amold1:
LGTM

unnatiagg:
LGTM Brian. Really appreciate your commit messages. Helps to review stuff.

unnatiagg (approved):

PR #455, opened 2024-08-14
[ci] lock golangci-lint version for now

AshleyDumaine:
**What this PR does / why we need it**: https://github.com/golangci/golangci-lint/releases/tag/v1.60.1 just came out and is causing go-analyze to fail on main. We should lock the version for now until we can address the necessary lint updates

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #489, opened 2024-08-29
[fix] golangci-lint is unlocked to use the latest version again

unnatiagg:
golangci-lint was locked to use v1.59.1 in this PR https://github.com/linode/cluster-api-provider-linode/pull/455.
It is switched back to use the latest verison and all the lint updates are taken care of. 

Fixes #
**The breaking upgrade that golangci-lint latest brought in was mandating passing of format strings for Printf and similar functions.**
For example: 
Printf and similar functions are usually called with a literal format string, followed by zero or more arguments. 
Even though Printf(v) is accepted, it is not the best way. This format string will fail if v string has unexpected percent sign.

 Function calls have been modified to look like: **conditions.MarkFalse(machineScope.LinodeMachine, ConditionPreflightRootDiskResized, string(cerrs.CreateMachineError), clusterv1.ConditionSeverityWarning, "%s", err.Error())**
 
This is better way to handle all types of cases where the intended use was a literal format string. Printf("%s", v). 

Other updates are pretty self explanatory.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


.golangci.yml
```diff
@@ -138,9 +138,8 @@ linters:
     - errname
     - errorlint
     - errcheck
-    - exportloopref
+    - copyloopvar
```
komer3, line 141: Question: is exportloopref no longer supported?

AshleyDumaine (approved):

controller/linodefirewall_controller.go
```diff
@@ -115,7 +115,7 @@ func (r *LinodeFirewallReconciler) reconcile(
 		if err != nil {
 			fwScope.LinodeFirewall.Status.FailureReason = util.Pointer(failureReason)
 			fwScope.LinodeFirewall.Status.FailureMessage = util.Pointer(err.Error())
-			conditions.MarkFalse(fwScope.LinodeFirewall, clusterv1.ReadyCondition, string(failureReason), clusterv1.ConditionSeverityError, err.Error())
+			conditions.MarkFalse(fwScope.LinodeFirewall, clusterv1.ReadyCondition, string(failureReason), clusterv1.ConditionSeverityError, "%s", err.Error())
```
komer3, line 118: Why are we adding `%s`?

.golangci.yml
unnatiagg: I did this change due to the following warning: level=warning msg="The linter 'exportloopref' is deprecated (since v1.60.2) due to: Since Go1.22 (loopvar) this linter is no longer relevant. Replaced by copyloopvar."

controller/linodefirewall_controller.go
unnatiagg: conditions.MarkFalse expects internally calls Message:  fmt.Sprintf(messageFormat, messageArgs...). err.Error() is working as string format instead of an actual string. If it has any unexpected % sign, this will fail. So golangci-lint has got picker with such things. %s works as a format string.

komer3 (approved):

rahulait (approved):

PR #498, opened 2024-09-04
[fix] block credentials deletion until all dependent resources are cleaned up

amold1:
We are currently relying on the credentials secret to fetch token for the linodeclient. The machine and cluster controllers rely on this to be present.
We need to block the deletion of said secret until all resources that rely on this secret are deleted.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controller/linodecluster_controller.go
+++ b/controller/linodecluster_controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"time"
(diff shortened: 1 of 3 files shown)


controller/linodecluster_controller.go
```diff
@@ -219,6 +220,10 @@ func (r *LinodeClusterReconciler) reconcileDelete(ctx context.Context, logger lo
 	if clusterScope.LinodeCluster.Spec.Network.NodeBalancerID == nil && clusterScope.LinodeCluster.Spec.Network.LoadBalancerType != lbTypeDNS {
 		logger.Info("NodeBalancer ID is missing, nothing to do")
 
+		if len(clusterScope.LinodeMachines.Items) > 0 {
+			return errors.New("Waiting for associated LinodeMachine objects to be deleted")
```
eljohnson92, line 224: should this be an error or just a requeue?

amold1: this method only returns errors - I am letting [this](https://github.com/linode/cluster-api-provider-linode/blob/main/controller/linodecluster_controller.go#L144-L150) take care of the re-queue

eljohnson92: ah, I see, it feels a bit weird that we are returning an error just because we aren't ready to delete yet, but that is probably all we can do without changing this to return a requeue request instead

eljohnson92 (approved):

AshleyDumaine (approved):

PR #507, opened 2024-09-09
[fix] create scope with default credentials for all resources

unnatiagg:
For all reconciliation requests, the scope was previously created with the Resource.Spec.credentialsRef details (if present).
The scope creation failed if the secret disappears on deletion due to any race conditions.

With this change, the scope for all resources is set from the default credentials always. And later the token is updated using the SetToken method provided with LinodeGo API.

To further avoid any race conditions on the secret deletion, CredentialRefFinalizers are added onto secret using LinodeCluster, LinodeFirewall, LinodeVPC controllers. 
With all these finalizers set on the secret from various resources, the deletion of the secret does not happen as quickly as it did before.

**If in future we want to explicitly add a finalizer on the secret itself, I believe VPC controller would be best way to handle this.** 

Resources whose scope creation is changed:
LinodeCluster
LinodeMachine
LinodeVPC
LinodeFirewall
LinodePlacementGroup
LinodeOBJKey

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [⟢] adds unit tests


controller/linodecluster_controller.go
```diff
@@ -157,6 +157,11 @@ func (r *LinodeClusterReconciler) reconcile(
 		return res, err
 	}
 
+	if err := clusterScope.SetCredentialRefTokenForLinodeClients(ctx); err != nil {
```
amold1, line 160: I think we should set before we handle the deletes. When object(LinodeCluster in this case) is marked for deletion, it will enter the `if !clusterScope.LinodeCluster.DeletionTimestamp.IsZero() {` block and never reach this

amold1, line 160: I think we should set before we handle the deletes. When object(LinodeCluster in this case) is marked for deletion, it will enter the `if !clusterScope.LinodeCluster.DeletionTimestamp.IsZero() {` block and never reach this

eljohnson92 (approved):

PR #511, opened 2024-09-12
[fix] - handling external network spec

unnatiagg:
As part of https://jira.linode.com/browse/CNS-2959 we want to support external controlplane endpoint and skip creating a new NodeBalancer or DNS entries in this case. 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controller/linodecluster_controller_helpers.go
+++ b/controller/linodecluster_controller_helpers.go
@@ -2,7 +2,6 @@ package controller
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"slices"
 	"strings"
@@ -30,12 +29,6 @@ func addMachineToLB(ctx context.Context, clusterScope *scope.ClusterScope) error
 			return err
 		}
 		return nil
-	}
-	if clusterScope.LinodeCluster.Spec.Network.NodeBalancerID == nil {
-		return errors.New("nil NodeBalancer Config ID")
-	}
-	if clusterScope.LinodeCluster.Spec.Network.ApiserverNodeBalancerConfigID == nil {
-		return errors.New("nil NodeBalancer Config ID")
 	}
 	nodeBalancerNodes, err := clusterScope.LinodeClient.ListNodeBalancerNodes(
 		ctx,
--- a/controller/linodecluster_controller.go
+++ b/controller/linodecluster_controller.go
@@ -178,10 +178,12 @@ func (r *LinodeClusterReconciler) reconcile(
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


controller/linodecluster_controller_helpers.go
```diff
@@ -31,12 +30,6 @@ func addMachineToLB(ctx context.Context, clusterScope *scope.ClusterScope) error
 		}
 		return nil
 	}
-	if clusterScope.LinodeCluster.Spec.Network.NodeBalancerID == nil {
```
unnatiagg, line 34: If feel this check is redundant, because the createReconcile explicitly has it own requeueing and sets at the very last. `clusterScope.LinodeCluster.Spec.ControlPlaneEndpoint = clusterv1.APIEndpoint{}`. Ou execution will only reach inside the loop when both LinodeCluster.Spec.Network.NodeBalancerID and clusterScope.LinodeCluster.Spec.Network.ApiserverNodeBalancerConfigID are correctly set.

api/v1alpha2/linodecluster_types.go
```diff
@@ -104,7 +104,8 @@ func (lm *LinodeCluster) SetConditions(conditions clusterv1.Conditions) {
 // NetworkSpec encapsulates Linode networking resources.
 type NetworkSpec struct {
 	// LoadBalancerType is the type of load balancer to use, defaults to NodeBalancer if not otherwise set
-	// +kubebuilder:validation:Enum=NodeBalancer;dns
+	// +kubebuilder:validation:Enum=NodeBalancer;dns;external
+	// +kubebuilder:default=NodeBalancer
```
unnatiagg, line 108: Added the ++kubebuilder:default=NodeBalancer, which was missing. Not sure if it was done on purpose.

unnatiagg:
Although the deletion logic works out of box, I feel this is an opportunity to refactor it quite a bit.  as it a bit complex. Let me know your thoughts? Also an observation, `
		if len(clusterScope.LinodeMachines.Items) > 0 {}`, clusterScope.LinodeMachines.Items only has controlNodes in the item List, I think we might want to wait for all the machines to be deleted and not just the controlplane nodes.

unnatiagg:
STILL FIXING THE DELETION LOGIC.

amold1 (dismissed):
LGTM

eljohnson92 (approved):

eljohnson92 (approved):

PR #519, opened 2024-09-23
[improvement] : Limit post calls to /v4/linode/instances endpoint

rahulait:
This PR takes into account linode's 5 req/15 secs and 10 req/30 secs rate limits for POST to /v4/linode/instances endpoint.

We maintain count for number of requests made for each token and if we have used up all allowed requests, we'll wait until the limits reset. We rely on the response headers for counting how many more requests can be made and when the rate-limits will reset.
Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests


config/manager/manager.yaml
```diff
@@ -113,10 +113,9 @@ spec:
           # More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
           resources:
             limits:
-              cpu: 500m
-              memory: 128Mi
+              memory: 2Gi
```
amold1, line 116: we have taken of the cpu limit and increased the memory limit and the requests too. can you add a note about why we need this?

controller/linodemachine_controller.go
```diff
@@ -278,14 +278,36 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
 
+// createInstance provisions linode instance after checking if the request will be within the rate-limits
+// Note: it takes a lock before checking for the rate limits and releases it after making request to linode API or when returning from function
+func (r *LinodeMachineReconciler) createInstance(ctx context.Context, logger logr.Logger, machineScope *scope.MachineScope, createOpts *linodego.InstanceCreateOptions) (*linodego.Instance, error, bool) {
+	ctr := util.GetPostReqCounter(machineScope.TokenHash)
+	ctr.Mu.Lock()
+	defer ctr.Mu.Unlock()
+
+	if ctr.IsPOSTLimitReached() {
+		logger.Info(fmt.Sprintf("Cannot make more requests as max requests have been made. Waiting and retrying after %v seconds", reconciler.SecondaryLinodeTooManyPOSTRequestsErrorRetryDelay))
```
amold1, line 289: are the limits an internal thing only? if not, can you add the max request to the log?

controller/linodemachine_controller.go
```diff
@@ -278,14 +278,36 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 	return ctrl.Result{}, nil
 }
 
+// createInstance provisions linode instance after checking if the request will be within the rate-limits
+// Note: it takes a lock before checking for the rate limits and releases it after making request to linode API or when returning from function
+func (r *LinodeMachineReconciler) createInstance(ctx context.Context, logger logr.Logger, machineScope *scope.MachineScope, createOpts *linodego.InstanceCreateOptions) (*linodego.Instance, error, bool) {
```
amold1, line 283: I think this function could be reworked to avoid extra return values. We should define specific type of errors, and perform checks for those specific errors when the function is called.
```
errInstanceCreate
errRateLimit
```
and remove bool

This will also help in testing by creating tests for specific errors

controller/linodemachine_controller.go
```diff
@@ -295,6 +317,11 @@ func (r *LinodeMachineReconciler) reconcilePreflightCreate(ctx context.Context,
 		}
 		return retryIfTransient(err)
 	}
+
+	if linodeInstance == nil {
```
amold1, line 321: Is this check necessary? Based on the above implementation, linodeInstance will be nil only when the err is not nil and we are already performing that check above

controller/linodemachine_controller.go
AshleyDumaine, line 283: nit: we don't need this to be a reconciler function. Might also want to move this into the helpers for the machine controller

rahulait: updated code based on suggestion

controller/linodemachine_controller.go
rahulait: updated with newer implementation

controller/linodemachine_controller.go
rahulait: updated

config/manager/manager.yaml
rahulait: we discussed about the increase over chat, resolving the conversation

controller/linodemachine_controller.go
rahulait: updated

amold1 (approved):
thanks for addressing the comments... lgtm!

amold1 (approved):

PR #522, opened 2024-09-26
[improvement] : add new phase to check if we can use cloudinit or not to reduce API calls

rahulait:
Previously, we were doing GET for image and region during every reconcileCreate(). If there is an error during reconcileCreate(), the request is requeued and we again do same GET calls during next reconcile as well. We found that for a 100 node cluster, we were making 1200+ GET calls for regions and 1200+ GET calls for images. This change will help in reducing that number down to 1 call per machine.

When provisioning 100 node cluster with older changes:
```
cat logs/old/concurrency10-1cluster-100node1.txt | grep "GET  /v4beta/regions" | wc -l
1252
cat logs/old/concurrency10-1cluster-100node1.txt | grep "GET  /v4beta/images" | wc -l
1091
```
When provisioning 100 node cluster with newer changes:
```
cat logs/new/concurrency10-1cluster-100node1.txt | grep "GET  /v4beta/regions" | wc -l
103
cat logs/new/concurrency10-1cluster-100node1.txt | grep "GET  /v4beta/images" | wc -l
103
```
Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests


api/v1alpha2/linodemachine_types.go
```diff
@@ -161,6 +161,11 @@ type LinodeMachineStatus struct {
 	// Addresses contains the Linode instance associated addresses.
 	Addresses []clusterv1.MachineAddress `json:"addresses,omitempty"`
 
+	// CloudinitMetadataSupport determines whether to use cloud-init or not.
+	// +optional
+	// +kubebuilder:default=true
+	CloudinitMetadataSupport bool `json:"cloudinitMetadataSupport,omitempty"`
```
eljohnson92, line 167: This seems to be a chance separate from the rest of the PR, do you think we now need this field as well?

rahulait: This is required as we store the result of getRegion() and getImage() in this variable. We set it in reconcilePreflightMetadataSupportConfigure(). It gets used later where we decide to use cloud-init or stackscripts based on the value of this variable.

rahulait: It will automatically set it. We are doing this in code here: **https://github.com/linode/cluster-api-provider-linode/pull/522/files#diff-8daab1045cecf480419b0a31c2f8558e51197d31a7be910386ef4e420db0f7a4R310**

AshleyDumaine (approved):
LGTM

PR #525, opened 2024-09-27
[feat] VLAN support for CAPL clusters

tchinmai7:

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


api/v1alpha2/linodecluster_types.go
```diff
@@ -43,6 +43,11 @@ type LinodeClusterSpec struct {
 	// +optional
 	Network NetworkSpec `json:"network"`
 
+	// UseVlan provisions a cluster that uses VLANs instead of VPCs. IPAM is managed internally.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// +optional
+	UseVlan bool `json:"useVlan"`
```
AshleyDumaine, line 49: Not sure if we want to keep this here or move it to the NetworkSpec

eljohnson92: agreed I think this makes sense in the NetworkSpec

controller/linodemachine_controller_helpers.go
```diff
@@ -345,6 +376,84 @@ func getFirewallID(ctx context.Context, machineScope *scope.MachineScope, logger
...
+		ipString = currentIp.String()
+	}
+	return ipString
+}
+
+func reserveNextIP(ctx context.Context, machineScope *scope.MachineScope, logger logr.Logger) (string, error) {
+	namespace := machineScope.Cluster.Namespace
+	clusterName := machineScope.Cluster.Name
+	var nextIP string
+	var ipsMap corev1.ConfigMap
```
AshleyDumaine, line 398: You'll need to update linodemachine_controller.go to include configmaps for the kubebuilder RBAC annotation

tchinmai7: done, thanks!

controller/linodemachine_controller.go
```diff
@@ -101,6 +101,7 @@ type LinodeMachineReconciler struct {
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machines,verbs=get;watch;list
 // +kubebuilder:rbac:groups="",resources=events,verbs=get;list;watch;create;update;patch
 // +kubebuilder:rbac:groups="",resources=secrets;,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=configmaps;,verbs=get;watch;list;create;update;patch
```
AshleyDumaine, line 104:
```suggestion
// +kubebuilder:rbac:groups="",resources=configmaps;,verbs=get;watch;list;create;update;patch;delete
```

AshleyDumaine:
For this to work we will need [this PR merged](https://github.com/linode/linode-cloud-controller-manager/pull/236), a release for the CCM cut, and the CCM version bumped here, otherwise external IPs will not get set on the Nodes.

templates/flavors/rke2/vlan/kustomization.yaml
```diff
@@ -0,0 +1,148 @@
...
+      group: bootstrap.cluster.x-k8s.io
+      version: v1beta1
+      kind: RKE2ConfigTemplate
+    patch: |-
+      - op: replace
+        path: /spec/template/spec/preRKE2Commands
+        value:
+          - |
+            mkdir -p /etc/rancher/rke2/config.yaml.d/
+            echo "node-ip: $(hostname -I | grep -oE 10\.0\.[0-9]+\.[0-9]+)" >> /etc/rancher/rke2/config.yaml.d/capi-config.yaml
```
AshleyDumaine, line 116:
```suggestion
            echo "node-ip: $(hostname -I | grep -oE ^10\.0\.[0-9]+\.[0-9]+)" >> /etc/rancher/rke2/config.yaml.d/capi-config.yaml
```
avoids an edge case where extra IPs get matched that shouldn't

templates/flavors/rke2/vlan/kustomization.yaml
```diff
@@ -0,0 +1,148 @@
...
+      group: controlplane.cluster.x-k8s.io
+      version: v1beta1
+      kind: RKE2ControlPlane
+    patch: |-
+      - op: replace
+        path: /spec/preRKE2Commands
+        value:
+          - |
+            mkdir -p /etc/rancher/rke2/config.yaml.d/
+            echo "node-ip: $(hostname -I | grep -oE 10\.0\.[0-9]+\.[0-9]+)" >> /etc/rancher/rke2/config.yaml.d/capi-config.yaml
```
AshleyDumaine, line 85:
```suggestion
            echo "node-ip: $(hostname -I | grep -oE ^10\.0\.[0-9]+\.[0-9]+)" >> /etc/rancher/rke2/config.yaml.d/capi-config.yaml
```

controller/linodemachine_controller_helpers.go
```diff
@@ -345,6 +377,95 @@ func getFirewallID(ctx context.Context, machineScope *scope.MachineScope, logger
...
+			break
+		}
+		currentIp = currentIp.Next()
+		ipString = currentIp.String()
+	}
+	return ipString
+}
+
+func createIPsConfigMap(ctx context.Context, machineScope *scope.MachineScope, ip string) error {
+	return machineScope.Client.Create(ctx, &corev1.ConfigMap{
```
eljohnson92, line 396: can you give an example of what is in this configmap? I wonder if we could use part of the LInodeCluster spec for this information instead

tchinmai7: it is a simple map of machineName -> IP address
```
apiVersion: v1
data:
  schinmai-test-control-plane-f4g7c: 10.0.0.3
  schinmai-test-control-plane-mfxh6: 10.0.0.2
  schinmai-test-control-plane-mhczj: 10.0.0.1
kind: ConfigMap
metadata:
  creationTimestamp: "2024-10-02T19:12:38Z"
  labels:
    clusterctl.cluster.x-k8s.io/move: "true"
  name: schinmai-test-ips
  namespace: default
  resourceVersion: "7452"
  uid: cf59bb0a-4340-4461-8317-c02748d72050
```

tchinmai7: I opted for a configmap so that we can have a record of what IP was assigned to which machine for updates (the linodeMachine spec doesn't have it)

templates/flavors/rke2/vlan/kustomization.yaml
```diff
@@ -0,0 +1,148 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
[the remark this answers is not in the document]
tchinmai7: full includes the etcd backups, autoscaling etc are all part of the full template - do we not want it?

controller/linodecluster_controller.go
```diff
@@ -220,6 +220,23 @@ func (r *LinodeClusterReconciler) reconcileCreate(ctx context.Context, logger lo
 	return nil
 }
 
+func (r *LinodeClusterReconciler) cleanupVlanConfigmaps(ctx context.Context, logger logr.Logger, clusterScope *scope.ClusterScope) error {
+	if clusterScope.LinodeCluster.Spec.Network.UseVlan {
```
AshleyDumaine, line 224: we should probably tweak this to something like:
```	
	if !clusterScope.LinodeCluster.Spec.Network.UseVlan {
		return nil
	}
	...
```

controller/linodecluster_controller.go
```diff
@@ -220,6 +220,23 @@ func (r *LinodeClusterReconciler) reconcileCreate(ctx context.Context, logger lo
...
+		var ipsMap corev1.ConfigMap
+		err := clusterScope.Client.Get(ctx, client.ObjectKey{Namespace: clusterScope.Cluster.Namespace, Name: fmt.Sprintf("%s-ips", clusterScope.Cluster.Name)}, &ipsMap)
+		if err != nil && !apierrors.IsNotFound(err) {
+			logger.Error(err, "failed to get ips configmap")
+			return err
+		}
+		err = clusterScope.Client.Delete(ctx, &ipsMap)
```
AshleyDumaine: Do we need this extra get if we're ignoring not found errors?
e.g.:
```
if err := clusterScope.Client.Delete(ctx, &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Namespace: clusterScope.Cluster.Namespace, Name: fmt.Sprintf("%s-ips", clusterScope.Cluster.Name)}}); err != nil && !apierrors.IsNotFound(err) {
			logger.Error(err, "failed to delete ips configmap")
			return err
		}
```

templates/flavors/rke2/vlan/kustomization.yaml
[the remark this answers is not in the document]
AshleyDumaine: No, this kustomization should only include patches needed to run the default flavor with VLAN support instead of VPCs (might want to base off vpcless instead)

templates/flavors/rke2/vlan/kustomization.yaml
```diff
@@ -0,0 +1,148 @@
...
+        name: ${VPC_NAME:=${CLUSTER_NAME}}
+  - target:
+      group: infrastructure.cluster.x-k8s.io
+      version: v1alpha2
+      kind: LinodeCluster
+    patch: |-
+      - op: remove
+        path: /spec/vpcRef
+      - op: add
+        path: /spec/useVlan
```
AshleyDumaine, line 147: `path: /spec/network/useVlan`

tchinmai7: good catch, thanks!

templates/flavors/rke2/vlan/kustomization.yaml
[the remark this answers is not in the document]
tchinmai7: VPCless has a very different CiliumPolicy (no direct routing etc) - could build on default though

controller/linodemachine_controller_helpers.go
eljohnson92: could we just add this info to the status on each of the `LinodeMachines` in this case?

eljohnson92: still not totally sure on the consumer of this information so I might be missing part of the context

tchinmai7: this info is indeed saved into the `status` of linodeMachines - When X `linodeMachines` are reconciled in parallel, the linodeMacine controller needs to pick an IP for each instance, independently. To avoid assigning the same IP to multiple machines, I'm using the configmap as a locking mechanism, and a record of the IPs assigned to each machine. IF the configmap was sucessfully updated by the reconcile loop, it guaranteees that no other machine can get that IP and we eliminate conflicts

util/vlanips.go
```diff
@@ -0,0 +1,69 @@
...
+	vlanIPsMu sync.Mutex
+	// vlanIPsMap stores clusterName and a list of VlanIPs assigned to that cluster
+	vlanIPsMap  = make(map[string][]string, 0)
+	vlanIPRange = "10.0.0.0/8"
+)
+
+// GetNextVlanIP returns the next available IP for a cluster
+func GetNextVlanIP(clusterName, namespace string) string {
+	vlanIPsMu.Lock()
+	defer vlanIPsMu.Unlock()
```
rahulait, line 36: Instead of having one global lock, can we use two locks here? One for global map and second for individual cluster? I see the current implementation becoming a bottleneck if the cluster size increases as then we would be iterating for a longer duration to find the next free ip. I have a similar example here: https://github.com/linode/cluster-api-provider-linode/blob/main/util/ratelimits.go#L32-L43

tchinmai7: once a IP is assigned, its going to end up having to update the global map - so its going to have to lock the global map regardless, if i'm not missing something

tchinmai7: pushed up a change with a per-cluster lock - let me know what you think

api/v1alpha2/linodecluster_types.go
```diff
@@ -148,6 +148,11 @@ type NetworkSpec struct {
 	// additionalPorts contains list of ports to be configured with NodeBalancer.
 	// +optional
 	AdditionalPorts []LinodeNBPortConfig `json:"additionalPorts,omitempty"`
+
+	// UseVlan provisions a cluster that uses VLANs instead of VPCs. IPAM is managed internally.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// +optional
+	UseVlan bool `json:"useVlan"`
```
rahulait, line 155:
```suggestion
	UseVlan bool `json:"useVlan,omitempty"`
```

util/vlanips.go
```diff
@@ -0,0 +1,79 @@
...
+	"net/netip"
+	"slices"
+	"sync"
+)
+
+var (
+	vlanIPsMu sync.RWMutex
+	// vlanIPsMap stores clusterName and a list of VlanIPs assigned to that cluster
+	vlanIPsMap  = make(map[string]*ClusterIPs, 0)
+	vlanIPRange = "10.0.0.0/8"
```
rahulait, line 30: We are using /11 and /8 both. Not sure if thats intentional or we wanted to use just one.

tchinmai7: this is used to inform `netip.Prefix` what the network prefix is, and is what makes `nextIP` work - the `/11` is what gets assigned on the node - it could be much lower too

eljohnson92 (approved):

PR #531, opened 2024-10-03
[improvement] Add env var toggle for gzip compression of cloud-init user data

AshleyDumaine:
### NOTE: Using this toggle requires using OS images with an [updated cloud-init](https://github.com/canonical/cloud-init/commit/ce80781478f8e48d8edd13f69b8c1bc006114bde) version (>24.3.1)

**What this PR does / why we need it**: This adds support for gzipping the userData before base64 encoding it so we can fit more configuration into cloud-init.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controller/linodemachine_controller_helpers_test.go
+++ b/controller/linodemachine_controller_helpers_test.go
@@ -1,7 +1,10 @@
 package controller
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
+	"crypto/rand"
 	b64 "encoding/base64"
 	"fmt"
 	"testing"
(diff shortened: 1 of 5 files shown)


controller/linodemachine_controller_helpers_test.go
```diff
@@ -160,39 +168,6 @@ func TestSetUserData(t *testing.T) {
 				}}, nil)
 			},
 		},
-		{
```
eljohnson92, line 163: fwiw I think this is still a valid test case, I hope we don't hit it often but userdata can be too large still

AshleyDumaine: fair, just need to figure out how to make data that gzips to more than 16kB

eljohnson92: maybe we could decrease the limit for the sake of the test to validate a smaller gzip gets rejected? not sure if that would require too much refactoring to handle

eljohnson92 (approved):

PR #534, opened 2024-10-07
[improvement] : retry on ratelimit-reset than always waiting for 15 secs

rahulait:
First 5 POST requests take close to 5 secs or so. If we requeue subsequent requests with duration of 15 secs, then they will be next picked up at 20th second or so. This results in first 5 seconds of 15 sec window to not do anything. Since we are sleeping extra 5 or 6 secs in every 15 sec window, we end up taking additional 60+ seconds doing nothing when provisioning big clusters (say 100 node cluster). Ideally, we should sleep for 15 - 5 secs, i.e. 10 secs in this case. This PR takes account of ratelimit-reset value and uses that when requeueing.

Logs from cluster with new changes (bringing up 100 worker nodes)
```
k logs deploy/capl-controller-manager -n capl-system -f --timestamps > out.txt

cat out.txt  | grep "POST  /v4beta/linode/instances " | awk -F':' '{print $2}'  | sort -n | uniq -c | awk '{print $2 " - " $1}'
31 - 5
32 - 20
33 - 17
34 - 18
35 - 20
36 - 20
37 - 8

// First and last requests:
cat out.txt | grep "POST  /v4beta/linode/instances "

2024-10-07T19:31:49.369582068Z POST  /v4beta/linode/instances  HTTP/1.1
...
...
2024-10-07T19:37:29.415749236Z POST  /v4beta/linode/instances  HTTP/1.1
```
From above logs, total time taken for 100 node cluster: 5 min 40 secs.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/controller/linodemachine_controller_helpers.go
+++ b/controller/linodemachine_controller_helpers.go
@@ -695,7 +695,11 @@ func getDefaultInstanceConfig(ctx context.Context, machineScope *scope.MachineSc
 }
 
 // createInstance provisions linode instance after checking if the request will be within the rate-limits
-// Note: it takes a lock before checking for the rate limits and releases it after making request to linode API or when returning from function
+// Note:
+//  1. this method represents the critical section. It takes a lock before checking for the rate limits and releases it after making request to linode API or when returning from function
+//  2. returned time duration here is not always used.
+//     a) In case of an error, we calculate for how long to requeue in method which checks if its a transient error or not.
+//     b) If POST limit is reached, only then the returned time duration is used to retry after that time has elapsed.
 func createInstance(ctx context.Context, logger logr.Logger, machineScope *scope.MachineScope, createOpts *linodego.InstanceCreateOptions) (*linodego.Instance, time.Duration, error) {
 	ctr := util.GetPostReqCounter(machineScope.TokenHash)
 	ctr.Mu.Lock()
--- a/controller/linodemachine_controller_helpers.go
+++ b/controller/linodemachine_controller_helpers.go
@@ -27,6 +27,7 @@ import (
 	"net/netip"
 	"slices"
 	"sort"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/google/uuid"
@@ -695,16 +696,17 @@ func getDefaultInstanceConfig(ctx context.Context, machineScope *scope.MachineSc
 
 // createInstance provisions linode instance after checking if the request will be within the rate-limits
 // Note: it takes a lock before checking for the rate limits and releases it after making request to linode API or when returning from function
-func createInstance(ctx context.Context, logger logr.Logger, machineScope *scope.MachineScope, createOpts *linodego.InstanceCreateOptions) (*linodego.Instance, error) {
+func createInstance(ctx context.Context, logger logr.Logger, machineScope *scope.MachineScope, createOpts *linodego.InstanceCreateOptions) (*linodego.Instance, time.Duration, error) {
 	ctr := util.GetPostReqCounter(machineScope.TokenHash)
 	ctr.Mu.Lock()
 	defer ctr.Mu.Unlock()
 
 	if ctr.IsPOSTLimitReached() {
 		logger.Info(fmt.Sprintf("Cannot make more POST requests as rate-limit is reached (%d per %v seconds). Waiting and retrying after %v seconds", reconciler.SecondaryPOSTRequestLimit, reconciler.SecondaryLinodeTooManyPOSTRequestsErrorRetryDelay, reconciler.SecondaryLinodeTooManyPOSTRequestsErrorRetryDelay))
-		return nil, util.ErrRateLimit
+		return nil, ctr.RetryAfter(), util.ErrRateLimit
 	}
 
 	machineScope.LinodeClient.OnAfterResponse(ctr.ApiResponseRatelimitCounter)
-	return machineScope.LinodeClient.CreateInstance(ctx, *createOpts)
+	inst, err := machineScope.LinodeClient.CreateInstance(ctx, *createOpts)
+	return inst, time.Duration(reconciler.SecondaryLinodeTooManyPOSTRequestsErrorRetryDelay.Seconds()), err
 }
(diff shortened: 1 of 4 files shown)


controller/linodemachine_controller_helpers.go
```diff
@@ -684,16 +685,17 @@ func getDefaultInstanceConfig(ctx context.Context, machineScope *scope.MachineSc
...
 	if ctr.IsPOSTLimitReached() {
 		logger.Info(fmt.Sprintf("Cannot make more POST requests as rate-limit is reached (%d per %v seconds). Waiting and retrying after %v seconds", reconciler.SecondaryPOSTRequestLimit, reconciler.SecondaryLinodeTooManyPOSTRequestsErrorRetryDelay, reconciler.SecondaryLinodeTooManyPOSTRequestsErrorRetryDelay))
-		return nil, util.ErrRateLimit
+		return nil, ctr.RetryAfter(), util.ErrRateLimit
 	}
 
 	machineScope.LinodeClient.OnAfterResponse(ctr.ApiResponseRatelimitCounter)
-	return machineScope.LinodeClient.CreateInstance(ctx, *createOpts)
+	inst, err := machineScope.LinodeClient.CreateInstance(ctx, *createOpts)
+	return inst, time.Duration(reconciler.SecondaryLinodeTooManyPOSTRequestsErrorRetryDelay.Seconds()), err
```
AshleyDumaine, line 700: Is this time only useful if the error is non-nil? If so we should add a comment explaining that

util/ratelimits.go
```diff
@@ -58,25 +58,30 @@ func (c *PostRequestCounter) ApiResponseRatelimitCounter(resp *resty.Response) e
...
+		c.RefreshTime = secondaryLimitRefreshTime
+	}
 	return nil
 }
 
 // IsPOSTLimitReached checks whether POST limits have been reached.
 func (c *PostRequestCounter) IsPOSTLimitReached() bool {
 	// TODO: Once linode API adjusts rate-limits, remove secondary rate limit and simplify accordingly
 	currTime := time.Now().Unix()
+	return currTime <= int64(c.RefreshTime) && (c.ReqRemaining == 0 || c.ReqRemaining == reconciler.SecondaryPOSTRequestLimit)
```
AshleyDumaine, line 75:
```suggestion
	return time.Now().Unix() <= int64(c.RefreshTime) && (c.ReqRemaining == 0 || c.ReqRemaining == reconciler.SecondaryPOSTRequestLimit)
```
Also would add a comment explaining why the `c.ReqRemaining == reconciler.SecondaryPOSTRequestLimit` check is needed

rahulait: Sure, let me update it. Basically if we have made 5 requests (5 remaining) or 10 requests (0 remaining), then we want to wait until rate-limit has reset for that window.

util/ratelimits.go
```diff
@@ -58,25 +58,30 @@ func (c *PostRequestCounter) ApiResponseRatelimitCounter(resp *resty.Response) e
...
-		return true
+// RetryAfter returns how long to wait in seconds for rate-limit to reset
+func (c *PostRequestCounter) RetryAfter() time.Duration {
+	currTime := time.Now().Unix()
+	if currTime > int64(c.RefreshTime) {
+		return 0
 	}
-	return false
+	return time.Duration(c.RefreshTime-int(currTime)+1) * time.Second
 }
```
AshleyDumaine, line 85: I think using time instead of casting to int / int64 makes this more readable, e.g.:
```
type PostRequestCounter struct {
	Mu           sync.RWMutex
	ReqRemaining int
	RefreshTime  time.Time
}
```

```
func (c *PostRequestCounter) IsPOSTLimitReached() bool {
	// TODO: Once linode API adjusts rate-limits, remove secondary rate limit and simplify accordingly
	return time.Now().Before(c.RefreshTime) && (c.ReqRemaining == 0 || c.ReqRemaining == reconciler.SecondaryPOSTRequestLimit)
}
```

```
func (c *PostRequestCounter) RetryAfter() time.Duration {
	currTime := time.Now()
	if currTime.After(c.RefreshTime) {
		return 0
	}
	return c.RefreshTime.Sub(currTime)
}
```

rahulait: updated code with the suggestion

util/ratelimits.go
rahulait: Updated

controller/linodemachine_controller_helpers.go
AshleyDumaine: Sounds good, thanks for the clarification!

util/ratelimits.go
```diff
@@ -54,29 +54,35 @@ func (c *PostRequestCounter) ApiResponseRatelimitCounter(resp *resty.Response) e
...
 		return err
 	}
+	c.RefreshTime = time.Unix(int64(epochTime), 0)
+	secondaryRefreshTime := time.Unix(int64(epochTime)-int64(reconciler.SecondaryLinodeTooManyPOSTRequestsErrorRetryDelay.Seconds()), 0)
+
+	// TODO: remove when rate-limits are simplified
+	currTime := time.Now()
+	if c.ReqRemaining >= reconciler.SecondaryPOSTRequestLimit && currTime.Before(secondaryRefreshTime) {
+		c.RefreshTime = secondaryRefreshTime
+	}
```
AshleyDumaine, line 68:
```suggestion
	epochTime, err := strconv.ParseInt(resp.Header().Get("X-Ratelimit-Reset"), 10, 64)
	if err != nil {
		return err
	}
	c.RefreshTime = time.Unix(epochTime, 0)
	// We Add a negative number because for some reason Add returns a Time but Sub returns a Duration...
	secondaryRefreshTime := time.Unix(epochTime, 0).Add(reconciler.SecondaryLinodeTooManyPOSTRequestsErrorRetryDelay * -1)
	// TODO: remove when rate-limits are simplified
	currTime := time.Now()
	if c.ReqRemaining >= reconciler.SecondaryPOSTRequestLimit && currTime.Before(secondaryRefreshTime) {
		c.RefreshTime = secondaryRefreshTime
	}
```
to get rid of some of the int64 casting

rahulait: thanks, updated.

AshleyDumaine (approved):
LGTM, thanks!

PR #535, opened 2024-10-08
[improvement] : add conditions to check if linodevpc and linodefirewall are ready

rahulait:
This PR adds conditions for linodeVPC and linodeFirewall so that we can check if they are ready before continuing with provisioning of cluster/machines.

Fixes #

**TODOs**:
Add unittests

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/controller/linodecluster_controller.go
+++ b/controller/linodecluster_controller.go
@@ -169,8 +169,8 @@ func (r *LinodeClusterReconciler) reconcile(
 	}
 
 	// Create
-	if !reconciler.ConditionTrue(clusterScope.LinodeCluster, ConditionPreflightLinodeVPCReady) {
-		if clusterScope.LinodeCluster.Spec.VPCRef != nil {
+	if clusterScope.LinodeCluster.Spec.VPCRef != nil {
+		if !reconciler.ConditionTrue(clusterScope.LinodeCluster, ConditionPreflightLinodeVPCReady) {
 			res, err := r.reconcilePreflightLinodeVPCCheck(ctx, logger, clusterScope)
 			if err != nil || !res.IsZero() {
 				conditions.MarkFalse(clusterScope.LinodeCluster, ConditionPreflightLinodeVPCReady, string("linode vpc not yet available"), clusterv1.ConditionSeverityError, "")
@@ -178,7 +178,6 @@ func (r *LinodeClusterReconciler) reconcile(
 			}
 		}
 		conditions.MarkTrue(clusterScope.LinodeCluster, ConditionPreflightLinodeVPCReady)
-		return ctrl.Result{}, nil
 	}
 
 	if clusterScope.LinodeCluster.Spec.ControlPlaneEndpoint.Host == "" {
@@ -230,7 +229,7 @@ func (r *LinodeClusterReconciler) reconcilePreflightLinodeVPCCheck(ctx context.C
 			return ctrl.Result{}, err
 		}
 		return ctrl.Result{RequeueAfter: reconciler.DefaultClusterControllerReconcileDelay}, nil
-	} else if !linodeVPC.Status.Ready || linodeVPC.Spec.VPCID == nil {
+	} else if !linodeVPC.Status.Ready {
 		logger.Info("LinodeVPC is not yet available")
 		return ctrl.Result{RequeueAfter: reconciler.DefaultClusterControllerReconcileDelay}, nil
 	}
--- a/controller/linodecluster_controller.go
+++ b/controller/linodecluster_controller.go
@@ -170,11 +170,15 @@ func (r *LinodeClusterReconciler) reconcile(
 
 	// Create
 	if !reconciler.ConditionTrue(clusterScope.LinodeCluster, ConditionPreflightLinodeVPCReady) {
-		res, err := r.reconcilePreflightLinodeVPCCheck(ctx, logger, clusterScope)
-		if err != nil || !res.IsZero() {
-			conditions.MarkFalse(clusterScope.LinodeCluster, ConditionPreflightLinodeVPCReady, string("linode vpc not yet available"), clusterv1.ConditionSeverityError, "")
-			return res, err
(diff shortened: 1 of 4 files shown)


controller/linodemachine_controller.go
```diff
@@ -214,7 +216,7 @@ func (r *LinodeMachineReconciler) reconcile(ctx context.Context, logger logr.Log
 	// Make sure bootstrap data is available and populated.
 	if machineScope.Machine.Spec.Bootstrap.DataSecretName == nil {
 		logger.Info("Bootstrap data secret is not yet available")
-		conditions.MarkFalse(machineScope.LinodeMachine, ConditionPreflightMetadataSupportConfigured, WaitingForBootstrapDataReason, clusterv1.ConditionSeverityInfo, "")
+		conditions.MarkFalse(machineScope.LinodeMachine, ConditionPreflightLinodeFirewallReady, WaitingForBootstrapDataReason, clusterv1.ConditionSeverityInfo, "")
```
AshleyDumaine, line 219: I think we still want to leave this line as is?

controller/linodecluster_controller.go
```diff
@@ -192,6 +210,34 @@ func (r *LinodeClusterReconciler) reconcile(
...
+	}
+	if err := clusterScope.Client.Get(ctx, client.ObjectKeyFromObject(&linodeVPC), &linodeVPC); err != nil {
+		logger.Error(err, "Failed to fetch LinodeVPC")
+		if reconciler.RecordDecayingCondition(clusterScope.LinodeCluster,
+			ConditionPreflightLinodeVPCReady, string(cerrs.CreateClusterError), err.Error(),
+			reconciler.DefaultTimeout(r.ReconcileTimeout, reconciler.DefaultClusterControllerReconcileTimeout)) {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{RequeueAfter: reconciler.DefaultClusterControllerReconcileDelay}, nil
+	} else if !linodeVPC.Status.Ready || linodeVPC.Spec.VPCID == nil {
```
AshleyDumaine, line 233: Is it possible to have the status be Ready with a nil VPC ID?

controller/linodemachine_controller.go
```diff
@@ -287,6 +301,33 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
+	}
+	if err := machineScope.Client.Get(ctx, client.ObjectKeyFromObject(&linodeFirewall), &linodeFirewall); err != nil {
+		logger.Error(err, "Failed to find linode Firewall")
+		if reconciler.RecordDecayingCondition(machineScope.LinodeMachine,
+			ConditionPreflightLinodeFirewallReady, string(cerrs.CreateMachineError), err.Error(),
+			reconciler.DefaultTimeout(r.ReconcileTimeout, reconciler.DefaultMachineControllerWaitForPreflightTimeout)) {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{RequeueAfter: reconciler.DefaultMachineControllerRetryDelay}, nil
+	} else if !linodeFirewall.Status.Ready || linodeFirewall.Spec.FirewallID == nil {
```
AshleyDumaine, line 324: same as https://github.com/linode/cluster-api-provider-linode/pull/535/files#r1793647511

controller/linodecluster_controller.go
rahulait: I was re-using existing logic in other parts of capl-controller but I think it makes sense to just check for either statur of vpcid. I've updated the code to just check for status to be ready.

controller/linodemachine_controller.go
rahulait: updated

controller/linodemachine_controller.go
rahulait: We don't have any condition for bootstrap data secret. Here, we are returning the first condition we are going to hit. Previously, it was `ConditionPreflightCreated`. When we added step to calculate whether to use cloud-init or stackscripts, we changed it to first condition which was `ConditionPreflightMetadataSupportConfigured `. Since now we are checking if firewall condition is successful or not, we are changing it to that first condition. Do you think we should add another condition here for bootstrap secret and return accordingly?

AshleyDumaine: We do have a condition for the bootstrap data since it takes a few seconds to create. I don't think it really matters which one comes first since they are both needed.

rahulait: Added new condition for this. Showing `ConditionPreflightMetadataSupportConfigured ` or `ConditionPreflightLinodeFirewallReady` with reason `WaitingForBootstrapData` in status is a bit confusing, hence thought we should have its own condition `ConditionPreflightBootstrapDataSecretReady`

controller/linodemachine_controller.go
```diff
@@ -287,6 +301,33 @@ func (r *LinodeMachineReconciler) reconcileCreate(
 	return ctrl.Result{}, nil
 }
 
+func (r *LinodeMachineReconciler) reconcilePreflightLinodeFirewallCheck(ctx context.Context, logger logr.Logger, machineScope *scope.MachineScope) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
rahulait: Thats a good point. We would like to keep our linode API calls as minimum as possible. Since this falls in linodemachine_controller, doing GET before provisioning would mean we would be doing this for each instance, reconciles on error would mean we'll be redoing the calls and I feel we can simply avoid it and rely on k8s having the linodefirewall ready/not-ready to decide if firewall is available or not. We do search based on firewall name, a typo from customer would mean firewall not found and will get stuck with firewall not ready condition. In worst case if it matches with another linode firewall in cluster (same name and in same namespace which has a very low probability), thats a customer configuration and I would expect us to honor it. Hence leaning towards not making an API call here.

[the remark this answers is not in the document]
AshleyDumaine: Yeah agreed. Sounds good to leave as is 👍

AshleyDumaine (approved):
LGTM

PR #539, opened 2024-10-10
[Feat] Enable adding NodeBalancer to Linode Firewall

komer3:

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


templates/infra/linodeCluster.yaml
```diff
@@ -11,3 +11,7 @@ spec:
     apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
     kind: LinodeVPC
     name: ${VPC_NAME:=${CLUSTER_NAME}}
+  firewallRef:
```
eljohnson92, line 14: I think we will need a patch to the dns-based loadbal flavor to patch this out since it only will work for nodebalancers

api/v1alpha2/linodecluster_types.go
```diff
@@ -47,6 +47,11 @@ type LinodeClusterSpec struct {
 	// +optional
 	VPCRef *corev1.ObjectReference `json:"vpcRef,omitempty"`
 
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// +optional
+	// FirewallRef is a reference to a firewall object. This makes the linode use the specified firewall.
+	FirewallRef *corev1.ObjectReference `json:"firewallRef,omitempty"`
+
```
eljohnson92, line 54: I think we want to rename this and firewallID so it is specific to just nodebalancers since other loadbalancers will not support this

templates/infra/linodeCluster.yaml
komer3: ah, good catch!

api/v1alpha2/linodecluster_types.go
komer3: hmmm any thoughts on what we should rename it to? May be `LinodeFirewallRef`? 

Shouldn't we keep things consistent for the FirewallRef and FirewallID to what we are doing for Linode Machine?

rahulait: Since this firewall is used only by nodebalancers, we should rename it as `NodebalancerFirewallRef` or something similar. Someone might get confused seeing firewall ref in cluster and in linodemachine as they both can differ from one another.

templates/infra/linodeCluster.yaml
komer3: Patched out the nb firewall out of dns-lb, cilium-bgp-lb, and dual-stack flavors

api/v1alpha2/linodecluster_types.go
```diff
@@ -47,6 +47,11 @@ type LinodeClusterSpec struct {
 	// +optional
 	VPCRef *corev1.ObjectReference `json:"vpcRef,omitempty"`
 
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
```
rahulait, line 50: I wonder if we should make it immutable or it can be changed later on. I am ok with either approach, just want to know what others think about it.

cloud/services/loadbalancers.go
```diff
@@ -35,15 +37,69 @@ func EnsureNodeBalancer(ctx context.Context, clusterScope *scope.ClusterScope, l
...
 	logger.Info(fmt.Sprintf("Creating NodeBalancer %s", clusterScope.LinodeCluster.Name))
+
 	createConfig := linodego.NodeBalancerCreateOptions{
 		Label:  util.Pointer(clusterScope.LinodeCluster.Name),
 		Region: clusterScope.LinodeCluster.Spec.Region,
 		Tags:   []string{string(clusterScope.LinodeCluster.UID)},
 	}
 
+	// get firewall ID from firewallRef if it exists
+	if clusterScope.LinodeCluster.Spec.NodeBalancerFirewallRef != nil {
```
rahulait, line 48: Can we add it as a condition for linodecluster like we did for VPC?
https://github.com/linode/cluster-api-provider-linode/blob/main/controller/linodecluster_controller.go#L172-L181
https://github.com/linode/cluster-api-provider-linode/blob/main/controller/linodemachine_controller.go#L250-L258

komer3: ah nice. I was wondering if I should do it like that. I'll switch it

api/v1alpha2/linodecluster_types.go
komer3: I think either is fine. Not making it immutable makes it more flexible but not sure what use cases that would be useful in

controller/linodecluster_controller.go
```diff
@@ -180,6 +181,18 @@ func (r *LinodeClusterReconciler) reconcile(
...
+	// Check if the firewall is created
+	if clusterScope.LinodeCluster.Spec.NodeBalancerFirewallRef != nil {
+		if !reconciler.ConditionTrue(clusterScope.LinodeCluster, ConditionPreflightLinodeFirewallReady) {
+			res, err := r.reconcilePreflightLinodeFirewallCheck(ctx, logger, clusterScope)
+			if err != nil || !res.IsZero() {
+				conditions.MarkFalse(clusterScope.LinodeCluster, ConditionPreflightLinodeFirewallReady, string("linode firewall not yet available"), clusterv1.ConditionSeverityError, "")
+				return res, err
+			}
+		}
+		conditions.MarkTrue(clusterScope.LinodeCluster, ConditionPreflightLinodeFirewallReady)
```
rahulait, line 193: Should we change this to `ConditionPreflightNodebalancerFirewallReady`?

komer3: Hmmm good question! Makes sense to switch it since we are dealing with different firewalls here. Let me update that

komer3: Done!

cloud/services/loadbalancers.go
komer3: Done!

controller/linodemachine_controller.go
```diff
@@ -62,6 +62,7 @@ const (
 	// conditions for preflight instance creation
 	ConditionPreflightBootstrapDataSecretReady  clusterv1.ConditionType = "PreflightBootstrapDataSecretReady"
 	ConditionPreflightLinodeFirewallReady       clusterv1.ConditionType = "PreflightLinodeFirewallReady"
+	ConditionPreflightLinodeNBFirewallReady     clusterv1.ConditionType = "PreflightLinodeNBFirewallReady"
```
rahulait, line 65: Can we move this to here? https://github.com/linode/cluster-api-provider-linode/blob/main/controller/linodecluster_controller.go#L56

komer3: Done!

rahulait (dismissed):
lgtm

templates/flavors/kubeadm/konnectivity/nb-firewall.yaml
```diff
@@ -0,0 +1,74 @@
...
+  inboundRules:
+    - action: ACCEPT
+      addresses:
+        ipv4:
+          - 0.0.0.0/0
+        ipv6:
+          - ::/0
+      description: accept all api server related traffic from nodebalancers
+      label: inbound-api-server
+      ports: ${APISERVER_PORT:="6443"}, ${KONNECTIVITY_PORT:=8132}
```
rahulait, line 73: I wonder if we need quotes here for both ports or not. Just looks odd that one has quotes, other didn't

komer3: Good catch! We needed that. One of the e2e test was failing because of that

rahulait (dismissed):

eljohnson92 (dismissed):

eljohnson92 (approved):

PR #540, opened 2024-10-11
[improvement] : simplify rate limits as we now have only one limit for POST calls

rahulait:
Since the rate limits are now simplified to 5 req/15 secs, this PR updates CAPL to use that. Logs from a deployed cluster when worker nodes are scaled up from 0 to 100 nodes
```
cat logs/14oct/100nodes1.txt | grep "POST  /v4beta/linode/instances " | wc -l
100

cat logs/14oct/100nodes1.txt | grep "POST  /v4beta/linode/instances " | awk -F':' '{print $2}'  | sort -n | uniq -c | awk '{print $2 " - " $1}'
21 - 2
22 - 18
23 - 20
24 - 17
25 - 18
26 - 19
27 - 6
```

Time it took to bring up 100 nodes: 5 min 40 secs

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controller/linodemachine_controller_helpers.go
+++ b/controller/linodemachine_controller_helpers.go
@@ -63,21 +63,10 @@ var (
 	errNoPublicIPv6SLAACAddrs = errors.New("no public SLAAC address set")
 )
 
-func handleTooManyRequestsError(err error) (ctrl.Result, error) {
-	newErr := linodego.NewError(err)
-	if newErr.Response == nil {
-		return ctrl.Result{RequeueAfter: reconciler.DefaultLinodeTooManyRequestsErrorRetryDelay}, nil
-	}
-	if newErr.Response.Request.Method != http.MethodPost {
-		return ctrl.Result{RequeueAfter: reconciler.DefaultLinodeTooManyRequestsErrorRetryDelay}, nil
-	}
-	return ctrl.Result{RequeueAfter: reconciler.DefaultLinodeTooManyPOSTRequestsErrorRetryDelay}, nil
-}
-
 func retryIfTransient(err error) (ctrl.Result, error) {
 	if util.IsRetryableError(err) {
 		if linodego.ErrHasStatus(err, http.StatusTooManyRequests) {
-			return handleTooManyRequestsError(err)
+			return ctrl.Result{RequeueAfter: reconciler.DefaultLinodeTooManyRequestsErrorRetryDelay}, nil
 		}
 		return ctrl.Result{RequeueAfter: reconciler.DefaultMachineControllerRetryDelay}, nil
 	}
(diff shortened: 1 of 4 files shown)


util/reconciler/defaults.go
```diff
@@ -66,14 +66,10 @@ const (
...
-	// DefaultLinodeTooManyPOSTRequestsErrorRetryDelay is the default requeue delay if there is Linode API error for POST request. Currently, it is set to 10 requests per 30 seconds
-	DefaultLinodeTooManyPOSTRequestsErrorRetryDelay = 30 * time.Second
-	// SecondaryLinodeTooManyPOSTRequestsErrorRetryDelay is the secondary requeue delay if there is Linode API error for POST request. Currently, it is set to 5 requests per 15 seconds
-	SecondaryLinodeTooManyPOSTRequestsErrorRetryDelay = 15 * time.Second
-	// DefaultPOSTRequestLimit is the default limit of how many POST requests can be made to /linode/instances endpoint in 30 seconds before rate-limit reset.
-	DefaultPOSTRequestLimit = 10
-	// SecondaryPOSTRequestLimit is the secondary limit of how many POST requests can be made to /linode/instances endpoint in 15 seconds before rate-limit kicks in.
-	SecondaryPOSTRequestLimit = 5
+	// DefaultLinodeTooManyPOSTRequestsErrorRetryDelay is the default requeue delay if there is Linode API error for POST request
+	DefaultLinodeTooManyPOSTRequestsErrorRetryDelay = 15 * time.Second
```
eljohnson92, line 70: can this be taken directly from headers instead of hard coded here?

rahulait: We don't have this in the headers unfortunately. Here is a sample header:
```
[Access-Control-Allow-Credentials:[true] Access-Control-Allow-Headers:[Authorization, Origin, X-Requested-With, Content-Type, Accept, X-Filter] Access-Control-Allow-Methods:[HEAD, GET, OPTIONS, POST, PUT, DELETE] Access-Control-Allow-Origin:[*] Access-Control-Expose-Headers:[X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Status] Cache-Control:[max-age=0, no-cache, no-store] Connection:[keep-alive] Content-Length:[859] Content-Security-Policy:[default-src 'none'] Content-Type:[application/json] Date:[Mon, 14 Oct 2024 18:01:50 GMT] Expires:[Mon, 14 Oct 2024 18:01:50 GMT] Pragma:[no-cache] Retry-After:[13] Strict-Transport-Security:[max-age=31536000] Vary:[Authorization, X-Filter] X-Accepted-Oauth-Scopes:[linodes:read_write] X-Content-Type-Options:[nosniff] X-Customer-Uuid:[---UUID HERE---] X-Frame-Options:[DENY DENY] X-Oauth-Scopes:[*] X-Ratelimit-Limit:[5] X-Ratelimit-Remaining:[3] X-Ratelimit-Reset:[1728928924] X-Spec-Version:[4.189.1] X-Xss-Protection:[1; mode=block]]
```

rahulait: Updated the PR to remove default values.

controller/linodemachine_controller_helpers.go
```diff
@@ -63,21 +63,14 @@ var (
...
-func handleTooManyRequestsError(err error) (ctrl.Result, error) {
-	newErr := linodego.NewError(err)
-	if newErr.Response == nil {
-		return ctrl.Result{RequeueAfter: reconciler.DefaultLinodeTooManyRequestsErrorRetryDelay}, nil
-	}
-	if newErr.Response.Request.Method != http.MethodPost {
-		return ctrl.Result{RequeueAfter: reconciler.DefaultLinodeTooManyRequestsErrorRetryDelay}, nil
-	}
-	return ctrl.Result{RequeueAfter: reconciler.DefaultLinodeTooManyPOSTRequestsErrorRetryDelay}, nil
+func handleTooManyRequestsError() (ctrl.Result, error) {
```
AshleyDumaine, line 66: Since we're only using this in one place and it's one line, we should probably get rid of it

rahulait: Yeah, makes sense. Let me update it. The only drawback of not keeping the default limit of 15 secs in a var and using it to retry faster for POST calls is that if there are more than 1 clusters using token from same account, one can end up using all allowed rate-limit requests for the account and other will fail with 429 error. Since the call will fail, new code will simply retry after 1 min where it should have retried after 15 secs for POST calls. Though the possibility of this happening is low as its only for POST calls and it looks like it will be ok to wait for 1 min in such edge cases.

rahulait: Updated.

AshleyDumaine (approved):
LGTM

PR #548, opened 2024-10-24
[improvement] : store subnet id in linodeVPC and remove GetVPC call

rahulait:
This PR enables us to store subnet id in linodeVPC spec. This is useful as now we can rely on linodeVPC spec for subnet id than making a linode API call to get subnet id for the VPC.
Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


api/v1alpha2/linodevpc_types.go
```diff
@@ -54,6 +53,9 @@ type VPCSubnetCreateOptions struct {
 	Label string `json:"label,omitempty"`
 	// +optional
 	IPv4 string `json:"ipv4,omitempty"`
+	// ID is subnet id for the subnet
+	// +optional
+	ID int `json:"id,omitempty"`
```
komer3, line 58: question: would it make more sense to call this subnetID?

api/v1alpha2/linodevpc_types.go
```diff
@@ -54,6 +53,9 @@ type VPCSubnetCreateOptions struct {
 	Label string `json:"label,omitempty"`
 	// +optional
 	IPv4 string `json:"ipv4,omitempty"`
+	// ID is subnet id for the subnet
```
AshleyDumaine, line 56: Should we rename this to SubnetID for clarity?

controller/linodemachine_controller_helpers.go
```diff
@@ -403,29 +402,13 @@ func getVPCInterfaceConfig(ctx context.Context, machineScope *scope.MachineScope
...
-	if len(vpc.Subnets) == 0 {
+	if len(linodeVPC.Spec.Subnets) == 0 {
 		logger.Error(nil, "Failed to find subnet")
 
 		return nil, errors.New("failed to find subnet")
 	}
-	// Place node into the least busy subnet
-	sort.Slice(vpc.Subnets, func(i, j int) bool {
-		return len(vpc.Subnets[i].Linodes) > len(vpc.Subnets[j].Linodes)
-	})
```
komer3, line 426: question: because we are picking a subnet in linodevpc controller now and saving it in the spec we no longer need to look for least busy subnet? Just want to clarify!

api/v1alpha2/linodevpc_types.go
AshleyDumaine, line 56: Should we rename this to SubnetID for clarity similar to how we have VPCID instead of just ID?

controller/linodemachine_controller_helpers.go
rahulait: Yeah, we never used this logic and are not sure if we will ever need it in future. Currently, all our VPCs have single subnet and we use the first subnet returned by GetVPC call [here](https://github.com/linode/cluster-api-provider-linode/blob/main/controller/linodemachine_controller_helpers.go#L428). We don't want to make linode API GetVPC call for each instance creation as it slows things down and also causes ratelimiting issues if too many machines are brought up at the same time. Hence, we are getting rid of this call and simplifying our logic to use the one subnet we have.

komer3 (approved):
LGTM!

api/v1alpha2/linodevpc_types.go
rahulait: Yeah, sounds good. Updated

api/v1alpha2/linodevpc_types.go
eljohnson92: sorry to chime in on this late, I don't have a strong opinion here(and I think we break this other places as well) but for what it's worth `subnet.subnetID` is redundant but I'm fine with keeping it if that is the standard we have elsewhere, just wanted to call that out

eljohnson92 (approved):

AshleyDumaine (dismissed):

komer3 (approved):

AshleyDumaine (approved):

PR #551, opened 2024-10-24
[improvement] : simplify addr calculation and fetch instance config only for vlans

rahulait:
This PR simplifies buildInstanceAddrs function and makes getInstanceConfigs call only when required(in case of vlan). Otherwise, it relies on the output of `/v4/linode/instances/<id>/ips` to get all ips associated with the instance.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/controller/linodemachine_controller_helpers.go
+++ b/controller/linodemachine_controller_helpers.go
@@ -193,11 +193,13 @@ func buildInstanceAddrs(ctx context.Context, machineScope *scope.MachineScope, i
 	})
 
 	// check if a node has vpc specific ip and store it
-	if len(addresses.IPv4.VPC) != 0 {
-		ips = append(ips, clusterv1.MachineAddress{
-			Address: *addresses.IPv4.VPC[0].Address,
-			Type:    clusterv1.MachineInternalIP,
-		})
+	for _, vpcIP := range addresses.IPv4.VPC {
+		if *vpcIP.Address != "" {
+			ips = append(ips, clusterv1.MachineAddress{
+				Address: *vpcIP.Address,
+				Type:    clusterv1.MachineInternalIP,
+			})
+		}
 	}
 
 	if machineScope.LinodeCluster.Spec.Network.UseVlan {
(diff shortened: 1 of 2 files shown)


controller/linodemachine_controller_helpers.go
```diff
@@ -198,21 +192,30 @@ func buildInstanceAddrs(ctx context.Context, machineScope *scope.MachineScope, i
...
-				Address: netip.MustParsePrefix(iface.IPAMAddress).Addr().String(),
-				Type:    clusterv1.MachineInternalIP,
-			})
+	// check if a node has vpc specific ip and store it
+	if len(addresses.IPv4.VPC) != 0 {
+		ips = append(ips, clusterv1.MachineAddress{
+			Address: *addresses.IPv4.VPC[0].Address,
+			Type:    clusterv1.MachineInternalIP,
+		})
+	}
```
komer3, line 201: question: Just want to clarify, we are moving this logic out of the loop since we just have one VPC subnet?

rahulait: Previously, VPC ips were not returned in GetInstanceIPAddresses https://github.com/linode/linodego/blob/v1.30.0/instance_ips.go#L17-L22. This was implemented in newer versions and we now do get VPC specific ips in response of this call. This code is just to rely on that response. We now need ListInstanceConfigs only for VLAN specific ips as they are still not returned in GetInstanceIPAddresses call. The code restructures in a way that we make only one linode API call if we are not using VLANs.

rahulait: But your point is right, maybe I can add a for loop here to add all VPC ips? Currently we don't support more than one, so not sure if we really need it.

komer3: If we can just note that as a comment for future reference that should be fine. We can add the loop in future if we support more than one VPC ip

komer3 (approved):
LGTM!

komer3 (approved):

PR #552, opened 2024-10-24
[improvement] remove gob to save memory

AshleyDumaine:
**What this PR does / why we need it**: We're currently calling encode and decode for various objects which consumes a bit of CPU and memory. Instead, we explicitly convert the objects. This also bans the further use of gob going forward via depguard. 

Fixes #

**Special notes for your reviewer**: I also included the banning of reflect while enabling depguard. LMK if we for some reason don't want this.

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controller/linodemachine_controller_helpers.go
+++ b/controller/linodemachine_controller_helpers.go
@@ -17,10 +17,8 @@ limitations under the License.
 package controller
 
 import (
-	"bytes"
 	"context"
 	b64 "encoding/base64"
-	"encoding/gob"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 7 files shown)


controller/linodemachine_controller_helpers.go
```diff
@@ -431,21 +429,42 @@ func getVPCInterfaceConfig(ctx context.Context, machineScope *scope.MachineScope
...
+	for idx, iface := range machineSpec.Interfaces {
+		creatOpt := linodego.InstanceConfigInterfaceCreateOptions{
+			IPAMAddress: iface.IPAMAddress,
+			Label:       iface.Label,
+			Purpose:     iface.Purpose,
+			Primary:     iface.Primary,
+			SubnetID:    iface.SubnetID,
+			IPRanges:    iface.IPRanges,
+		}
+		if iface.IPv4 != nil && iface.IPv4.NAT1To1 != "" {
```
rahulait, line 442: I think we do support not having public ip on vpc interface if I am correct. Not sure this will work there.

controller/linodemachine_controller_helpers.go
```diff
@@ -431,21 +429,42 @@ func getVPCInterfaceConfig(ctx context.Context, machineScope *scope.MachineScope
...
+	for idx, iface := range machineSpec.Interfaces {
+		creatOpt := linodego.InstanceConfigInterfaceCreateOptions{
+			IPAMAddress: iface.IPAMAddress,
+			Label:       iface.Label,
+			Purpose:     iface.Purpose,
+			Primary:     iface.Primary,
+			SubnetID:    iface.SubnetID,
+			IPRanges:    iface.IPRanges,
+		}
+		if iface.IPv4 != nil {
```
rahulait, line 442: Instead of using iface.IPv4, should we use `iface.Purpose == linodego.InterfacePurposeVPC` here?

eljohnson92: this should be mapping machineSpec.Interfaces directly to instanceCOnfigOptions, I thought the logic to handle any interface specific configuration is handled in `getVPCInterfaceConfig`

rahulait: Yeah, first we have linodeMachineSpecToInstanceCreateConfig() and then getVPCInterfaceConfig(). I wonder why we need to set VPC config here.

rahulait:
Thank you for doing the changes and getting rid of gob. I am going to rebase the branch from main and have tests run one more time. Overall, the PR looks good to me.

rahulait (approved):
LGTM 🚀

PR #559, opened 2024-11-01
[feat] Add CIDR transformation for single IP addresses in firewall rules

komer3:

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


controller/linodefirewall_controller_helpers.go
```diff
@@ -157,11 +180,17 @@ func processACL(firewall *infrav1alpha2.LinodeFirewall) (
 		ruleIPv6s := []string{}
 
 		if rule.Addresses.IPv4 != nil {
-			ruleIPv4s = append(ruleIPv4s, *rule.Addresses.IPv4...)
+			// Transform each IPv4 address
+			for _, ip := range *rule.Addresses.IPv4 {
+				ruleIPv4s = append(ruleIPv4s, transformToCIDR(ip))
```
rahulait: Overall the code change looks good to me, not sure if we should validate if its valid ipv4 here or its getting done later on.

komer3: I was thinking of letting linode api do it? What are your thoughts?

rahulait: Sounds good to me. I got confused by the comment `// If not a valid IP, return as-is (will be validated later)` and thought if we missed validating.

rahulait (approved):
lgtm

PR #563, opened 2024-11-07
[fix] handle case where instance already exists

rahulait:
If create instance call fails due to context timeout error, instance gets provisioned on the linode side but CAPL thinks instance is still not created. It keeps on retrying and ends up hitting [400 label must be unique] error as it cannot provision another instance with same name. This results in CAPL continuously retrying to create the instance and ending up hitting rate-limits. This PR fixes the issue by doing a GET if instance already exists and using it during next reconciles.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/controller/linodemachine_controller_helpers.go
+++ b/controller/linodemachine_controller_helpers.go
@@ -722,6 +722,27 @@ func createInstance(ctx context.Context, logger logr.Logger, machineScope *scope
 
 	machineScope.LinodeClient.OnAfterResponse(ctr.ApiResponseRatelimitCounter)
 	inst, err := machineScope.LinodeClient.CreateInstance(ctx, *createOpts)
+
+	// if instance already exists, we get 400 response. get respective instance and return
+	if linodego.ErrHasStatus(err, http.StatusBadRequest) {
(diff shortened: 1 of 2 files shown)


controller/linodemachine_controller_helpers.go
```diff
@@ -721,6 +721,27 @@ func createInstance(ctx context.Context, logger logr.Logger, machineScope *scope
 
 	machineScope.LinodeClient.OnAfterResponse(ctr.ApiResponseRatelimitCounter)
 	inst, err := machineScope.LinodeClient.CreateInstance(ctx, *createOpts)
+
+	// if instance already exists, we get 400 response. get respective instance and return
+	if linodego.ErrHasStatus(err, http.StatusBadRequest) {
```
AshleyDumaine, line 726: 400 is a very generic error thrown by the linode API, could we check the error itself to confirm that an instance exists with the existing label before trying to fetch the instance to save some API calls?

AshleyDumaine (approved):
LGTM

komer3 (approved):

komer3 (approved):

PR #573, opened 2024-11-19
[fix] : avoid resetting condition with severity error to severity warning

rahulait:
This PR makes sure we don't reset condition severity from error to warning during next reconciles. Last transition time for condition changes when its updated from warning to error. During next reconcile, if we look at the last reconcile time only, then it will be the updated time and current logic was assuming that time limit has not been reached and was flipping the condition back from error to warning.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/util/reconciler/conditions.go
+++ b/util/reconciler/conditions.go
@@ -31,13 +31,12 @@ func HasConditionSeverity(from conditions.Getter, typ clusterv1.ConditionType, s
 }
 
 func RecordDecayingCondition(to conditions.Setter, typ clusterv1.ConditionType, reason, message string, timeout time.Duration) bool {
-	conditions.MarkFalse(to, typ, reason, clusterv1.ConditionSeverityWarning, "%s", message)
-
 	if HasStaleCondition(to, typ, timeout) {
 		conditions.MarkFalse(to, typ, reason, clusterv1.ConditionSeverityError, "%s", message)
 		return true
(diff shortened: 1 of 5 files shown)


util/reconciler/conditions.go
```diff
@@ -31,13 +31,12 @@ func HasConditionSeverity(from conditions.Getter, typ clusterv1.ConditionType, s
...
 
 func RecordDecayingCondition(to conditions.Setter, typ clusterv1.ConditionType, reason, message string, timeout time.Duration) bool {
-	conditions.MarkFalse(to, typ, reason, clusterv1.ConditionSeverityWarning, "%s", message)
-
 	if HasStaleCondition(to, typ, timeout) {
 		conditions.MarkFalse(to, typ, reason, clusterv1.ConditionSeverityError, "%s", message)
 		return true
 	}
 
+	conditions.MarkFalse(to, typ, reason, clusterv1.ConditionSeverityWarning, "%s", message)
```
AshleyDumaine, line 39: won't this set the severity to warning for a stale condition instead of error?

rahulait: `HasStaleCondition()` method returns true for stale condition and sets the severity to error and returns. SeverityWarning is only added if the condition doesn't exist or if its not stale.

PR #586, opened 2024-12-03
[improvement] add cloud-init boostrap via cluster object store

cbang-akamai:

This change introduces an optional **Cluster Object Store** in the `LinodeCluster` resource definition. The Cluster Object Store definition references an object storage bucket used for internal cluster operations, e.g. bootstrapping.

Currently, for all Cluster API bootstrap providers (e.g. [Cluster API Provider RKE2](https://github.com/rancher/cluster-api-provider-rke2)) , CAPL uses [cloud-init](https://cloud-init.io/) as the default configuration engine to bootstrap workload cluster machines. Cloud-init bootstrap configurations (cloud-configs) are deployed to the cluster's Linodes via the [Metadata service](https://techdocs.akamai.com/cloud-computing/docs/overview-of-the-metadata-service) as user data. Unfortunately, the Metadata service currently the limits amount of data the can be transmitted to a Linode. The restrictions are:

> Base64-encoded cloud-config data.
>
> Cannot be modified after provisioning. To update, use either the Clone a Linode or Rebuild a Linode operations.
>
> Must not be included when cloning to an existing Linode.
>
> **Unencoded data must not exceed 65535 bytes, or about 16kb encoded.**
>
> Source: https://techdocs.akamai.com/linode-api/reference/post-linode-instance

This limitation restricts the adoption of CAPL, as users may need to include custom configuration to in the bootstrap configuration of their Linodes, which can result in a cloud-config that exceeds this 16kB Metadata service limit. To bypass this limitation, this submission introduces the following Linode creation workflow changes:

1. During `LinodeMachine` creation, check the [Cluster API bootstrap Secret](https://release-1-3.cluster-api.sigs.k8s.io/developer/providers/bootstrap#bootstrap-secret) to determine if the generated bootstrap cloud-config fits within the Metadata or Stackscript service limits. 
2. If it does, continue the Linode creation as normal.
3. Otherwise, if the cloud-config is too large, upload the original cloud-config to the cluster's Object Store and generate a signed URL for the uploaded object.
4. Generate a "pointer" cloud-config (using cloud-init's [Include file](https:/cloudinit.readthedocs.io/en/latest/explanation/format.html#include-file) user data format) to deliver the cloud-config by reference to the Linode.
5. Pass the pointer cloud-config to the cloud-init bootstrap service instead. Cloud-init will read the pointer cloud-config after instance initialization and download the original cloud-config, thereby bypassing any service limits. 

Fixes #

**Verification:**

1. For your selected flavor, modify the bootstrap configuration to generate a large (>16kb) cloud-config:

```sh
$ tr -dc A-Za-z0-9 < /dev/urandom | head -c 32kB >> chonk.txt

$ kubectl create secret generic chonk-secret --from-file=chonk.txt

# Example: Kubeadm
$ cat 0001-DEBUG-kubeadm-add-dummy-thicc-file.patch
From ea5a83b70eab030fb9df2a708cc5a974ae97e967 Mon Sep 17 00:00:00 2001
From: Example <example@example.com>
Date: Tue, 3 Dec 2024 09:28:37 -0500
Subject: [PATCH] DEBUG: kubeadm: add dummy thicc file

---
 .../flavors/kubeadm/default/kubeadmConfigTemplate.yaml      | 6 ++++++
 1 file changed, 6 insertions(+)

diff --git a/templates/flavors/kubeadm/default/kubeadmConfigTemplate.yaml b/templates/flavors/kubeadm/default/kubeadmConfigTemplate.yaml
index 43e19db..3834eb3 100644
--- a/templates/flavors/kubeadm/default/kubeadmConfigTemplate.yaml
+++ b/templates/flavors/kubeadm/default/kubeadmConfigTemplate.yaml
@@ -6,6 +6,12 @@ metadata:
 spec:
   template:
     spec:
+      files:
+        - path: /chonk.txt
+          contentFrom:
+            secret:
+              key: chonk.txt
+              name: chonk-secret
       preKubeadmCommands:
         - curl -fsSL https://github.com/linode/cluster-api-provider-linode/raw/dd76b1f979696ef22ce093d420cdbd0051a1d725/scripts/pre-kubeadminit.sh | bash -s ${KUBERNETES_VERSION}
         - hostnamectl set-hostname '{{ ds.meta_data.label }}' && hostname -F /etc/hostname
--
2.25.1

$ git am --keep-cr 0001-DEBUG-kubeadm-add-dummy-thicc-file.patch
Applying: DEBUG: kubeadm: add dummy thicc file
```

2. Create a cluster:
```sh
$ make local-release

$ clusterctl generate cluster $CLUSTER_NAME --flavor kubeadm-full --kubernetes-version v1.29.1 --infrastructure local-linode:v0.0.0 | kubectl apply -f -
```

3.  Check all `Machine`(s) and `LindoeMachine`(s) are able to successfully provision:
```sh
$ kubectl get linodemachine
NAME                                  CLUSTER           STATE     READY   PROVIDERID          MACHINE
${CLUSTER_NAME}-control-plane-j6twt   ${CLUSTER_NAME}   running   true    linode://00000000   ${CLUSTER_NAME}-control-plane-j6twt
${CLUSTER_NAME}-md-0-4ttnz-9js7n      ${CLUSTER_NAME}   running   true    linode://00000000   ${CLUSTER_NAME}-md-0-4ttnz-9js7n

$ kubectl get machine
NAME                                  CLUSTER           NODENAME                              PROVIDERID          PHASE     AGE     VERSION
${CLUSTER_NAME}-control-plane-j6twt   ${CLUSTER_NAME}   ${CLUSTER_NAME}-control-plane-j6twt   linode://00000000   Running   12m     v1.29.1
${CLUSTER_NAME}-md-0-4ttnz-9js7n      ${CLUSTER_NAME}   ${CLUSTER_NAME}-md-0-4ttnz-9js7n      linode://00000000   Running   3m35s   v1.29.1
```

4. Check the Linode configuration:
```sh
$ export IP="$(kubectl get machine ${CLUSTER_NAME}-md-0-4ttnz-9js7n -oyaml | yq '.status.addresses[] | select (.type == "ExternalIP") | .address' | head -n 1)"

$ ssh root@$IP cloud-init query userdata && ls -lah /chonk.txt
#include
🏃‍♀️💨🔫👮🚔

-rw-r--r-- 1 root root 32K Dec  3 17:29 /chonk.txt
```

**TODOs**:

- [X] squashed commits
- [X] includes documentation
- [X] adds unit tests


cloud/scope/common.go
```diff
@@ -33,6 +38,9 @@ const (
 
 	// MaxBodySize is the max payload size for Akamai edge dns client requests
 	maxBody = 131072
+
+	// defaultObjectStorageSignedUrlExpiry is the default expiration for Object Storage signed URls
+	defaultObjectStorageSignedUrlExpiry = 900 * time.Second
```
AshleyDumaine, line 43: Could we do `15 * time.Minute` instead for clarity?

cloud/services/object_storage_objects.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	// Key by UUID for shared buckets.
+	key := string(mscope.LinodeMachine.ObjectMeta.UID)
+
+	if _, err := mscope.S3Client.PutObject(ctx, &s3.PutObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(key),
+		Body:   s3manager.ReadSeekCloser(bytes.NewReader(data)),
+		// TODO: Enable SSE?
+		// ServerSideEncryption: types.ServerSideEncryptionAes256,
```
AshleyDumaine, line 50: I would enable it unless we have to pass an encryption key for this, I forget if that's necessary since I looked at using SSE a bit over a year ago.

templates/infra/linodeCluster.yaml
```diff
@@ -15,3 +15,49 @@ spec:
     apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
     kind: LinodeFirewall
     name: ${CLUSTER_NAME}-nb
+  objectStore:
```
eljohnson92, line 18: I think we want a separate flavor for this rather than add it to the default flavor since this will increase the cost for creating a cluster

cloud/services/object_storage_objects.go
cbang-akamai: It looks like for the Linode Object Storage provided method ([SSE-C](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html)), the client needs to set specific SSE-related headers in the request ([1](https://aws.amazon.com/blogs/developer/generating-amazon-s3-pre-signed-urls-with-sse-c-part-4/), [2](https://www.linode.com/community/questions/22648/download-from-object-storage-using-sse-c-signed-urls)). This level of granularity is currently possible in cloud-init's [include file](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#include-file) user data format.

cloud/scope/common.go
```diff
@@ -104,6 +112,47 @@ func CreateLinodeClient(config ClientConfig, opts ...Option) (LinodeClient, erro
...
+			awsconfig.WithRegion("auto"),
+		}
+
+		clientOpts = []func(*s3.Options){}
+	)
+
+	// If we have a cluster object store bucket, get its configuration.
+	if cluster.Spec.ObjectStore != nil {
+		secret, err := getCredentials(ctx, crClient, cluster.Spec.ObjectStore.CredentialsRef, cluster.GetNamespace())
+		if err == nil {
```
AshleyDumaine, line 127: Curious why we're doing this instead of the usual `err != nil`

cloud/services/object_storage_objects.go
```diff
@@ -0,0 +1,135 @@
...
+		return "", errors.New("nil cluster object store")
+	}
+
+	if len(data) == 0 {
+		return "", errors.New("got empty data")
+	}
+
+	bucket, err := mscope.GetBucketName(ctx)
+	if err != nil || bucket == "" {
+		return "", errors.New("no bucket name")
```
AshleyDumaine, line 40: Could we define these in a var block up top similar to internal/controller/linodemachine_controller_helpers.go? Might be useful for tests to have these defined.

AshleyDumaine: Might also make sense to create a helper function to check all the required variables are set and that we can get the bucket since there seems to be some overlap in the DeleteObject function

cloud/services/object_storage_objects.go
```diff
@@ -0,0 +1,135 @@
...
+			ae  smithy.APIError
+			kne *types.NoSuchKey
+			bne *types.NoSuchBucket
+		)
+		switch {
+		// TODO: Check if this edge-case is also present with Linode Object Storage.
+		// In the case that the IAM policy does not have sufficient
+		// permissions to get the object, we will attempt to delete it
+		// anyway for backwards compatibility reasons.
+		case errors.As(err, &ae) && ae.ErrorCode() == "Forbidden":
```
AshleyDumaine, line 113: I'm surprised this is a string instead of an int. Is there at least some kind of const for this from smithy?

cloud/services/object_storage_objects_test.go
```diff
@@ -0,0 +1,287 @@
...
+					assert.Error(t, err)
+				}),
+				Result("nil s3 client", func(ctx context.Context, mck Mock) {
+					_, err := CreateObject(ctx, &scope.MachineScope{
+						LinodeCluster: &infrav1alpha2.LinodeCluster{
+							Spec: infrav1alpha2.LinodeClusterSpec{
+								ObjectStore: &infrav1alpha2.ObjectStore{
+									CredentialsRef: corev1.SecretReference{Name: "fake"}}}}},
+						[]byte("fake"), log)
+					assert.Error(t, err)
```
AshleyDumaine, line 44: When defining the errors as variables from https://github.com/linode/cluster-api-provider-linode/pull/586/files#r1876417132, we should also check the error is the expected error

cloud/scope/common.go
cbang-akamai: Because these extra configuration options in the AWS SDK clients are only set if the credentials are successfully retrieved.

internal/controller/linodemachine_controller.go
```diff
@@ -513,6 +513,14 @@ func (r *LinodeMachineReconciler) reconcileUpdate(ctx context.Context, logger lo
 		machineScope.LinodeMachine.Status.Ready = true
 		conditions.MarkTrue(machineScope.LinodeMachine, clusterv1.ReadyCondition)
 	}
+
+	// Clean up after instance creation.
+	if linodeInstance.Status == linodego.InstanceRunning && machineScope.Machine.Status.Phase == "Running" {
```
AshleyDumaine, line 518: Do we need the second check here?

cloud/services/object_storage_objects.go
cbang-akamai: > define these errors in a var block
QQ: Why do we want(?) this level of insight into the implementation details in our unit tests?

> create a helper function to check all the required variables are set
Sounds good!

internal/controller/linodemachine_controller.go
```diff
@@ -513,6 +513,14 @@ func (r *LinodeMachineReconciler) reconcileUpdate(ctx context.Context, logger lo
...
+		if err := deleteBootstrapData(ctx, machineScope); err != nil {
+			logger.Error(err, "Fail to bootstrap data")
```
AshleyDumaine, line 520: I think we might want to have deleteBootstrapData function log the deletion error. We might also not want to log it as an error if we're not doing anything with that error like requeuing to retry the deletion (though we might want to do that in the case we get some kind of error like rate-limiting?)

internal/controller/linodemachine_controller.go
cbang-akamai: Yeah, because the first status indicates of the state of the infrastructure and the second status is used to check if the node has successfully bootstrapped.

templates/flavors/rke2/full-vpcless/kustomization.yaml
```diff
@@ -7,6 +7,7 @@ resources:
   - ../../../addons/etcd-backup-restore
   - ../../../addons/machine-health-check
   - ../../../addons/placementGroups
+    - ../../../addons/cluster-object-store
```
AshleyDumaine, line 10:
```suggestion
  - ../../../addons/cluster-object-store
```

cloud/services/object_storage_objects.go
cbang-akamai: So I think the Smithy error type here is just a container -- the actual errors come (and are unique to) the AWS services themselves, e.g. [Amazon S3](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList).

internal/controller/linodemachine_controller_helpers.go
```diff
@@ -553,6 +537,81 @@ func setUserData(ctx context.Context, machineScope *scope.MachineScope, createCo
...
+	case size < limit:
+		return bootstrapdata, nil
+	// Compromise case (Metadata): Use compression.
+	case machineScope.LinodeMachine.Status.CloudinitMetadataSupport && gzipCompressionEnabled:
+		compressed, err = compressUserData(bootstrapdata)
+		// Fallback to Object Storage workaround on compression failure.
+		if err != nil {
+			logger.Info(fmt.Sprintf("Failed to compress bootstrap data: %v", err))
+			logger.Info("Falling back to Object Storage")
+			goto obj
```
AshleyDumaine, line 570: Can we instead have a helper function called something like `storeMetadataInObj`?

cbang-akamai: Yeah, I can pull out the logic _after_ this entire switch into it's own helper function. 

However, is the aim to remove the `goto` statements here? I think we still need to keep this since the aim of this logic is to bail out early when we know we're forced to used the Cluster Object Store workaround.

AshleyDumaine: Yeah I meant doing something like this:
```
	case machineScope.LinodeMachine.Status.CloudinitMetadataSupport && gzipCompressionEnabled:
		compressed, err = compressUserData(bootstrapdata)
		// Fallback to Object Storage workaround on compression failure.
		if err != nil {
			logger.Info(fmt.Sprintf("Failed to compress bootstrap data: %v", err))
			logger.Info("Falling back to Object Storage")
			return storeMetadataInObj(ctx, machineScope, logger, bootstrapdata)
		}
		size = len(compressed)
		if size < limit {
			return compressed, nil
		}
		return storeMetadataInObj(ctx, machineScope, logger, bootstrapdata)
	// Worst case: Object Storage workaround.
	default:
		logger.Info("decoded bootstrap data exceeds size limit", "limit", limit, "size", size)
		return storeMetadataInObj(ctx, machineScope, logger, bootstrapdata)
	}
```

cloud/services/object_storage_objects.go
cbang-akamai: I didn't want to litter the package with error globals here, so I just matched the error string in the tests 😇

internal/controller/linodemachine_controller_helpers.go
cbang-akamai: I reworked the logic inside the switch statement to:
- Reduce repetition between all the different edge-cases and preserve the overall "layered" visual structure
- Make the (now single) `goto` statement function more like a `break` out of the nested switch logic and the continue executing the rest of the function

cloud/services/object_storage_objects.go
```diff
@@ -0,0 +1,133 @@
...
+		return "", err
+	}
+	if len(data) == 0 {
+		return "", errors.New("empty data")
+	}
+
+	bucket, err := mscope.GetBucketName(ctx)
+	if err != nil || bucket == "" {
+		return "", errors.New("missing bucket name")
+	}
```
AshleyDumaine, line 45: We should separate these conditions so we can return the error from GetBucketName if there is one instead of replacing it with this other error

cloud/services/object_storage_objects.go
```diff
@@ -0,0 +1,133 @@
...
+
+func DeleteObject(ctx context.Context, mscope *scope.MachineScope) error {
+	if err := validateObjectScopeParams(mscope); err != nil {
+		return err
+	}
+
+	bucket, err := mscope.GetBucketName(ctx)
+	if err != nil || bucket == "" {
+		return errors.New("missing bucket name")
+	}
```
AshleyDumaine, line 87: same as https://github.com/linode/cluster-api-provider-linode/pull/586/files#r1880707999

AshleyDumaine (dismissed):
LGTM

AshleyDumaine (approved):

PR #588, opened 2024-12-06
[feat][improvement][docs] implement FirewallRule CRD and support in firewall controller, remove finalizers on AddressSets

AshleyDumaine:
**What this PR does / why we need it**: Adds a new CRD that can be reused to construct LinodeFirewalls in case rules are common between firewalls (e.g. common rules for control plane and worker node Firewalls)

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

Example using AddressSets and FirewallRules:
```
kind: LinodeFirewall
metadata:
  name: sample-fw
spec:
  enabled: true
  inboundPolicy: DROP
  inboundRules:
    - action: ACCEPT
      label: inbound-api-server
      ports: "6443"
      protocol: TCP
      addresses:
        ipv4:
          - "192.168.255.0/24"
    - action: ACCEPT
      label: intra-cluster
      ports: "1-65535"
      protocol: "TCP"
      addressSetRefs:  # Can be used together with .addresses if desired.
        - name: vpc-addrset
          kind: AddressSet
  inboundRuleRefs:  # Can be used together with .inboundRules if desired
    - name: example-fwrule-udp
      kind: FirewallRule
    - name: example-fwrule-icmp
      kind: FirewallRule
  # outboundRules: []
  # outboundRuleRefs: []
  # outboundPolicy: ACCEPT
---
kind: AddressSet
metadata:
  name: vpc-addrset
spec:
  ipv4:
    - "10.0.0.0/8"
---
kind: FirewallRule
metadata:
  name: example-fwrule-udp
spec:
  action: ACCEPT
  label: intra-cluster-udp
  ports: "1-65535"
  protocol: "UDP"
  addresses:
    ipv4:
      - "10.0.0.0/8"
---
kind: FirewallRule
metadata:
  name: example-fwrule-icmp
spec:
  action: ACCEPT
  label: intra-cluster-icmp
  protocol: "ICMP"
  addressSetRefs:  # Can be used together with .addresses if desired.
    - name: vpc-addrset
      kind: AddressSet
```


api/v1alpha2/linodefirewall_types.go
```diff
@@ -39,7 +38,12 @@ type LinodeFirewallSpec struct {
 	Enabled bool `json:"enabled,omitempty"`
 
 	// +optional
-	InboundRules []FirewallRule `json:"inboundRules,omitempty"`
+	InboundRules []FirewallRuleSpec `json:"inboundRules,omitempty"`
+
+	// InboundRuleRefs is a list of references to FirewallRules as an alternative to
+	// using InboundRules but can be used in conjunction with it
+	// +optional
+	InboundRuleRefs []*corev1.ObjectReference `json:"inboundRuleRefs,omitempty"`
```
cbang-akamai, line 46: We might want to have some sort of validation that limits the maximum number of rules (across inbound and outbound rulesets) to 25: https://techdocs.akamai.com/linode-api/reference/post-firewalls

AshleyDumaine: We do that check [here](https://github.com/linode/cluster-api-provider-linode/blob/main/internal/controller/linodefirewall_controller_helpers.go#L461) but we could add some extra pre-validation on a webhook

cbang-akamai: Sounds good! But it doesn't need to need to be a blocker here 🙅‍♀️

internal/controller/linodefirewall_controller.go
```diff
@@ -205,23 +207,47 @@ func (r *LinodeFirewallReconciler) reconcile(
...
 		if !apierrors.IsNotFound(err) {
-			logger.Error(err, "failed to fetch referenced AddressSet", "namespace", addrSetRef.Namespace, "name", addrSetRef.Name)
+			logger.Error(err, "failed to fetch referenced Object", "kind", objRef.Kind, "namespace", objRef.Namespace, "name", objRef.Name)
 			return err
 		}
 	}
 	if controllerutil.RemoveFinalizer(addrSet, getFinalizer(fwScope.LinodeFirewall)) {
 		if err := r.Update(ctx, addrSet); err != nil {
-			logger.Error(err, "failed to remove finalizer from AddressSet")
+			logger.Error(err, "failed to remove finalizer from Object", "kind", objRef.Kind, "namespace", objRef.Namespace, "name", objRef.Name)
```
cbang-akamai, line 226: Does this create a race condition when a user tries to delete multiple `LinodeFirewall`s that reference the same `AddressSet` simultaneously?

internal/controller/linodefirewall_controller.go
```diff
@@ -205,23 +207,47 @@ func (r *LinodeFirewallReconciler) reconcile(
...
+	}
+	if err := r.Client.Get(ctx, client.ObjectKey{Namespace: objRef.Namespace, Name: objRef.Name}, firewallRule); err != nil {
+		if !apierrors.IsNotFound(err) {
+			logger.Error(err, "failed to fetch referenced Object", "kind", objRef.Kind, "namespace", objRef.Namespace, "name", objRef.Name)
+			return err
+		}
+	}
+	if controllerutil.RemoveFinalizer(firewallRule, getFinalizer(fwScope.LinodeFirewall)) {
+		if err := r.Update(ctx, firewallRule); err != nil {
+			logger.Error(err, "failed to remove finalizer from Object", "kind", objRef.Kind, "namespace", objRef.Namespace, "name", objRef.Name)
```
cbang-akamai, line 250: Same as [here](TODO), but for `FirewallRule`s

internal/controller/linodefirewall_controller_helpers.go
```diff
@@ -232,52 +239,47 @@ func transformToCIDR(ip string) string {
...
-	ruleLabel := formatRuleLabel(lfw.Spec.OutboundPolicy, rule.Label)
+	ruleIPv4s, ruleIPv6s = filterDuplicates(ruleIPv4s, ruleIPv6s)
 
-	// Process IPv4
-	processIPv4Rules(ruleIPv4s, rule, ruleLabel, &createOpts.Rules.Outbound)
+	ruleLabel := formatRuleLabel(rule.Action, rule.Label)
 
-	// Process IPv6
-	processIPv6Rules(ruleIPv6s, rule, ruleLabel, &createOpts.Rules.Outbound)
+	if ruleType == "inbound" {
```
cbang-akamai, line 276:
```suggestion
	if ruleType == ruleTypeInbound {
```

internal/controller/linodefirewall_controller_helpers.go
```diff
@@ -232,52 +239,47 @@ func transformToCIDR(ip string) string {
...
+		processIPRules(ruleIPv4s, rule, ruleLabel, linodego.IPTypeIPv4, &createOpts.Rules.Inbound)
+		processIPRules(ruleIPv6s, rule, ruleLabel, linodego.IPTypeIPv6, &createOpts.Rules.Inbound)
+	} else if ruleType == "outbound" {
```
cbang-akamai, line 279:
```suggestion
	} else if ruleType == ruleTypeOutbound {
```

internal/controller/linodefirewall_controller_helpers.go
```diff
@@ -382,44 +384,56 @@ func processIPv4Rules(ips []string, rule infrav1alpha2.FirewallRule, ruleLabel s
...
 
-	// If no IPs, return early
-	if len(ips) == 0 {
-		return
+			return err
+		}
+		finalizer := getFinalizer(firewall)
+		if !controllerutil.ContainsFinalizer(rule, finalizer) {
+			controllerutil.AddFinalizer(rule, finalizer)
+			if err := k8sClient.Update(ctx, rule); err != nil {
```
cbang-akamai, line 427: Same as [here](TODO), but for adding the `LinodeFirewall` finalizer on the `FirewallRule` when creating multiple `LinodeFirewall`s referencing the same `FirewallRule`

config/crd/bases/infrastructure.cluster.x-k8s.io_firewallrules.yaml
```diff
@@ -0,0 +1,143 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.16.5
+  labels:
+    clusterctl.cluster.x-k8s.io/move-hierarchy: "true"
+  name: firewallrules.infrastructure.cluster.x-k8s.io
```
cbang-akamai, line 9: nit: Would calling this CRD `LinodeFirewallRule` be better since we're using the shared `infrastructure.cluster.x-k8s.io` group here and `github.com/linode/linodego`-specific types inside the definition?

AshleyDumaine: I didn't want to do that because I want Linode* resources to correspond to a cloud resource. For example, AddressSet does not.

internal/controller/linodefirewall_controller_helpers.go
AshleyDumaine: nice, I forgot these variables existed 😅

internal/controller/linodefirewall_controller.go
AshleyDumaine: Added the retry package for both adding and removing finalizers for firewallrules and addressets 👍

internal/controller/linodefirewall_controller_helpers.go
```diff
@@ -313,13 +314,26 @@ func processAddresses(addresses *infrav1alpha2.NetworkAddresses) (ipv4s, ipv6s [
...
 
+func addFinalizer(ctx context.Context, k8sClient clients.K8sClient, lfw *infrav1alpha2.LinodeFirewall, obj client.Object) error {
+	finalizer := getFinalizer(lfw)
+	if !controllerutil.ContainsFinalizer(obj, finalizer) {
+		controllerutil.AddFinalizer(obj, finalizer)
+		if retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			return k8sClient.Update(ctx, obj)
+		}); retryErr != nil {
+			return retryErr
+		}
```
cbang-akamai, line 330: I'm not sure if this retry logic solves the race condition from trying to manage the same object concurrently [[1]].

From the [GoDoc](https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict):

>  RetryOnConflict is used to make an update to a resource when you have to worry about conflicts caused by other code making unrelated updates to the resource at the same time. **fn should fetch the resource to be modified, make appropriate changes to it, try to update it, and return (unmodified) the error from the update function.**

I think you'll just be retrying the update using the same `client.Object`(with the same stale resource version [[2]]) here. Maybe you could also try a patch ([[3]], [[4]]) instead of retrying?

[1]: https://book.kubebuilder.io/reference/good-practices#why-is-it-recommended-to-avoid-a-scenario-where-multiple-controllers-are-updating-the-same-custom-resource-cr
[2]: https://alenkacz.medium.com/kubernetes-operators-best-practices-understanding-conflict-errors-d05353dff421
[3]: https://kubernetes.io/docs/reference/using-api/api-concepts/#api-verbs
[4]: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#Patch

cbang-akamai: slap

internal/controller/linodefirewall_controller_helpers.go
```diff
@@ -311,13 +314,53 @@ func processAddresses(addresses *infrav1alpha2.NetworkAddresses) (ipv4s, ipv6s [
...
 func getFinalizer(lfw *infrav1alpha2.LinodeFirewall) string {
 	return fmt.Sprintf("lfw.%s.%s", lfw.Namespace, lfw.Name)
 }
 
+func addAddrSetFinalizer(ctx context.Context, k8sClient clients.K8sClient, firewall *infrav1alpha2.LinodeFirewall, addrSetRef *corev1.ObjectReference) (*infrav1alpha2.AddressSet, error) {
+	addrSet := &infrav1alpha2.AddressSet{}
+	if addrSetRef.Namespace == "" {
+		addrSetRef.Namespace = firewall.Namespace
+	}
+	finalizer := getFinalizer(firewall)
```
cbang-akamai, line 327: If we delete a `FirewallRule` reference in the `LinodeFirewall`'s spec, does the `LinodeFirewall`'s finalizer get cleaned up?

AshleyDumaine: Short answer is no. Longer answer is there’s no way AFAICT to act on an outdated spec to fetch the resource and remove its finalizer on an update. There might be a way to check with an updateFunc predicate if refs were removed and remove the finalizer(s) but that’s going to have to ignore any errors it runs into because the UpdateFunc is just supposed to return a boolean if a reconcile needs to happen, it really shouldn’t be doing resource updating.
It might be preferable to remove the finalizers altogether on the dependent resources if there's no reasonable way to get the older resource version in the actual reconcile loop to remove the finalizer once it's de-referenced from the LinodeFirewall

internal/controller/linodefirewall_controller_helpers.go
```diff
@@ -382,44 +373,47 @@ func processIPv4Rules(ips []string, rule infrav1alpha2.FirewallRule, ruleLabel s
...
-	// Initialize rules if nil
-	if *rules == nil {
-		*rules = make([]linodego.FirewallRule, 0)
+func processFirewallRule(ctx context.Context, k8sClient clients.K8sClient, firewall *infrav1alpha2.LinodeFirewall, log logr.Logger, ruleRef *corev1.ObjectReference, ruleType string, createOpts *linodego.FirewallCreateOptions) error {
+	rule := &infrav1alpha2.FirewallRule{}
+	if ruleRef.Namespace == "" {
+		ruleRef.Namespace = firewall.Namespace
 	}
+	if err := k8sClient.Get(ctx, client.ObjectKey{Namespace: ruleRef.Namespace, Name: ruleRef.Name}, rule); err != nil {
+		log.Error(err, "failed to get AddressSet", "namespace", ruleRef.Namespace, "name", ruleRef.Name)
```
cbang-akamai, line 408:
```suggestion
		log.Error(err, "failed to get FirewallRule", "namespace", ruleRef.Namespace, "name", ruleRef.Name)
```

cbang-akamai (approved):

cbang-akamai (approved):
Put the finalizers back in 😤

cbang-akamai:
> > Put the finalizers back in 😤
> 
> 

image

PR #589, opened 2024-12-09
[improvement] Remove severity levels from controller logic

komer3:

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/controller/linodecluster_controller.go
+++ b/internal/controller/linodecluster_controller.go
@@ -252,7 +252,7 @@ func (r *LinodeClusterReconciler) reconcilePreflightLinodeFirewallCheck(ctx cont
 		if reconciler.HasStaleCondition(clusterScope.LinodeCluster,
 			ConditionPreflightLinodeNBFirewallReady,
 			reconciler.DefaultTimeout(r.ReconcileTimeout, reconciler.DefaultClusterControllerReconcileTimeout)) {
-			conditions.MarkFalse(clusterScope.LinodeCluster, ConditionPreflightLinodeNBFirewallReady, string(cerrs.CreateClusterError), "", err.Error())
+			conditions.MarkFalse(clusterScope.LinodeCluster, ConditionPreflightLinodeNBFirewallReady, string(cerrs.CreateClusterError), "", "%s", err.Error())
 			return ctrl.Result{}, err
 		}
 		return ctrl.Result{RequeueAfter: reconciler.DefaultClusterControllerReconcileDelay}, nil
@@ -284,7 +284,7 @@ func (r *LinodeClusterReconciler) reconcilePreflightLinodeVPCCheck(ctx context.C
 		if reconciler.HasStaleCondition(clusterScope.LinodeCluster,
 			ConditionPreflightLinodeVPCReady,
 			reconciler.DefaultTimeout(r.ReconcileTimeout, reconciler.DefaultClusterControllerReconcileTimeout)) {
-			conditions.MarkFalse(clusterScope.LinodeCluster, ConditionPreflightLinodeVPCReady, string(cerrs.CreateClusterError), "", err.Error())
+			conditions.MarkFalse(clusterScope.LinodeCluster, ConditionPreflightLinodeVPCReady, string(cerrs.CreateClusterError), "", "%s", err.Error())
 			return ctrl.Result{}, err
 		}
 		return ctrl.Result{RequeueAfter: reconciler.DefaultClusterControllerReconcileDelay}, nil
@@ -300,7 +300,7 @@ func setFailureReason(clusterScope *scope.ClusterScope, failureReason cerrs.Clus
 	clusterScope.LinodeCluster.Status.FailureReason = util.Pointer(failureReason)
 	clusterScope.LinodeCluster.Status.FailureMessage = util.Pointer(err.Error())
 
-	conditions.MarkFalse(clusterScope.LinodeCluster, clusterv1.ReadyCondition, string(failureReason), "", err.Error())
+	conditions.MarkFalse(clusterScope.LinodeCluster, clusterv1.ReadyCondition, string(failureReason), "", "%s", err.Error())
 
 	lcr.Recorder.Event(clusterScope.LinodeCluster, corev1.EventTypeWarning, string(failureReason), err.Error())
 }
--- a/internal/controller/linodecluster_controller.go
+++ b/internal/controller/linodecluster_controller.go
@@ -329,14 +329,14 @@ func (r *LinodeClusterReconciler) reconcileDelete(ctx context.Context, logger lo
 	switch {
 	case clusterScope.LinodeCluster.Spec.Network.LoadBalancerType == "external":
 		logger.Info("LoadBalacing managed externally, nothing to do.")
-		conditions.MarkFalse(clusterScope.LinodeCluster, clusterv1.ReadyCondition, clusterv1.DeletedReason, "", "Deletion in progress")
+		conditions.MarkFalse(clusterScope.LinodeCluster, clusterv1.ReadyCondition, clusterv1.DeletedReason, "", "%s", "Deletion in progress")
 		r.Recorder.Event(clusterScope.LinodeCluster, corev1.EventTypeWarning, "LoadBalacing managed externally", "LoadBalacing managed externally, nothing to do.")
 
 	case clusterScope.LinodeCluster.Spec.Network.LoadBalancerType == lbTypeDNS:
 		if err := removeMachineFromDNS(ctx, logger, clusterScope); err != nil {
 			return fmt.Errorf("remove machine from loadbalancer: %w", err)
 		}
-		conditions.MarkFalse(clusterScope.LinodeCluster, clusterv1.ReadyCondition, clusterv1.DeletedReason, "", "Load balancing for Type DNS deleted")
+		conditions.MarkFalse(clusterScope.LinodeCluster, clusterv1.ReadyCondition, clusterv1.DeletedReason, "", "%s", "Load balancing for Type DNS deleted")
 		r.Recorder.Event(clusterScope.LinodeCluster, corev1.EventTypeNormal, clusterv1.DeletedReason, "Load balancing for Type DNS deleted")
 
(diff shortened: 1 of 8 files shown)


internal/controller/linodemachine_controller_test.go
```diff
@@ -629,8 +630,12 @@ var _ = Describe("create", Label("machine", "create"), func() {
 
 			Expect(rutil.ConditionTrue(&linodeMachine, ConditionPreflightMetadataSupportConfigured)).To(BeTrue())
 			Expect(rutil.ConditionTrue(&linodeMachine, ConditionPreflightCreated)).To(BeFalse())
-			Expect(conditions.Get(&linodeMachine, ConditionPreflightCreated).Severity).To(Equal(clusterv1.ConditionSeverityWarning))
-			Expect(conditions.Get(&linodeMachine, ConditionPreflightCreated).Message).To(ContainSubstring("time is up"))
+			// Expect(conditions.Get(&linodeMachine, ConditionPreflightCreated).Message).To(ContainSubstring("time is up"))
```
AshleyDumaine, line 633: should we uncomment or remove this line?

komer3: Ah good catch! I thought I removed all the comments

internal/controller/linodevpc_controller.go
```diff
@@ -171,7 +171,8 @@ func (r *LinodeVPCReconciler) reconcile(
 		logger = logger.WithValues("vpcID", *vpcScope.LinodeVPC.Spec.VPCID)
 
 		err = r.reconcileUpdate(ctx, logger, vpcScope)
-		if err != nil && !reconciler.HasConditionSeverity(vpcScope.LinodeVPC, clusterv1.ReadyCondition, clusterv1.ConditionSeverityError) {
+		if err != nil && !reconciler.HasStaleCondition(vpcScope.LinodeVPC, clusterv1.ReadyCondition,
+			reconciler.DefaultTimeout(r.ReconcileTimeout, reconciler.DefaultVPCControllerReconcileTimeout)) {
```
rahulait, line 175: I think we should just use `reconciler.DefaultVPCControllerReconcileTimeout` and remove `reconciler.DefaultTimeout()` as I don't think it will be useful anytime. I don't think we ever want to set global default timeout and use it.

komer3: Do we want to do that in all of the controllers?

rahulait: Yeah, we can do this for all the controllers (places where you have changed in this PR) to keep things simple.

komer3: I just realized that removing reconciler.DefaultTimeout() will break a bunch of test cases. For some of cases, we are manually setting the r.ReconcileTimeout to 1 nano sec to trigger an immediate timeout and see how that gets handled. This is done is all of the controllers test cases. 

One example: https://github.com/linode/cluster-api-provider-linode/blob/7f99e91f3eae2fcf79455a9c567b094eeef7c253/internal/controller/linodefirewall_controller_test.go#L142

internal/controller/linodecluster_controller.go
```diff
@@ -154,7 +154,9 @@ func (r *LinodeClusterReconciler) reconcile(
 	// Handle deleted clusters
 	if !clusterScope.LinodeCluster.DeletionTimestamp.IsZero() {
 		if err := r.reconcileDelete(ctx, logger, clusterScope); err != nil {
-			if !reconciler.HasConditionSeverity(clusterScope.LinodeCluster, clusterv1.ReadyCondition, clusterv1.ConditionSeverityError) {
+			if !reconciler.HasStaleCondition(clusterScope.LinodeCluster,
+				clusterv1.ReadyCondition,
+				reconciler.DefaultTimeout(r.ReconcileTimeout, reconciler.DefaultClusterControllerReconcileTimeout)) {
 				logger.Info("re-queuing cluster/nb deletion")
 				return ctrl.Result{RequeueAfter: reconciler.DefaultClusterControllerReconcileDelay}, nil
```
rahulait, line 161: I wonder what our approach should be here. Currently we are doing:
```
if err != nil {
     hasStaleCondition() {
         # try delete every 5 seconds
         return {5 min}, nil
     }
     # try delete with exponential backoff
     return {...}, err
}
```
I wonder for delete, we should just have delete every 5 or 10 seconds and not have exponential backoff. With exponential backoffs, deletes could get stuck for 1000 secs (16+ mins) max if something goes wrong. Looking for thoughts from others.

komer3: I was thinking about this yesterday. Do we know how many times does it have to fail to get stuck for 1000+ sec?

internal/controller/linodevpc_controller.go
rahulait: Ah ok, that might be why its added in first place, hmm

internal/controller/linodecluster_controller.go
rahulait: This is what I have seen in past:
```
   0.004881143569946289
   0.022043943405151367
   0.04145383834838867
   0.08641910552978516
   0.1618349552154541
   0.3228580951690674
   0.6422019004821777
   1.2825469970703125
   2.561889171600342
   5.122460842132568
   10.242242097854614
   20.482059001922607
   40.962461948394775
   81.92205500602722
   0.0189058780670166
   163.84187817573547
   655.36194896698
   1000.0018429756165
   1000.0022790431976
   1000.0020320415497
   1000.0018348693848
```
So mostly failures 16-17 times. I see in most of the places we are doing exponential first and then serial when its a stale condition, I wonder if its a fixed pattern we want to use here as well or we want to have fixed for delete.

internal/controller/linodemachine_controller.go
```diff
@@ -330,11 +330,13 @@ func (r *LinodeMachineReconciler) reconcilePreflightVPC(ctx context.Context, log
 	}
 	if err := machineScope.Client.Get(ctx, client.ObjectKeyFromObject(&linodeVPC), &linodeVPC); err != nil {
 		logger.Error(err, "Failed to fetch LinodeVPC")
-		if reconciler.RecordDecayingCondition(machineScope.LinodeMachine,
-			ConditionPreflightLinodeVPCReady, string(cerrs.CreateClusterError), err.Error(),
+		if reconciler.HasStaleCondition(machineScope.LinodeMachine,
+			ConditionPreflightLinodeVPCReady,
 			reconciler.DefaultTimeout(r.ReconcileTimeout, reconciler.DefaultClusterControllerReconcileTimeout)) {
+			conditions.MarkFalse(machineScope.LinodeMachine, ConditionPreflightLinodeVPCReady, string(cerrs.CreateClusterError), "", "%s", err.Error())
```
rahulait, line 336: Will this lead to `lastTransitionTime` getting updated for the condition? If yes, then in next iteration `HasStaleCondition()` will return false as it only looks at lastTransitionTime now. Previously, if it was already stale (by checking error severity), we were marking it as stale even if lastTransitionTime changed. Not sure here how it will behave.

internal/controller/linodecluster_controller.go
komer3: I think it would be okay to do 10 or more seconds but I'm worried that a bad delete request could slow down the controllers from processing other requests Or could the controller handle that well?

internal/controller/linodemachine_controller.go
komer3: Ah good catch! Currently MarkFalse() will update the last transition time but I won't be using this func anymore in the follow up. I'll use Set() func which give more granular control over how to set conditions and I can just not set it then. 

But is this something to look out for in a lot of different places?

internal/controller/linodevpc_controller.go
```diff
@@ -298,7 +294,7 @@ func (r *LinodeVPCReconciler) reconcileDelete(ctx context.Context, logger logr.L
 		logger.Info("VPC ID is missing, nothing to do")
 	}
 
-	conditions.MarkFalse(vpcScope.LinodeVPC, clusterv1.ReadyCondition, clusterv1.DeletedReason, clusterv1.ConditionSeverityInfo, "VPC deleted")
+	conditions.MarkFalse(vpcScope.LinodeVPC, clusterv1.ReadyCondition, clusterv1.DeletedReason, "", "VPC deleted")
```
rahulait, line 297: Do we want to keep this consistent with other code? Like we are doing `conditions.MarkFalse(X, Y, Z, "", "%s", "Error msg or empty")`. I am ok with any approach. I just saw that you had `"%s", ""` at all places previously, hence was checking.

internal/controller/linodecluster_controller.go
rahulait: Since controllers pick from queue, a bad request would mean that request will keep on getting requeued every 10 secs if its using fixed requeue logic or exponentially if its using exponential logic. Since default concurrency is 10, I don't think it will slow down things much as its just for delete and we don't have too many deletes all the time. With exponential delete, I just see that someone might see their resources not getting cleaned up on UI fast if its stuck in exponential backoff. But that also is a rare case. I am ok with current approach, just wanted some extra eyes and thoughts here.

rahulait:
Other than couple of comments I have to test and confirm, the PR looks good to me.

internal/controller/linodemachine_controller.go
komer3: Yup I confirmed that `lastTransitionTime` is not changed or update if the condition stays the same. MarkFalse() uses Set() func which has this comment over it:

```
// NOTE: If a condition already exists, the LastTransitionTime is updated only if a change is detected
// in any of the following fields: Status, Reason, Severity and Message.
```

internal/controller/linodevpc_controller.go
komer3: Hmmm it works either way but for the sake of consistency I'll switch to using `"%s", ""`

internal/controller/linodecluster_controller.go
komer3: I'm fine with doing 10 sec delayed requeue. Its a rare case but probably will happen :)

AshleyDumaine (approved):

PR #614, opened 2025-01-09
[improvement] : allow CCM to work with different environments in CAPL

rahulait:
This PR allows CCM to work with different environments. Support to allow CAPL to work with different environments was added in https://github.com/linode/cluster-api-provider-linode/pull/546. This PR fixes CCM for child clusters. Along with other necessary vars required to provision a cluster, following vars also need to be set accordingly:
```
export LINODE_URL=<path of env's API>
export LINODE_CA=<env's CA file path on disk>
export LINODE_EXTERNAL_SUBNET=<network to be marked as public network>
export LINODE_CA_BASE64=<base64 encoded value of LINODE_CA>
```
Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/templates/addons/ccm-linode/ccm-linode.yaml
+++ b/templates/addons/ccm-linode/ccm-linode.yaml
@@ -9,7 +9,7 @@ spec:
   repoURL: https://linode.github.io/linode-cloud-controller-manager/
   chartName: ccm-linode
   namespace: kube-system
-  version: ${LINODE_CCM_VERSION:=v0.4.21}
+  version: ${LINODE_CCM_VERSION:=v0.4.22}
   options:
     waitForJobs: true
     wait: true
--- a/templates/addons/ccm-linode/ccm-linode.yaml
+++ b/templates/addons/ccm-linode/ccm-linode.yaml
@@ -23,3 +23,19 @@ spec:
       name: "linode-token-region"
     image:
       pullPolicy: IfNotPresent
+    env:
+      - name: LINODE_EXTERNAL_SUBNET
+        value: ${LINODE_EXTERNAL_SUBNET:=""}
+      - name: LINODE_URL
+        value: ${LINODE_URL:="https://api.linode.com"}
+      - name: SSL_CERT_DIR
+        value: "/tls"
+    volumeMounts:
+      - name: cacert
+        mountPath: /tls
+        readOnly: true
+    volumes:
+      - name: cacert
+        secret:
+          secretName: linode-ca
+          defaultMode: 420
--- a/templates/addons/cluster-resource-set/secret.yaml
+++ b/templates/addons/cluster-resource-set/secret.yaml
@@ -14,6 +14,21 @@ stringData:
       apiToken: ${LINODE_TOKEN}
       region: ${LINODE_REGION}
 ---
+apiVersion: v1
+kind: Secret
+type: addons.cluster.x-k8s.io/resource-set
+metadata:
+  name: linode-${CLUSTER_NAME}-crs-1
+stringData:
+  linode-ca.yaml: |-
+    kind: Secret
+    apiVersion: v1
+    metadata:
+      name: linode-ca
+      namespace: kube-system
+    data:
+      cacert.pem: ${LINODE_CA_BASE64:=""}
+---
 apiVersion: addons.cluster.x-k8s.io/v1beta1
 kind: ClusterResourceSet
 metadata:
@@ -25,5 +40,6 @@ spec:
   resources:
     - kind: Secret
       name: linode-${CLUSTER_NAME}-crs-0
+    - kind: Secret
+      name: linode-${CLUSTER_NAME}-crs-1
   strategy: Reconcile
----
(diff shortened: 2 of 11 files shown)


templates/addons/ccm-linode/ccm-linode.yaml
```diff
@@ -23,3 +23,19 @@ spec:
       name: "linode-token-region"
     image:
       pullPolicy: IfNotPresent
+    env:
+      - name: LINODE_EXTERNAL_SUBNET
+        value: ${LINODE_EXTERNAL_SUBNET:=""}
+      - name: LINODE_URL
+        value: ${LINODE_URL:="https://api.linode.com"}
```
komer3, line 30: qq: Do we need to append /v4 here or is that not needed?

rahulait: For a different environment, when testing, I had added /v4 and it kept on failing until I enabled debug mode for linodego. Say for `https://api.XXXX.YYY.ZZZ/v4` url, it was ListRegions() api call to `https://api.XXXX.YYY.ZZZ/v4/v4/regions` and that was failing. Once I removed `/v4` from URL, it started working fine for custom environments.
```
~~~ REQUEST ~~~
GET  /v4/v4/regions?page=1  HTTP/1.1
```
One can also test this using the following code:
```
package main

import (
	"context"
	"crypto/tls"
	"fmt"
	"log"
	"net/http"
	"os"
	"time"

	"github.com/linode/linodego"
)

func createLinodeClient() (linodego.Client, error) {
	timeout := 10 * time.Second

	// Use system cert pool if root CA cert was not provided explicitly for this client.
	// Works around linodego not using system certs if LINODE_CA is provided,
	// which affects all clients spawned via linodego.NewClient
	tlsConfig := &tls.Config{MinVersion: tls.VersionTLS12}
	linodeURL := os.Getenv("LINODE_URL")
	linodeCA := os.Getenv("LINODE_CA")

	httpClient := &http.Client{
		Timeout: timeout,
		Transport: &http.Transport{
			TLSClientConfig: tlsConfig,
		},
	}

	newClient := linodego.NewClient(httpClient)
	if linodeCA != "" {
		newClient.SetRootCertificate(linodeCA)
	}
	if linodeURL != "" {
		newClient.SetBaseURL(linodeURL)
	}

	return newClient, nil
}

func main() {
	_, ok := os.LookupEnv("LINODE_TOKEN")
	if !ok {
		log.Fatal("Could not find LINODE_TOKEN, please assert it is set.")
	}
	_, ok = os.LookupEnv("LINODE_URL")
	if !ok {
		log.Fatal("Could not find LINODE_URL, please assert it is set.")
	}
	_, ok = os.LookupEnv("LINODE_CA")
	if !ok {
		log.Fatal("Could not find LINODE_CA, please assert it is set.")
	}
	client, err := createLinodeClient()
	if err != nil {
		log.Fatal(err)
	}
	client.SetDebug(true)

	fmt.Printf("Getting list of regions\n")
	regions, err := client.ListRegions(context.TODO(), nil)
	if err != nil {
		log.Fatal(err)
	}
	for _, region := range regions {
		log.Println(region)
	}
}
```
Use `LINODE_TOKEN=YYY LINODE_CA="" LINODE_URL=https://api.linode.com/v4 go run test.go` to verify.

komer3: Oh cool. Good to know. I believe in CSI driver we had to pass /v4 at the end

komer3 (dismissed):

templates/addons/cluster-resource-set/secret.yaml
```diff
@@ -14,6 +14,21 @@ stringData:
       apiToken: ${LINODE_TOKEN}
       region: ${LINODE_REGION}
 ---
+apiVersion: v1
```
AshleyDumaine, line 17: is there a benefit of using an extra Secret instead of adding a new field in the other Secret's stringData?

rahulait: I was having issues setting the variable with cert content and then using it directly as `\n` were messing up things. So I created another secret with `data` instead of `stringdata` and provided the base64 encoded cert there. Not sure if there is a way I can pass cert to original secret without messing up things with formatting here.

AshleyDumaine: Sorry, specifically I meant doing:
```
apiVersion: v1
kind: Secret
type: addons.cluster.x-k8s.io/resource-set
metadata:
  name: linode-${CLUSTER_NAME}-crs-0
stringData:
  linode-token-region.yaml: |-
    kind: Secret
    apiVersion: v1
    metadata:
      name: linode-token-region
      namespace: kube-system
    stringData:
      apiToken: ${LINODE_TOKEN}
      region: ${LINODE_REGION}
  linode-ca.yaml: |-
    kind: Secret
    apiVersion: v1
    metadata:
      name: linode-ca
      namespace: kube-system
    data:
      cacert.pem: ${LINODE_CA_BASE64:=""}
```
So then we don't need a `linode-${CLUSTER_NAME}-crs-1`

rahulait: Good point, I missed it. Updated the PR with the change.

AshleyDumaine (approved):
LGTM

PR #636, opened 2025-01-21
[CI] : add workflow to test upgrade

rahulait:
This PR allows us to provision mgmt cluster using previous release and then run latest main on it to make sure nothing breaks
Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/.github/workflows/e2e-upgrade-test.yaml
+++ b/.github/workflows/e2e-upgrade-test.yaml
@@ -3,7 +3,7 @@ name: Run e2e upgrade test
 on:
   push:
     branches:
-      - main
+      - test-upgrade
     tags:
       - "*"
   workflow_dispatch:
--- a/.github/workflows/e2e-upgrade-test.yaml
+++ b/.github/workflows/e2e-upgrade-test.yaml
@@ -1,0 +1,98 @@
+name: Run e2e upgrade test
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "*"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pull-requests: read
+  actions: read
+
+concurrency:
+  group: e2e-upgrade-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      # Expose matched filters as job 'src' output variable
+      paths: ${{ steps.filter.outputs.changes }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Harden Runner
+        uses: step-security/harden-runner@v2
+        with:
+          disable-sudo: true
+          egress-policy: block
+          allowed-endpoints: >
+            api.github.com:443
+            github.com:443
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: .github/filters.yml
+  e2e-upgrade-test:
+    needs: changes
+    name: e2e-upgrade-test
+    if: ${{contains(fromJSON(needs.changes.outputs.paths), 'src')}}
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e-upgrade-test.yaml
```diff
@@ -0,0 +1,100 @@
+name: Run e2e upgrade test
+
+on:
+  push:
+    branches:
+      - test-upgrade
```
rahulait, line 6: This will be updated to main before merge

e2e/capl-cluster-flavors/kubeadm-capl-cluster/chainsaw-test.yaml
```diff
@@ -33,10 +35,12 @@ spec:
                 value: ($namespace)
               - name: CLUSTERCTL_CONFIG
                 value: (env('CLUSTERCTL_CONFIG'))
+              - name: KUBERNETES_VERSION
+                value: (env('KUBERNETES_VERSION') || 'v1.29.1')
```
AshleyDumaine, line 39: I'm wondering if we should set this to a more recent version like 1.31 or even 1.32 in the GHA workflow so it's easier to see what version we're testing with

rahulait: Yeah, thats what I want to do. Currently, the workflow first checks out immutable last tag (0.8.2 in this case) and runs it. Since we can't change it, we have to run v1.29.1 there. I am adding it so that in future when we cut a new tag, we can provide a newer version going forward here using env var.

Makefile
```diff
@@ -159,10 +159,43 @@ test: generate fmt vet envtest ## Run tests.
...
+	git checkout $(LAST_RELEASE)
+
+.PHONY: last-release-cluster
+last-release-cluster: kind ctlptl tilt kustomize clusterctl chainsaw kind-cluster checkout-last-release local-release local-deploy
+	GIT_REF=$(COMMON_CLUSTER_REF) LOCALBIN=$(CACHE_BIN) CLUSTERCTL_CONFIG=$(CLUSTERCTL_CONFIG) $(CHAINSAW) test --namespace $(COMMON_NAMESPACE) --assert-timeout 600s --skip-delete ./e2e/capl-cluster-flavors/kubeadm-capl-cluster
+
+.PHONY: test-upgrade
+test-upgrade: last-release-cluster checkout-latest-commit
+	$(MAKE) local-release
+	$(MAKE) local-deploy
```
komer3, line 192: nit: could we call these make target similar to how we did above?

rahulait: Previously, I had something like this:
```
last-release-cluster: kind ctlptl tilt kustomize clusterctl chainsaw kind-cluster checkout-last-release local-release local-deploy

test-upgrade: last-release-cluster checkout-latest-commit local-release local-deploy
```
Since local-release and local-deploy were also dependencies for last-release-cluster target, they would run there but won't run it again for test-upgrade target. Somehow Makefile was running them only once even though I marked them as .PHONY and thought it will re-run them. Here is a sample GHA run log when it tried to invoke next local-release but just skipped it.
https://github.com/linode/cluster-api-provider-linode/actions/runs/12896635374/job/35960077709#step:5:2217
In the end, I had to call them within the target itself.

komer3: Oh interesting. Thats good to know!

komer3 (approved):

AshleyDumaine (approved):

rahulait:
Whoops. Merged too early before reverting the branch name. Going to change in https://github.com/linode/cluster-api-provider-linode/pull/637

PR #637, opened 2025-01-22
[test] : fix max clustername length

rahulait:
We generate linodefirewall names using clustername. For ex: <clustername>-nb. Linode firewall has a limit of 3-32 chars for firewall label. Since we are adding `-nb` to a generated firewall label, we should make sure that total length should not exceed 32 chars. Hence, we are limiting clustername to 29 chars in e2e tests.
Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


komer3 (approved):

e2e/kind-config.yaml
komer3, line 1: We don't need this anymore?

rahulait: We are not using it anywhere it seems. We are using the one in .tilt dir.

komer3 (approved):

PR #640, opened 2025-01-23
[improvement] add support for custom CAPL CAs in kustomize

eljohnson92:
This PR moves the logic for using custom CAs with CAPL from tilt into kustomize so future releases can use custom CAs with the capi-operator
Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -116,6 +116,7 @@ capl_resources = [
     "linodeobjectstoragebuckets.infrastructure.cluster.x-k8s.io:customresourcedefinition",
     "linodeobjectstoragekeys.infrastructure.cluster.x-k8s.io:customresourcedefinition",
     "capl-mutating-webhook-configuration:mutatingwebhookconfiguration",
+    "capl-ca:secret",
     "capl-linodeclustertemplate-editor-role:clusterrole",
     "capl-linodeclustertemplate-viewer-role:clusterrole",
     "capl-linodemachine-editor-role:clusterrole",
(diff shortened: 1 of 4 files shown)


tchinmai7 (dismissed):

Tiltfile
```diff
@@ -176,8 +177,8 @@ for resource in manager_yaml:
         resource["stringData"]["dnsToken"] = os.getenv("LINODE_DNS_TOKEN")
         if os.getenv("LINODE_URL"):
             resource["stringData"]["LINODE_URL"] = os.getenv("LINODE_URL")
-        if os.getenv("LINODE_CA_BASE64"):
-            resource["stringData"]["SSL_CERT_DIR"] = "/tmp/linode-ca"
+    if resource["metadata"]["name"] == "capl-ca":
+        resource["data"]["cacert.pem"] = os.getenv("LINODE_CA_BASE64")
```
rahulait, line 181: Since its already base64 encoded, it should be stringData here. Adding full cert content as env var was causing issues, so it was base64 encoded and we added the var LINODE_CA_BASE64

config/default/capl-ca.yaml
```diff
@@ -0,0 +1,7 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ca
+data:
```
rahulait, line 6: should be stringData here.

config/default/capl-ca.yaml
```diff
@@ -0,0 +1,7 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ca
```
rahulait, line 5:
```suggestion
  name: capl-ca
```

rahulait: Nah, didn't helped in devcloud. Something else is off/wrong

config/manager/manager.yaml
```diff
@@ -102,5 +102,14 @@ spec:
             requests:
               cpu: 1000m
               memory: 512Mi
+          volumeMounts:
```
rahulait, line 105: Need SSL_CERT_DIR env var to be set in this file.

```
      containers:
        - command:
            - /manager
          ...
          env:
          ...
            - name: SSL_CERT_DIR
              value: /tls
              ...
```

rahulait: Tested this in custom environment, worked after this change.

eljohnson92: tested this as well and verified with this things are now working well in a custom environment

rahulait (approved):
LGTM

PR #641, opened 2025-01-23
[fix] fix cni plugins version - the release 1.6.1 is broken on github

tchinmai7:

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


rahulait (approved):
LGTM. Can you please add which functionality was broken so that we are aware of it in future? Looks like we are using pretty old containerd versions here.

tchinmai7:
> LGTM. Can you please add which functionality was broken so that we are aware of it in future? Looks like we are using pretty old containerd versions here.

the links to the container-plugin-1.6.1 was broken
see -  https://github.com/containernetworking/plugins/releases/tag/v1.6.2 
vs - https://github.com/containernetworking/plugins/releases/tag/v1.6.1 

specifically the linux-amd64 tarballs were removed in the 1.6.1 release

tchinmai7:
re: containerd 2.0 - its a pretty big change, so we should probably publish capl 0.9 if we want to move there. Has a different config file syntax etc.

We're on 1.7.24 while the latest is 1.7.25

rahulait:
Trying to understand if we should add some test case here to protect ourselves from such errors. Were you seeing specific errors when installing cluster because I installed default flavor and everything seems to come up fine and pods can communicate over the VPC interfaces.

tchinmai7:
> Trying to understand if we should add some test case here to protect ourselves from such errors. Were you seeing specific errors when installing cluster because I installed default kubeadm flavor and everything seems to come up fine and pods can communicate over the VPC interfaces.

do we set `CNI_PLUGIN_VERSIONS` anywhere? because the older default gives a 404.. 

```
❯ curl -L "https://github.com/containernetworking/plugins/releases/download/v1.6.1/cni-plugins-linux-amd64-v1.6.1.tgz"
```

rahulait:
Ah I see, we are using older commit hash in CAPL flavors https://github.com/linode/cluster-api-provider-linode/blob/main/templates/flavors/kubeadm/default/kubeadmConfigTemplate.yaml#L10

eljohnson92 (approved):

PR #657, opened 2025-02-03
[improvement] Add default machine count values for control plane and worker nodes

Andrews2024:
**Reason for change**:
This change adds default values of `1` for the `CONTROL_PLANE_MACHINE_COUNT` and `WORKER_MACHINE_COUNT` variables if they're not set in the user's environment variables. This prevents an issue where a cluster starts up with no machine deployments because the machine count environment variables aren't set and `replicas` is set to 0.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/e2e/capl-cluster-flavors/kubeadm-capl-cluster/chainsaw-test.yaml
+++ b/e2e/capl-cluster-flavors/kubeadm-capl-cluster/chainsaw-test.yaml
@@ -42,7 +42,6 @@ spec:
               clusterctl generate cluster $CLUSTER -n $NAMESPACE \
               --kubernetes-version ${KUBERNETES_VERSION} \
               --infrastructure local-linode:v0.0.0 \
-              --control-plane-machine-count 1 --worker-machine-count 1 \
               --config ${CLUSTERCTL_CONFIG:=${HOME}/.cluster-api/clusterctl.yaml} > default-cluster.yaml
             check:
               ($error == null): true
--- a/e2e/capl-cluster-flavors/kubeadm-flatcar-vpcless-capl-cluster/chainsaw-test.yaml
+++ b/e2e/capl-cluster-flavors/kubeadm-flatcar-vpcless-capl-cluster/chainsaw-test.yaml
@@ -79,7 +79,6 @@ spec:
               clusterctl generate cluster $CLUSTER -n $NAMESPACE \
               --kubernetes-version ${KUBERNETES_VERSION} \
               --infrastructure local-linode:v0.0.0 \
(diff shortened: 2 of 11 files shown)


cbang-akamai:
> I'm wondering it's a bug in clusterctl that the default for the worker node count is reporting as 1 on our main branch if it's hard-coded in clusterctl to be 0 by default:
> 
> ```
> clusterctl generate cluster test-cluster --kubernetes-version v1.29.1 --infrastructure local-linode:v0.0.0 --list-variables | grep COUNT
>   - CONTROL_PLANE_MACHINE_COUNT  (defaults to 1)
>   - WORKER_MACHINE_COUNT         (defaults to 1)
> ```
>
> `clusterctl generate cluster test-cluster --kubernetes-version v1.29.1 --infrastructure local-linode:v0.0.0 --list-variables | grep COUNT`

It looks like the `--list-variables` output somehow pulls the value from your environment variables first (🤷‍♀️): https://github.com/kubernetes-sigs/cluster-api/blob/850efe73eb6257d850c6e23e09f8e43b5f53dde9/cmd/clusterctl/cmd/util.go#L85-L106

For example:
```sh
$ unset CONTROL_PLANE_MACHINE_COUNT
$ unset WORKER_MACHINE_COUNT

$ clusterctl generate cluster test-cluster --kubernetes-version v1.29.1 --infrastructure local-linode:v0.0.0 --list-variables | grep COUNT
  - CONTROL_PLANE_MACHINE_COUNT  (defaults to 1)
  - WORKER_MACHINE_COUNT         (defaults to 0)
```

Andrews2024:
The default for `WORKER_MACHINE_COUNT` seems to be very intentionally 0 in the clusterctl source code, given that it's set to 0 in multiple places. Do we add something to here to set the environment variables for the user if they haven't been set? The end result of the worker machine count being defaulted to 0 is that cluster deployments don't work out of the box unless it's set to at least 1.

AshleyDumaine:
> The default for `WORKER_MACHINE_COUNT` seems to be very intentionally 0 in the clusterctl source code, given that it's set to 0 in multiple places. Do we add something to here to set the environment variables for the user if they haven't been set? The end result of the worker machine count being defaulted to 0 is that cluster deployments don't work out of the box unless it's set to at least 1.

I think we just need to document that the worker and control plane count can be set optionally since the default is a 1-node control plane only cluster otherwise: https://linode.github.io/cluster-api-provider-linode/topics/getting-started.html#setting-up-your-cluster-environment-variables

PR #677, opened 2025-02-25
[improvement] Select subnet for cluster

Andrews2024:
Adding the option to select which subnet to use for deploying nodes rather than using the first subnet by default. This change also modifies the templates to make the VPC and cluster IP CIDRs configurable.

**TODOs**:

- [ ] squashed commits
- [x] adds unit tests


templates/addons/ccm-linode/ccm-linode.yaml
```diff
@@ -17,7 +17,7 @@ spec:
   valuesTemplate: |
     routeController:
       vpcNames: {{ .InfraCluster.spec.vpcRef.name }}
-      clusterCIDR: 10.0.0.0/8
+      clusterCIDR: ${K8S_SERVICE_CIDR:=10.0.0.0/8}
```
rahulait, line 20: `10.0.0.0/8` is the entire VPC subnet. Lets rename this var to `VPC_NETWORK_CIDR`

Andrews2024: Updated here (https://github.com/linode/cluster-api-provider-linode/pull/677/files#diff-d1b9866c86b420a8086df2411916d47a8f86c34a7bf172a1f0ce3f5f75da0790R22) and here (https://github.com/linode/cluster-api-provider-linode/pull/677/files#diff-b3331ca1db3fef2d10e149f36d0c9db5900a11be5f8274a7d5130650d7799d9eR20)

templates/infra/cluster.yaml
AshleyDumaine, line 1: I think you'll also want this on templates/flavors/clusterclass-kubeadm/cluster-template.yaml, the others (dual-stack and full-vpcless) seem fine though.

templates/addons/cilium/cilium.yaml
AshleyDumaine, line 1: we'll also need this on templates/flavors/k3s/default/k3sControlPlane.yaml

AshleyDumaine:
Can we also update the VPC docs (docs/src/topics/vpc.md) to note the new env variables? 🙏

templates/infra/linodeCluster.yaml
```diff
@@ -15,3 +15,5 @@ spec:
     apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
     kind: LinodeFirewall
     name: ${CLUSTER_NAME}-nb
+  network:
+    subnetName: ${SUBNET_NAME:=default}
```
eljohnson92, line 19: nit: missing newline at the end of the file here

api/v1alpha2/linodecluster_types.go
```diff
@@ -174,6 +174,9 @@ type NetworkSpec struct {
 	// additionalPorts contains list of ports to be configured with NodeBalancer.
 	// +optional
 	AdditionalPorts []LinodeNBPortConfig `json:"additionalPorts,omitempty"`
+	// subnetName is the name/ label of the VPC subnet to be used by the cluster
```
komer3, line 177: is there a reason to have `/` after name in the comment here?

Andrews2024: We use name and label somewhat interchangeably. The actual subnet struct uses label, but in other places we use subnet name. This is just to say that for this field, they're the same.

rahulait (approved):
LGTM

PR #678, opened 2025-02-27
[feat] Nodebalancer VPC Integration

komer3:
With this PR, we are now able to configure Nodebalancer Backend using VPC. 

To try this feature, you need to add subnetRange (ex: 10.0.10.0/30) to LinodeCluster.Spec.Network. That will create the nodebalancer in the VPC (the one that LinodeVPC creates) and configure the nodes to use VPC IPs instead of private IPs for NB configuration. 

To test, generate default kubeadm manifest and edit the LinodeCluster to add the subnetRange `10.10.0.0/30`. Then simply apply the manifest. 

 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


api/v1alpha2/linodecluster_types.go
```diff
@@ -177,11 +177,14 @@ type NetworkSpec struct {
...
 	SubnetName string `json:"subnetName,omitempty"`
-
 	// UseVlan provisions a cluster that uses VLANs instead of VPCs. IPAM is managed internally.
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
 	// +optional
 	UseVlan bool `json:"useVlan,omitempty"`
+	// subnetRange is the range of we want to provide for creating nodebalancer in VPC.
+	// example: 10.10.10.0/30
+	// +optional
+	SubnetRange string `json:"subnetRange,omitempty"`
```
rahulait, line 187: I am thinking we should call this variable `NodeBalancerBackendIPv4Range` instead of `SubnetRange` as its specifically for NodeBalancer.

cloud/services/loadbalancers.go
```diff
@@ -70,6 +86,48 @@ func EnsureNodeBalancer(ctx context.Context, clusterScope *scope.ClusterScope, l
...
+	}
+	if len(linodeVPC.Spec.Subnets) == 0 {
+		err = errors.New("No subnets found in LinodeVPC")
+		logger.Error(err, "Failed to fetch LinodeVPC")
+		return -1, err
+	}
+
+	// We are using the first subnet as the default one
+	// get the subnetID from the linodeVPC object
+	subnetID := linodeVPC.Spec.Subnets[0].SubnetID
```
rahulait, line 121: If `clusterScope.LinodeCluster.Spec.Network.SubnetName` is specified, we should use its id as subnetID. A single VPC can have multiple clusters which means there can be more than one subnets and we might want to use different subnet than the first subnet.

api/v1alpha2/linodecluster_types.go
komer3: Hmm how about `NodeBalancerIPv4Range`? Similar but not as long

rahulait: I was thinking of keeping it consistent with what we have in CCM, but this also looks fine to me

komer3: Oh yeah we should keep it consistent. I'll just use what you suggested

cloud/services/loadbalancers.go
```diff
@@ -44,6 +44,22 @@ func EnsureNodeBalancer(ctx context.Context, clusterScope *scope.ClusterScope, l
 		Tags:   []string{string(clusterScope.LinodeCluster.UID)},
 	}
 
+	// if subnetRange is set, create the NodeBalancer in the specified VPC
+	if clusterScope.LinodeCluster.Spec.Network.NodeBalancerBackendIPv4Range != "" && clusterScope.LinodeCluster.Spec.VPCRef != nil {
+		logger.Info("Creating NodeBalancer in VPC", "subnetRange", clusterScope.LinodeCluster.Spec.Network.NodeBalancerBackendIPv4Range)
```
rahulait, line 49: nit: should we change `subnetRange` to `nodeBalancerBackendIPv4Range` here?

komer3: Thanks for catching that!

cloud/services/loadbalancers.go
```diff
@@ -70,6 +86,62 @@ func EnsureNodeBalancer(ctx context.Context, clusterScope *scope.ClusterScope, l
 	return clusterScope.LinodeClient.CreateNodeBalancer(ctx, createConfig)
 }
 
+// getSubnetID returns the subnetID of the first subnet in the LinodeVPC.
+// If no subnets or subnetID is found, it returns an error.
```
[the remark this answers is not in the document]
komer3: Fixed that in the latest commit!

cloud/services/loadbalancers.go
```diff
@@ -70,6 +86,62 @@ func EnsureNodeBalancer(ctx context.Context, clusterScope *scope.ClusterScope, l
...
+		if subnetID == 0 {
+			return -1, fmt.Errorf("subnet with label %s not found in VPC", subnetName)
+		}
+	} else {
+		subnetID = linodeVPC.Spec.Subnets[0].SubnetID
+	}
+
+	// Validate the selected subnet ID
+	if subnetID == 0 {
+		return -1, errors.New("selected subnet ID is 0")
```
rahulait, line 139: Should we return `0, error` in this method where we are returning -1? We generally have a check like `if subnetID == 0` or `if subnetID != 0` and setting -1 would satisfy `if subnetID != 0` which we don't want.

cloud/services/loadbalancers.go
```diff
@@ -169,60 +241,99 @@ func EnsureNodeBalancerConfigs(
...
+	useVPCIps := clusterScope.LinodeCluster.Spec.Network.NodeBalancerBackendIPv4Range != "" && clusterScope.LinodeCluster.Spec.VPCRef != nil
+	if useVPCIps {
+		// Get subnetID
+		subnetID, err := getSubnetID(ctx, clusterScope, logger)
+		if err != nil {
+			logger.Error(err, "Failed to fetch Linode Subnet ID")
+			return err
+		}
+		for _, IPs := range linodeMachine.Status.Addresses {
+			// Look for internal IPs that are NOT 192.168.* (likely VPC IPs)
```
rahulait, line 314: We do allow 192.168.0.0/17 subnets to be created within VPC. We don't allow 192.168.128.0/17 https://techdocs.akamai.com/cloud-computing/docs/manage-vpc-subnets
Here, we should check that internal ip is not in that 192.168.128.0/17 subnet instead of just 192.168.*

komer3: Addressed!

cloud/services/loadbalancers.go
komer3: Addressed!

cloud/services/loadbalancers.go
komer3: Addressed!

rahulait (dismissed):
LGTM

rahulait (approved):

rahulait (approved):

PR #699, opened 2025-03-11
refactor pause logic

scmeggitt:
**What this PR does / why we need it**: This PR refactors the pause logic to use either the pause annotation or check the capi cluster resource associated with the resource. To do this, I have added logic that adds the cluster-name label to resources reference by the linode cluster (vpc) or linode machine (placement group, firewall). If you want to pause an individual resource, you can add the annotation; however, if we want the cluster.spec.paused value to trickle down, we need to associate resources with cluster. This is needed for clusterctl move which will remove all annotations on a resource during the delete process, meaning resources depending only on that pause annotation will be deleted.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


internal/controller/linodecluster_controller.go
```diff
@@ -115,57 +114,42 @@ func (r *LinodeClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-	return r.reconcile(ctx, clusterScope, logger)
-}
-
-func (r *LinodeClusterReconciler) reconcilePause(ctx context.Context, clusterScope *scope.ClusterScope, logger logr.Logger) error {
-	// Pausing a cluster pauses the VPC as well.
-	// First thing to do is handle a paused Cluster. Paused clusters shouldn't be deleted.
-	isPaused, conditionChanged, err := paused.EnsurePausedCondition(ctx, clusterScope.Client, clusterScope.Cluster, clusterScope.LinodeCluster)
-	if err != nil {
-		return err
+	if _, _, err := paused.EnsurePausedCondition(ctx, clusterScope.Client, clusterScope.Cluster, clusterScope.LinodeCluster); err != nil {
```
tchinmai7, line 117: does it make sense to do `err != nil || isPaused` ? effectively replaces the next condition?

scmeggitt: yeah that is fair

internal/controller/linodecluster_controller.go
```diff
@@ -115,57 +114,42 @@ func (r *LinodeClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-	vpcPatchHelper, err := patch.NewHelper(&linodeVPC, clusterScope.Client)
-	if err != nil {
-		return fmt.Errorf("failed to build patch helper for linode VPC object: %w", err)
+	if _, ok := labels[clusterv1.ClusterNameLabel]; ok {
+		return nil
 	}
-	return vpcPatchHelper.Patch(ctx, &linodeVPC)
+	labels[clusterv1.ClusterNameLabel] = clusterScope.LinodeCluster.Name
+	linodeVPC.SetLabels(labels)
+	return clusterScope.Client.Update(ctx, &linodeVPC)
```
tchinmai7, line 152: should we wrap this in a retry.OnConflict() something?

internal/controller/linodecluster_controller.go
```diff
@@ -201,7 +185,8 @@ func (r *LinodeClusterReconciler) reconcile(
 		return res, err
 	}
 
-	if err := r.reconcilePause(ctx, clusterScope, logger); err != nil {
+	err := r.reconcileVPCLabel(ctx, clusterScope, logger)
+	if err != nil {
```
tchinmai7, line 189: nit: 
```suggestion
	if err := r.reconcileVPCLabel(ctx, clusterScope, logger); err != nil {
```

internal/controller/linodecluster_controller_test.go
```diff
@@ -329,8 +332,7 @@ var _ = Describe("cluster-lifecycle", Ordered, Label("cluster", "cluster-lifecyc
 					clusterKey := client.ObjectKeyFromObject(&linodeCluster)
 					Expect(k8sClient.Get(ctx, clusterKey, &linodeCluster)).To(Succeed())
 					Expect(linodeCluster.Status.Ready).To(BeTrue())
-					Expect(linodeCluster.Status.Conditions).To(HaveLen(4))
-					Expect(conditions.Get(&linodeCluster, clusterv1.PausedV1Beta2Condition).Status).To(Equal(metav1.ConditionFalse))
+					Expect(linodeCluster.Status.Conditions).To(HaveLen(3))
```
tchinmai7, line 335: are we removing paused conditions from the LinodeCluster?

util/reconciler/annotations.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/annotations"
+)
+
+func IsPaused(cluster *clusterv1.Cluster, o metav1.Object) bool {
+	if cluster != nil {
+		return annotations.IsPaused(cluster, o)
```
tchinmai7, line 11: shouldn't we look at `cluster.spec.paused`?

internal/controller/linodecluster_controller.go
```diff
@@ -115,57 +114,42 @@ func (r *LinodeClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 	}
-	linodeVPC.SetAnnotations(annotations)
-	vpcPatchHelper, err := patch.NewHelper(&linodeVPC, clusterScope.Client)
-	if err != nil {
-		return fmt.Errorf("failed to build patch helper for linode VPC object: %w", err)
+	if _, ok := labels[clusterv1.ClusterNameLabel]; ok {
+		return nil
 	}
-	return vpcPatchHelper.Patch(ctx, &linodeVPC)
+	labels[clusterv1.ClusterNameLabel] = clusterScope.LinodeCluster.Name
```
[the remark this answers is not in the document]
scmeggitt: there are some tests that don't have a capi cluster in the scope, which is why i did that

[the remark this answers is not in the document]
scmeggitt: also above there is a block of labels:
```
labels := map[string]string{
		clusterv1.ClusterNameLabel:         clusterScope.LinodeCluster.Name,
		clusterv1.MachineControlPlaneLabel: "",
	}
```

util/reconciler/annotations.go
scmeggitt: that is what annotation.IsPaused does: https://github.com/kubernetes-sigs/cluster-api/blob/e48409b8466e1c6f2ea72260025eba696cd4ae11/util/annotations/helpers.go#L29

scmeggitt: just happens to be in the annotation package upstream

internal/controller/linodecluster_controller.go
scmeggitt: makes sense, updated

internal/controller/linodecluster_controller.go
scmeggitt: fixed

internal/controller/linodecluster_controller.go
scmeggitt: yeah that's fair, updated it

internal/controller/linodecluster_controller.go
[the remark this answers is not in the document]
tchinmai7: in situations where the capi-cluster's name differs from a linodecluster, this would be problematic.. we should add the capi-cluster to the scope in the tests then

internal/controller/linodecluster_controller.go
```diff
@@ -115,57 +115,48 @@ func (r *LinodeClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		if isPaused {
+			logger.Info("linodeCluster or linked cluster is marked as paused, won't reconcile.")
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
 	}
-	if !(isPaused || conditionChanged) {
-		return nil
+
+	if reconciler.IsPaused(cluster, linodeCluster) {
```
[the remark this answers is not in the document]
scmeggitt: not quite because that depends on conditions and then nothing is checking if someone added the cluster.x-k8s.io/paused to the linodeCluster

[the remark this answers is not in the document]
scmeggitt: The capi contracts wants you to check both:
```
Providers SHOULD implement the pause behaviour for every object with a reconciliation loop. This is done by checking if spec.paused is set on the Cluster object and by checking for the cluster.x-k8s.io/paused annotation on the InfraCluster object.
```

internal/controller/linodecluster_controller.go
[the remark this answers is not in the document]
scmeggitt: do you want me to update that label above too?

internal/controller/linodecluster_controller.go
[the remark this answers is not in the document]
scmeggitt: oh ensure supports that so that is fine

internal/controller/linodecluster_controller.go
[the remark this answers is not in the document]
scmeggitt: i updated this to add cluster for all resources and checking if cluster != nil before using name for label

internal/controller/linodecluster_controller.go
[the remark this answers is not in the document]
scmeggitt: removed the other function

tchinmai7 (approved):
changes lgtm

tchinmai7 (dismissed):

tchinmai7 (approved):

PR #708, opened 2025-03-17
[feat] Adding ability to BYO VPC and configure with it

komer3:
This PR give users the ability to provide VPC IDs in LinodeCluster (propagates the vpc to linodemachine) or set it per machine level in LinodeMachine resource. 

To use the feature, you can set the VPC ID in the following ways:
- LinodeCluster: set `spec.vpcID`. This will make sure all the nodes under this LinodeCluster will have a vpc interfaces configured to it.
- LinodeMachine: set `spec.vpcID`. This allows you to set the vpc per machine level

fyi, make sure to update CCM helm chart configuration to add/update `routecontroller.vpcName` for the VPC you are using. This is important to configure the network routes.

For firewall, this feature is already available. You can set the nodebalancer firewall in LinodeCluster under `spec.network.nodeBalancerFirewallID`. Under `spec.firewallID` for LinodeMachine.

 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


internal/controller/linodecluster_controller.go
```diff
@@ -303,6 +304,37 @@ func (r *LinodeClusterReconciler) reconcilePreflightLinodeFirewallCheck(ctx cont
...
+		if len(vpc.Subnets) == 0 {
+			err := fmt.Errorf("VPC with ID %d has no subnets", vpcID)
+			logger.Error(err, "Failed preflight check: VPC has no subnets")
+			conditions.Set(clusterScope.LinodeCluster, metav1.Condition{
+				Type:    ConditionPreflightLinodeVPCReady,
+				Status:  metav1.ConditionFalse,
+				Reason:  util.CreateError,
+				Message: err.Error(),
+			})
+			return ctrl.Result{RequeueAfter: reconciler.DefaultClusterControllerReconcileDelay}, err
```
rahulait, line 331: If err is not nil, it will always use exponential backoff. Here, we can have:
1. `return ctrl.Result{}, err` for exponential backoff
2.  `return ctrl.Result{RequeueAfter: reconciler.DefaultClusterControllerReconcileDelay}, nil` for fixed time backoff

rahulait: Hitting linode API every 10-20 seconds on failure might be better than exponential which will hit it 10+ times in a second before it falls to more longer sleep durations. Both approaches have merits/demerits but I think fixed might be better for most of the times for us.

internal/controller/linodecluster_controller.go
```diff
@@ -303,6 +304,37 @@ func (r *LinodeClusterReconciler) reconcilePreflightLinodeFirewallCheck(ctx cont
...
+		vpc, err := clusterScope.LinodeClient.GetVPC(ctx, vpcID)
+		if err != nil {
+			logger.Error(err, "Failed to get VPC with provided ID", "vpcID", vpcID)
+			conditions.Set(clusterScope.LinodeCluster, metav1.Condition{
+				Type:    ConditionPreflightLinodeVPCReady,
+				Status:  metav1.ConditionFalse,
+				Reason:  util.CreateError,
+				Message: fmt.Sprintf("VPC with ID %d not found: %v", vpcID, err),
+			})
+			return ctrl.Result{RequeueAfter: reconciler.DefaultClusterControllerReconcileDelay}, err
```
rahulait, line 319: Same here, lets use fixed backoff

internal/controller/linodecluster_controller.go
```diff
@@ -221,23 +221,22 @@ func (r *LinodeClusterReconciler) reconcile(
...
+				// The condition is already set in reconcilePreflightLinodeVPCCheck, so we don't need to set it again
 				return res, err
 			}
+
+			// Only set to true if there was no error
+			conditions.Set(clusterScope.LinodeCluster, metav1.Condition{
+				Type:   ConditionPreflightLinodeVPCReady,
+				Status: metav1.ConditionTrue,
+				Reason: "LinodeVPCReady", // We have to set the reason to not fail object patching
+			})
```
rahulait, line 238: Can we set this in the end in reconcilePreflightLinodeVPCCheck?

internal/controller/linodemachine_controller.go
```diff
@@ -342,7 +342,64 @@ func (r *LinodeMachineReconciler) reconcileCreate(
...
+	// Check if machine has direct VPCID
+	if machineScope.LinodeMachine.Spec.VPCID != nil {
+		if err := r.validateVPC(ctx, *machineScope.LinodeMachine.Spec.VPCID, machineScope, logger, "Machine"); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// Check if cluster has direct VPCID
+	if machineScope.LinodeCluster.Spec.VPCID != nil {
```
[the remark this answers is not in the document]
komer3: I think that comes down to preference tbh. But i think `if -> else if` might be more readable and easier to understand. I'll switch that!

rahulait (approved):
LGTM 🚀

PR #709, opened 2025-03-19
Refactor reconciliation logic in Linode controllers to handle deletion cases gracefully

komer3:
Updated the Reconcile methods in LinodeFirewall, LinodePlacementGroup, and LinodeVPC controllers to allow for successful reconciliation when the associated cluster is not found during deletion. Added logging to indicate when a cluster is not found but the resource is being deleted, ensuring that the deletion process continues smoothly. Additionally, modified the pause logic to only check for paused conditions if the resource is not being deleted or if the cluster still exists.

 

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


AshleyDumaine (approved):
LGTM

PR #724, opened 2025-04-01
[deprecation] Deprecate stackscript

Andrews2024:
This PR keeps the `CloudinitMetadataSupport` field in the API for backwards compatibility but otherwise removes support for stackscript. Cloud-init is sufficiently supported so that stackscript can be removed.

**TODOs**:

- [x] squashed commits
- [X] includes documentation
- [X] updates e2e tests


rahulait (approved):
LGTM

PR #741, opened 2025-05-01
[improvement] Change LinodePlacementGroup Deletion Logic to requeue when nodes are attached

komer3:
This PR modifies the deletion process for `LinodePlacementGroup` resources.

Instead of attempting to forcefully unassign attached Linodes during deletion, the controller now checks if the Placement Group has any members. If it does, the deletion is requeued with a delay, waiting for the members to be detached externally (waiting for `LinodeMachine` controller to delete the nodes). A timeout (`DefaultPGControllerWaitForHasNodesTimeout`) is introduced; if Linodes remain attached after this period, the deletion fails gracefully with an error condition and event.

This change makes the deletion process safer by preventing the controller from disrupting potentially active Linodes. It ensures deletion only proceeds once the Placement Group is confirmed empty by the Linode API or the wait timeout is exceeded. The PR also improves API error handling during deletion and updates related tests and constants.

_Note: We now follow similar logic from `LinodeVPC` controller_

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/internal/controller/linodeplacementgroup_controller.go
+++ b/internal/controller/linodeplacementgroup_controller.go
@@ -24,7 +24,6 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	"github.com/linode/linodego"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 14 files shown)


internal/controller/linodeplacementgroup_controller.go
```diff
@@ -259,47 +258,67 @@ func (r *LinodePlacementGroupReconciler) reconcileDelete(ctx context.Context, lo
...
 
 		if pg != nil {
 			if len(pg.Members) > 0 {
-				logger.Info("Placement Group still has node(s) attached, unassigning them")
-				members := make([]int, 0, len(pg.Members))
-				for _, member := range pg.Members {
-					members = append(members, member.LinodeID)
+				logger.Info("Placement Group still has node(s) attached", "count", len(pg.Members))
+				waitTimeout := reconciler.DefaultTimeout(r.ReconcileTimeout, reconciler.DefaultPGControllerWaitForHasNodesTimeout)
+				if pgScope.LinodePlacementGroup.ObjectMeta.DeletionTimestamp.Add(waitTimeout).After(time.Now()) {
```
tchinmai7, line 285: should we modify the deletion timestamp? I feel like the delay introduced by `requeueAfter` should be enough

komer3: Mainly have this check for _final_ timeout (if pg still has members/nodes after waiting for 20+ mins). If 20+ mins have passed, then we want want to report error and set an event for visibility since this would be unexpected. 

If we keep requeuing, then if something goes wrong, when would we provide the user the feedback through an error eventually. 

Unless we want to return err along with requeue delay. But that can flood the logs with errors.

Thoughts?

tchinmai7 (approved):

PR #777, opened 2025-06-23
[ci]: remove unnecessary dependent resource deletion

AshleyDumaine:
**What this PR does / why we need it**: Now that we use ownerRefs for LFWs and LVPCs we don't need to delete them when the associated cluster is deleted.

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


Makefile
```diff
@@ -304,8 +304,6 @@ clean-release: clean-release-git
 .PHONY: clean-child-clusters
 clean-child-clusters: kubectl
 	$(KUBECTL) delete clusters -A --all --timeout=180s
-	$(KUBECTL) delete linodevpc -A --all --timeout=60s
```
amold1, line 307: been out of the loop but is this because the vpcs and firewalls now get cleaned up on cluster deletion?

rahulait (approved):

Makefile
AshleyDumaine: yes: https://github.com/linode/cluster-api-provider-linode/pull/746

amold1 (approved):

PR #783, opened 2025-06-30
[refactor] Allow manual tagging

rammanoj:
**What this PR does / why we need it:**

- Users should be able to tag managed Linodes through API/CLI/UI
- Fix bug that tags are given through .Spec.Tags during creation cannot be removed.

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [X] adds unit tests


rammanoj:
Close in favour of #790

PR #790, opened 2025-07-05
[feat, breaking] add LMT controller to support tag propagation, remove tag propagation via annotations

rammanoj:

- Add new controller to LMT which allows propagating configuration to LinodeMachine without recycle.
- Use LMT `spec.tags` and `status.tags` to propagate tags to LinodeMachine.
- Users should be able to tag managed Linodes through API/CLI/UI

**TODOs**:

- [ ] squashed commits
- [X] includes documentation
- [X] adds unit tests


internal/controller/linodemachine_controller_helpers.go
```diff
@@ -999,50 +993,33 @@ func configureFirewall(ctx context.Context, machineScope *scope.MachineScope, cr
...
-
-	if _, ok := machineScope.LinodeMachine.Annotations[machineTagsAnnotation]; ok {
-		var machineTags []string
-		if err := json.Unmarshal([]byte(machineScope.LinodeMachine.Annotations[machineTagsAnnotation]), &machineTags); err != nil {
-			return nil, err
-		}
-		for _, tag := range machineTags {
-			machineTagSet[tag] = struct{}{}
-		}
+func addArrayToSet(arr []string, strSet map[string]struct{}) {
```
AshleyDumaine, line 996: Using a variadic function is a bit cleaner IMO:
```func constructSet(arrs ...[]string) map[string]struct{} {
	strSet := make(map[string]struct{})
	for _, arr := range arrs {
		for _, elem := range arr {
			strSet[elem] = struct{}{}
		}
	}
	return strSet
}
```

internal/controller/linodemachine_controller_helpers.go
```diff
@@ -999,50 +993,33 @@ func configureFirewall(ctx context.Context, machineScope *scope.MachineScope, cr
...
+func getTags(machineScope *scope.MachineScope, instanceTags []string) []string {
+	machineTagSet := make(map[string]struct{})
+	addArrayToSet(instanceTags, machineTagSet)
+	addArrayToSet(machineScope.LinodeMachine.Spec.Tags, machineTagSet)
+	addArrayToSet(util.GetAutoGenTags(machineScope.LinodeCluster), machineTagSet)
+
+	desiredMachineTags := make(map[string]struct{})
+	addArrayToSet(machineScope.LinodeMachine.Spec.Tags, desiredMachineTags)
+
+	for _, tag := range machineScope.LinodeMachine.Status.Tags {
```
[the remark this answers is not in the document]
rammanoj: No. It is trying to delete the tags that a user deleted via NodePool.

internal/controller/linodemachinetemplate_controller.go
```diff
@@ -141,9 +142,13 @@ func (lmtr *LinodeMachineTemplateReconciler) reconcile(ctx context.Context, lmtS
...
 
-	lmtScope.LinodeMachineTemplate.Status.Tags = slices.Clone(lmtScope.LinodeMachineTemplate.Spec.Template.Spec.Tags)
-
-	lmtr.Logger.Info("Successfully reconciled LinodeMachineTemplate", "name", lmtScope.LinodeMachineTemplate.Name)
+	// update the LMT status.tags if all the linodeMachines spec.tags is successfully updated.
+	if outErr == nil {
+		lmtScope.LinodeMachineTemplate.Status.Tags = slices.Clone(lmtScope.LinodeMachineTemplate.Spec.Template.Spec.Tags)
+		lmtr.Logger.Info("Successfully reconciled LinodeMachineTemplate", "name", lmtScope.LinodeMachineTemplate.Name)
+	} else {
+		lmtr.Logger.Error(outErr, "Error in reconciling LinodeMachineTemplate, retrying..", "name", lmtScope.LinodeMachineTemplate.Name)
```
AshleyDumaine, line 150: Retry won't happen unless returning something like `ctrl.Result{RequeueAfter: reconciler.DefaultLMTControllerReconcilerDelay}, nil`. I'm not sure there are any cases where we'd want to retry given we're just listing and patching machines.

AshleyDumaine:
Changes look good. One last thing: can you add a chainsaw e2e test for LMT to make sure setting tags on that correctly creates a LM with the expected tag and status? You should be able to test it locally with `make e2etest` since GHA e2e won't run unless you set a `LINODE_TOKEN` in your fork's GHA secrets.

rammanoj:
> Changes look good. One last thing: can you add a chainsaw e2e test for LMT to make sure setting tags on that correctly creates a LM with the expected tag and status? You should be able to test it locally with `make e2etest` since GHA e2e won't run unless you set a `LINODE_TOKEN` in your fork's GHA secrets.

I have tested the specific test locally. The tests went well. PFA image. I also added `LINODE_TOKEN` as a secret on my fork. Can you trigger the jobs? 


image

PR #799, opened 2025-07-10
feat: Propagate LabelPrefix through LMT

rammanoj:

The PR introduces a way to propagate LabelPrefix through LMT. These LabelPrefixes are then added as prefixes to Linode instance labels, adhering to [these rules](https://github.com/linode/cluster-api-provider-linode/pull/799/files#diff-2b48475b2e2a99733fafb372bac1130eb4d69785dc8ee14f1de74638c07f6327R1046).

Fixes #

**TODOs**:

- [ ] squashed commits
- [X] includes documentation
- [X] adds unit tests

### DIFF
--- a/.github/workflows/build_test_ci.yml
+++ b/.github/workflows/build_test_ci.yml
@@ -56,12 +56,11 @@ jobs:
             golang.org:443
             proxy.golang.org:443
             sum.golang.org:443
-            objects.githubusercontent.com:443
+            *.githubusercontent.com:443
             storage.googleapis.com:443
             cli.codecov.io:443
             api.codecov.io:443
             ingest.codecov.io:443
-            raw.githubusercontent.com:443
 
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


internal/controller/linodemachine_controller.go
```diff
@@ -747,7 +746,7 @@ func (r *LinodeMachineReconciler) reconcileUpdate(ctx context.Context, logger lo
 
 	// update the tags if needed
 	machineTags := getTags(machineScope, linodeInstance.Tags)
-	if !slices.Equal(machineTags, linodeInstance.Tags) {
```
eljohnson92, line 750: Is there a reason slices.Equal won't work in this case?

tchinmai7: slices.Equal only works if the order is the same - https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/slices/slices.go;l=20
we'd either have to sort both these lists and then do `slices.Equal` or take this approach

rammanoj: `machineTags` is created from `linodeInstance.Tags`. `linodeInstance.Tags` is initially converted to a map. After performing a few operations, it gets converted back to slice. Practically, the order is not changed but theoretically the map is not an orderedMap. So, there is no guarantee for the order.

PR #807, opened 2025-07-21
[fix] Make ID the source of truth for vpc and subnet names

cbbond:

We currently filter the Linode API's ListVPCs response by both VPC ID and label when deciding whether to create a new VPC or update an existing one. This PR removes the latter filter and only checks for existing VPCs that match the ID found on the LinodeVPC resource. VPC ID is immutable and should be a sufficient filter for finding VPCs. Without this fix, any change to the VPC's label outside of CAPL will result in CAPL trying to create a new VPC when the underlying resource still exists.

Similarly, we currently build a map of subnets by label based on a list of VPC subnets we get from the Linode API. This method is vulnerable to the same issue in which any change to a subnet label after creation will prevent CAPL from reconciling the subnet further. This PR fixes this issue by building two maps that are keyed by both label and subnet ID. The code defaults to ID where possible but falls back to the label when the subnet ID is not yet populated (i.e. the first time we are reconciling a new subnet).  

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: N/A

- Existing behavior is preserved for subnets, we either adopt an existing subnet or create a new one based on whether the subnet is in the API response map
- The Linode API is the new source of truth for subnet labels, i.e. if a subnet label is updated via CloudManager then CAPL will adopt that label and update the LinodeVPC spec
- Added a test case where both the VPC and subnet labels change

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/controller/linodevpc_controller_helpers.go
+++ b/internal/controller/linodevpc_controller_helpers.go
@@ -77,23 +77,35 @@ func reconcileVPC(ctx context.Context, vpcScope *scope.VPCScope, logger logr.Log
 func reconcileExistingVPC(ctx context.Context, vpcScope *scope.VPCScope, vpc *linodego.VPC) error {
 	setVPCFields(&vpcScope.LinodeVPC.Spec, vpc)
 
-	// build a map of existing subnets to easily check for existence
-	existingSubnets := make(map[string]int, len(vpc.Subnets))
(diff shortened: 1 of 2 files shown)


komer3 (approved):
LGTM!

PR #809, opened 2025-07-23
[improvement] : allow specifying ipv6 slaac public ranges

rahulait:

This PR implements support for IPv6 SLAAC (Stateless Address Autoconfiguration) configuration for Linode machines by adding two new optional fields to LinodeMachineSpec: `EnableSLAAC` and `IsPublicIPv6`. When `EnableSLAAC` is enabled, instances can automatically configure IPv6 addresses, and `IsPublicIPv6` determines whether these addresses are publicly routable.

- Adds `EnableSLAAC` and `IsPublicIPv6` fields to LinodeMachineSpec API
- Refactors IPv6 configuration logic to support both SLAAC and traditional range-based configuration
- Updates interface creation logic to handle SLAAC IPv6 configuration alongside existing range-based approach

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/controller/linodemachine_controller_helpers.go
+++ b/internal/controller/linodemachine_controller_helpers.go
@@ -489,7 +489,7 @@ func getVPCInterfaceConfig(ctx context.Context, machineScope *scope.MachineScope
 		if netInterface.Purpose == linodego.InterfacePurposeVPC {
 			interfaces[i].SubnetID = &subnetID
 			// If IPv6 range config is not empty, add it to the interface configuration
-			if ipv6Config != nil {
+			if !isIPv6ConfigEmpty(ipv6Config) {
 				interfaces[i].IPv6 = ipv6Config
 			}
 			return nil, nil //nolint:nilnil // it is important we don't return an interface if a VPC interface already exists
@@ -506,7 +506,7 @@ func getVPCInterfaceConfig(ctx context.Context, machineScope *scope.MachineScope
 	}
 
 	// If IPv6 config is not empty, add it to the interface configuration
-	if ipv6Config != nil {
+	if !isIPv6ConfigEmpty(ipv6Config) {
 		vpcIntfCreateOpts.IPv6 = ipv6Config
 	}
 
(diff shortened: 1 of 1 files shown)


api/v1alpha2/linodemachine_types.go
```diff
@@ -112,6 +112,18 @@ type LinodeMachineSpec struct {
 	// +optional
 	VPCID *int `json:"vpcID,omitempty"`
 
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// EnableSLAAC is an option to enable SLAAC (Stateless Address Autoconfiguration) for the instance.
+	// This is useful for IPv6 addresses, allowing the instance to automatically configure its own IPv6 address.
+	// Defaults to false.
+	// +optional
+	EnableSLAAC *bool `json:"enableSLAAC,omitempty"`
```
rosskirkpat, line 120: What about creating a new `Ipv6Options` type with `EnableSLAAC` and `IsPublic` fields and embedding the new type instead? I think that would be a cleaner approach

rahulait: yeah, makes sense. I'll add that implementation

api/v1alpha2/linodemachine_types.go
```diff
@@ -121,6 +127,31 @@ type LinodeMachineSpec struct {
...
+	// +optional
+	EnableRanges *bool `json:"enableRanges,omitempty"`
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// IsPublicIPv6 is an option to enable public IPv6 for the instance.
+	// If set to true, the instance will have a publicly routable IPv6 range.
+	// Defaults to false.
+	// +optional
+	IsPublicIPv6 *bool `json:"isPublicIPv6,omitempty"`
+}
```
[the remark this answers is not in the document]
rahulait: I would prefer using a bool instead of pointer but I saw discussions where using pointers was considered better with k8s as treating nil is same as default/false value. I've shared that chat with you.

rosskirkpat (approved):

PR #813, opened 2025-07-24
[feat] make firewallID mutable, propagate it from lmt->lm

tchinmai7:
* Make the FirewallID mutable on the LinodeMachineTemplate
* Propagate the change from `LinodeMachineTemplate` -> `LinodeMachine`
* Update the firewalls attached to a linode when the spec changes

Fixes #

**TODOs**:

- [ ] squashed commits
- [x] includes documentation
- [x] adds unit tests


internal/controller/linodemachine_controller.go
```diff
@@ -771,6 +771,30 @@ func (r *LinodeMachineReconciler) reconcileUpdate(ctx context.Context, logger lo
...
+		return ctrl.Result{RequeueAfter: reconciler.DefaultMachineControllerWaitForRunningDelay}, nil
+	}
+
+	attachedFirewalls := make([]int, 0, len(firewalls))
+	for _, fw := range firewalls {
+		attachedFirewalls = append(attachedFirewalls, fw.ID)
+	}
+
+	// update the firewallID if needed.
+	if !slices.Contains(attachedFirewalls, machineScope.LinodeMachine.Spec.FirewallID) {
```
tchinmai7, line 787: We're replacing an array with an individual ID - is this OK?

internal/controller/linodemachine_controller.go
```diff
@@ -782,6 +787,41 @@ func (r *LinodeMachineReconciler) reconcileUpdate(ctx context.Context, logger lo
...
 
+func (r *LinodeMachineReconciler) reconcileFirewallID(ctx context.Context, logger logr.Logger, machineScope *scope.MachineScope, instanceID int) (ctrl.Result, error) {
+	// get the instance's firewalls
+	firewalls, err := machineScope.LinodeClient.ListInstanceFirewalls(ctx, instanceID, nil)
+	if err != nil {
+		logger.Error(err, "Failed to list firewalls for Linode instance")
+		return ctrl.Result{RequeueAfter: reconciler.DefaultMachineControllerWaitForRunningDelay}, nil
+	}
+
+	attachedFirewalls := make([]int, 0, len(firewalls))
```
AshleyDumaine, line 798: nit: can we rename this to something like `attachedFWIDs`?

internal/controller/linodemachine_controller.go
```diff
@@ -782,6 +787,41 @@ func (r *LinodeMachineReconciler) reconcileUpdate(ctx context.Context, logger lo
...
+	for _, fw := range firewalls {
+		attachedFirewalls = append(attachedFirewalls, fw.ID)
+	}
+
+	var desiredFirewalls []int
```
AshleyDumaine, line 803: same here: `desiredFWIDs`

internal/controller/linodemachine_controller.go
```diff
@@ -782,6 +787,41 @@ func (r *LinodeMachineReconciler) reconcileUpdate(ctx context.Context, logger lo
...
+	if machineScope.LinodeMachine.Spec.FirewallID != 0 {
+		desiredFirewalls = []int{machineScope.LinodeMachine.Spec.FirewallID}
```
AshleyDumaine, line 805: Out of scope for this PR, but I wonder if we want to support specifying multiple FW IDs if the API supports an instance being attached to multiple FWs.

tchinmai7: the api-supports attaching multiple firewalls, but only one can be active - imo that will make the CAPL api more complicated if we do it

internal/controller/linodemachine_controller.go
tchinmai7: done

AshleyDumaine (dismissed):
LGTM

rosskirkpat (dismissed):

AshleyDumaine (approved):

PR #816, opened 2025-07-28
[improvement] : Update dualstack flavors

rahulait:
This PR now uses linode specific ipv6 allocation and uses CCM to do nodeipam and assign both ipv4 and ipv6 ranges to nodes. We no longer need to have separate nodeipam-ccm flavor and it is removed in this PR.
Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


templates/flavors/k3s/dual-stack/kustomization.yaml
```diff
@@ -65,18 +80,29 @@ patches:
             namespace: kube-system
           spec:
             targetNamespace: kube-system
-            version: ${CILIUM_VERSION:=1.15.4}
+            version: ${CILIUM_VERSION:=1.16.10}
```
AshleyDumaine, line 83: Is there a reason for 1.16.10 instead of 1.16.12 for latest patch release on 1.16?

rahulait: Nope, the cilium [addon](https://github.com/linode/cluster-api-provider-linode/blob/main/templates/addons/cilium/cilium.yaml#L12) is at 1.16.10 and hence tried to keep it same.

AshleyDumaine (approved):
[verified](https://kubernetes.io/docs/tasks/network/validate-dual-stack/) latest dual-stack manifest works 👍

AshleyDumaine (approved):

PR #841, opened 2025-08-21
[improvement] enable the `optionalorrequired` linter for the various APIs

eljohnson92:

Fixes #

**TODOs**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/api/v1alpha2/addressset_types.go
+++ b/api/v1alpha2/addressset_types.go
@@ -23,8 +23,10 @@ import (
 // AddressSetSpec defines the desired state of AddressSet
 type AddressSetSpec struct {
 	// ipv4 defines a list of IPv4 address strings
+	// +optional
 	IPv4 *[]string `json:"ipv4,omitempty"`
 	// ipv6 defines a list of IPv6 address strings
+	// +optional
 	IPv6 *[]string `json:"ipv6,omitempty"`
 }
 
@@ -41,10 +43,13 @@ type AddressSetStatus struct {
 type AddressSet struct {
 	metav1.TypeMeta `json:",inline"`
 	// metadata is the standard object's metadata.
+	// +optional
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 	// spec is the desired state of the AddressSet
+	// +optional
 	Spec AddressSetSpec `json:"spec,omitempty"`
 	// status is the observed state of the AddressSet
+	// +optional
 	Status AddressSetStatus `json:"status,omitempty"`
 }
 
--- a/api/v1alpha2/firewallrule_types.go
+++ b/api/v1alpha2/firewallrule_types.go
@@ -25,20 +25,33 @@ import (
 // FirewallRuleSpec defines the desired state of FirewallRule.
 type FirewallRuleSpec struct {
 	// action is the action to take when the rule matches.
+	// +required
 	Action string `json:"action"`
+
 	// label is the label of the rule.
+	// +required
 	Label string `json:"label"`
+
 	// description is the description of the rule.
+	// +optional
 	Description string `json:"description,omitempty"`
+
 	// ports is the ports to apply the rule to.
+	// +optional
 	Ports string `json:"ports,omitempty"`
+
 	// protocol is the protocol to apply the rule to.
 	// +kubebuilder:validation:Enum=TCP;UDP;ICMP;IPENCAP
+	// +required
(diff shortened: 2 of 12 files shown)


api/v1alpha2/linodecluster_types.go
```diff
@@ -102,11 +103,17 @@ type LinodeClusterStatus struct {
...
 type LinodeCluster struct {
 	metav1.TypeMeta `json:",inline"`
+
 	// metadata is the standard object's metadata.
+	// +optional
 	metav1.ObjectMeta `json:"metadata,omitempty"`
+
 	// spec is the desired state of the LinodeCluster.
+	// +optional
 	Spec LinodeClusterSpec `json:"spec,omitempty"`
```
komer3, line 113: Do we want this to be optional?

eljohnson92: I am on board with making this change, it is a change in behaviour as it is currently implicitly optional, but in reality I don't think it is possible to pass in a linode* resource without a spec field

api/v1alpha2/linodeobjectstoragekey_types.go
```diff
@@ -71,10 +77,12 @@ type LinodeObjectStorageKeySpec struct {
 
 	// keyGeneration may be modified to trigger a rotation of the access key.
 	// +kubebuilder:default=0
+	// +required
 	KeyGeneration int `json:"keyGeneration"`
```
komer3, line 81: should this be required? My understands of how this is used might be murky but we only change this to rotate the key secret right?

api/v1alpha2/firewallrule_types.go
```diff
@@ -66,11 +82,17 @@ type FirewallRuleStatus struct {
...
 type FirewallRule struct {
 	metav1.TypeMeta `json:",inline"`
+
 	// metadata is the standard object's metadata.
+	// +optional
 	metav1.ObjectMeta `json:"metadata,omitempty"`
+
 	// spec is the desired state of the FirewallRule.
+	// +optional
 	Spec FirewallRuleSpec `json:"spec,omitempty"`
```
komer3, line 92: Should this also be required?

eljohnson92: good catch, this is now fixed

api/v1alpha2/linodeobjectstoragekey_types.go
eljohnson92: the tags added(outside of `spec` now) are all added to preserve what is generated in the CRD. I believe this is a "soft" required because the default comes into play before the required check so it is optional but set to 0 by default so always set anyways, I am open to changing that here and relaxing the openAPI validation but we can also do that in a separate PR if we want to try and avoid CRD changes outside of `spec`

komer3: gotcha. No this is fine. I was just looking for clarification

api/v1alpha2/addressset_types.go
```diff
@@ -41,10 +43,13 @@ type AddressSetStatus struct {
 type AddressSet struct {
 	metav1.TypeMeta `json:",inline"`
 	// metadata is the standard object's metadata.
+	// +optional
```
AshleyDumaine, line 46: metadata is optional? I would think `Name` or `GenerateName` is required in metadata

api/v1alpha2/firewallrule_types.go
```diff
@@ -80,6 +102,7 @@ type FirewallRule struct {
 type FirewallRuleList struct {
 	metav1.TypeMeta `json:",inline"`
 	// metadata is the standard object's metadata.
+	// +optional
```
AshleyDumaine, line 105: Missing optional below for `Items`? But also, wouldn't `Items` for all the `*List` structs be `required` since we don't have `omitempty`?

api/v1alpha2/addressset_types.go
```diff
@@ -41,10 +43,13 @@ type AddressSetStatus struct {
...
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 	// spec is the desired state of the AddressSet
+	// +required
```
AshleyDumaine, line 49: We probably only want `required` if `omitempty` isn't present.

komer3 (approved):

PR #842, opened 2025-08-22
[test][fix] E2E test improvements, add missing client for the LOBJ webhook

AshleyDumaine:
**What this PR does / why we need it**: Sometimes resources get orphaned by the chainsaw tests and because the names are tied to the git ref, retriggering a run because of a transient issue like a 503 on a dependency will cause the test to fail until the orphaned resource is cleaned up. This PR uses the kyverno chainsaw random function to fix that so tests can be retried regardless of the orphaned resource.
This also gets rid of the min-obj bucket tests in favor of force deletion since we want test buckets to clean up.

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


AshleyDumaine:
FWIW e2e tests are broken on main: https://github.com/linode/cluster-api-provider-linode/actions/runs/17217877581

This at least mitigates the issue where LOBJ resources couldn't even be applied because of the missing client:
```
Error from server: error when creating "test-cluster-2.yaml": admission webhook "validation.linodeobjectstoragebucket.infrastructure.cluster.x-k8s.io" denied the request: panic: runtime error: invalid memory address or nil pointer dereference [recovered]
Error from server: error when creating "test-cluster-2.yaml": admission webhook "validation.linodeobjectstoragebucket.infrastructure.cluster.x-k8s.io" denied the request: panic: runtime error: invalid memory address or nil pointer dereference [recovered]
```

eljohnson92 (dismissed):

eljohnson92 (approved):

PR #1052, opened 2026-03-28
fix: update TracedClient to use `wrappedruntimeclient.DefaultDecorator`

ChihweiLHBird:
Not sure if there was a historic reason, but this seems to be a typo to me. It wraps a client so it should use `wrappedruntimeclient`, right?


AshleyDumaine (approved):
Good catch, thanks!

PR #1063, opened 2026-04-15
[fix] restore pause handling during pivot

komer3:
## Summary
This fixes a regression in CAPL pause handling during cluster pivoting.

## What changed
- reintroduced `GetConditions` / `SetConditions` on the CAPL infrastructure types that participate in pause handling
- switched the affected controllers to use upstream `paused.EnsurePausedCondition(...)`
- updated the controller envtests to verify that owner-cluster pause sets the `Paused` condition on the corresponding Linode objects and skips reconciliation/finalizer setup
- updated the `LinodeFirewall`, `LinodeVPC`, and `LinodePlacementGroup` watch predicates to allow annotation-only updates, so direct `cluster.x-k8s.io/paused` toggles are not filtered out by generation-only event handling
- added a new Chainsaw E2E under `e2e/cluster-pause/` that creates a workload cluster, toggles `Cluster.spec.paused` on the owner cluster, and verifies pause/unpause propagation across CAPL resources
- wired the new E2E with a dedicated `cluster-pause` selector and `all` coverage so it runs on `main` without adding weight to PR `quick`

## Why
During cluster pivot, Cluster API pauses reconciliation while resources are being moved between management clusters. CAPL had drifted away from the upstream pause-condition contract, so some controllers were checking for a paused condition that was never being set. That leaves a gap where a controller can continue reconciling during pivot-driven delete/recreate flows.

Using upstream `paused.EnsurePausedCondition(...)` also means CAPL should honor the object-level `cluster.x-k8s.io/paused` annotation path. For resources still using `GenerationChangedPredicate{}`, annotation-only pause/unpause updates would be dropped before reconcile. The predicate updates close that gap so annotation-based pausing works end-to-end for the affected controllers.

## Impact
This brings CAPL back in line with the upstream pause utilities so CAPL resources honor cluster pause state consistently. The main user-facing impact is safer cluster pivoting: infrastructure objects should stop reconciling while the pivot is in progress instead of continuing into delete paths.

The predicate change also makes direct annotation-based pausing reliable for the affected controllers, instead of depending on a later spec change to trigger reconciliation.

The new E2E adds management-cluster coverage for the pause contract itself by verifying that CAPL resources follow `Cluster.spec.paused=true/false` in a real cluster lifecycle.

## Validation
- `make test`
- `make e2etest E2E_SELECTOR=cluster-pause E2E_FLAGS='--assert-timeout 20m0s'`


moshevayner (approved):
LGTM!
@komer3 my only concern is the codecov check failing - should we worry about it now? Or is it better to handle that after this is merged?

PR #1073, opened 2026-05-05
ci: allow devbox bootstrap egress

komer3:
## Summary
- allow the Jetify and Nix bootstrap endpoints needed by `devbox` in CI workflows
- allow `index.docker.io` for the docker build workflow under harden-runner `block` mode
- keep e2e on harden-runner `block` by pre-creating an IPv4-only shared Docker `kind` network before `ctlptl`/kind runs
- allow `*.linodeobjects.com:443` for object-storage e2e coverage under harden-runner `block` mode

## Problem
The original failure was `e2e-test (all) / all-e2e-tests` stopping in `Install devbox`, which pointed to missing bootstrap egress rather than a cache issue.

After fixing the devbox bootstrap allowlist, e2e still failed before the suite started while Docker/kind created the management cluster:
`failed to advertise addresses: write ip ::1->ff02::1: sendmsg: operation not permitted`

Our goal here was to keep harden-runner on `egress-policy: block`, not relax it to `audit`.

## Approach
- add the small set of bootstrap endpoints needed for `devbox`
- avoid the remaining Docker/kind startup failure by pre-creating the shared Docker `kind` network as IPv4-only in the e2e workflows
- keep the fix workflow-local and minimal rather than broadening the policy
- after the kind startup issue was fixed, add `*.linodeobjects.com:443` for the object-storage e2e traffic that showed up in the next block-mode run

## Validation
- `Install devbox` passed under `block` after the allowlist update: https://github.com/linode/cluster-api-provider-linode/actions/runs/25401633761
- `quick` e2e under `block` reached and ran the suite after the IPv4-only `kind` network change: https://github.com/linode/cluster-api-provider-linode/actions/runs/25455607906
- broader validation also passed with `e2e-selector=all` under `block`: https://github.com/linode/cluster-api-provider-linode/actions/runs/25457168014


komer3:
Left on a comment on what seems like a similar type of multi cast issue on the harden runner repo. We can track any fixes for our use case and remove the work around in future. 

Harden Runner issue: https://github.com/step-security/harden-runner/issues/228#issuecomment-4391918829

moshevayner (dismissed):

moshevayner (approved):

PR #1075, opened 2026-05-15
[feat]: implement maintenance watcher for linodeCluster

tchinmai7:
Automatically detects linodes that are scheduled for maintenance, and adds a condition on the machines. Maintenance can be disruptive, depending on the workload running on the clusters, and operators may want to know ahead of time, and surge replace nodes. This allows operators to configure a `MachineHealthCheck` to detect when machines are schedueld for maintenance and replace them.

Fixes #

**TODOs**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests


internal/controller/linodecluster_controller.go
```diff
@@ -218,9 +222,82 @@ func (r *LinodeClusterReconciler) reconcile(
...
+	}
+	maintenances, err := clusterScope.LinodeClient.ListMaintenances(ctx, &linodego.ListOptions{Filter: string(filter)})
+	if err != nil {
+		logger.Error(err, "Failed to fetch maintenance information from Linode API")
+		return nil, err
+	}
+
+	maintenanceLabels := make(map[string]struct{}, len(maintenances))
+	for _, maint := range maintenances {
+		if maint.Entity.Type != "linode" {
```
komer3, line 290: I see that entity is a pointer to a struct in linodego. Would linode API ever return nil for entity? If yes, we might want to check that here?

internal/controller/linodecluster_controller.go
```diff
@@ -218,9 +222,82 @@ func (r *LinodeClusterReconciler) reconcile(
...
+			continue
+		}
+		maintenanceLabels[maint.Entity.Label] = struct{}{}
```
komer3: Might be better to use linode IDs for comparision rather than label. Labels can be pretty fragile and could cause us issues in future

internal/controller/linodecluster_controller.go
```diff
@@ -218,9 +222,82 @@ func (r *LinodeClusterReconciler) reconcile(
...
+	}
+
+	var machinesForMaintenance []infrav1alpha2.LinodeMachine
+	for _, lm := range clusterScope.LinodeMachines.Items {
+		if _, ok := maintenanceLabels[lm.Name]; ok {
+			logger.Info("Found maintenance information for", "LinodeMachine", lm.Name)
+			machinesForMaintenance = append(machinesForMaintenance, lm)
+		}
```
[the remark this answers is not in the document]
komer3: good catch here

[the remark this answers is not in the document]
tchinmai7: fixing

internal/controller/linodecluster_controller.go
```diff
@@ -69,7 +73,8 @@ type LinodeClusterReconciler struct {
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=linodeclusters,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=linodeclusters/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=linodeclusters/finalizers,verbs=update
-
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machines,verbs=get;watch;list
```
AshleyDumaine: Do we need to add patch here too?

docs/src/topics/health-checking.md
```diff
@@ -23,3 +23,96 @@ on the infrastructure provider.
...
+
+## Replacing Machines Scheduled for Maintenance
+
+CAPL detects upcoming Linode infrastructure maintenance windows and sets a `MaintenanceScheduled` condition on
+the corresponding CAPI `Machine` objects. This condition can be used as a trigger for `MachineHealthCheck` to
+automatically replace machines before their maintenance window begins.
+
+### How it works
+
+During each `LinodeCluster` reconciliation, CAPL queries the Linode API for maintenance events scheduled within
```
AshleyDumaine, line 35: Wouldn't we need a new controller for this that periodically requeues since we're not regularly requeuing with this one once the cluster is fully provisioned?

AshleyDumaine: Nevermind, looks like things should be periodically resynced every 10 hours via the underlying controller-runtime cache refresh: https://github.com/kubernetes-sigs/controller-runtime/blob/v0.24.1/pkg/cache/cache.go#L45

internal/controller/linodecluster_controller.go
```diff
@@ -218,9 +222,82 @@ func (r *LinodeClusterReconciler) reconcile(
 		return retryIfTransient(err, logger)
 	}
 
+	if err := r.setMaintenanceConditions(ctx, clusterScope, logger); err != nil {
+		return retryIfTransient(err, logger)
+	}
```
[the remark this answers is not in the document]
tchinmai7: controller-runtime runs with a default requeue of 10Hrs - https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/cache/cache.go#L45

internal/controller/linodecluster_controller.go
```diff
@@ -218,9 +222,82 @@ func (r *LinodeClusterReconciler) reconcile(
...
+			continue
+		}
+		conditions.Set(capiMachine, metav1.Condition{
+			Type:               ConditionMaintenanceScheduled,
+			Status:             metav1.ConditionTrue,
+			LastTransitionTime: metav1.Now(),
+			Reason:             ConditionMaintenanceScheduled,
+		})
+		if err := patchHelper.Patch(ctx, capiMachine); err != nil {
+			errs = append(errs, fmt.Errorf("failed to patch Machine %s: %w", capiMachine.Name, err))
```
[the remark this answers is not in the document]
tchinmai7: added

PR #1091, opened 2026-06-26
[improvement] use VPC_NETWORK_CIDR if set in templates instead of assuming 10.0.0.0/8, update admission webhook check for VPC

AshleyDumaine:
This removes the "private IPs only" validating admission webhook check on VPC creation and fixes the templates to use the `VPC_NETWORK_CIDR` if set instead of assuming `10.0.0.0/8` (though that still is the default).

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 #####################################################################
 # top-level Makefile for cluster-api-provider-linode
 #####################################################################
+REGISTRY            ?= docker.io/linode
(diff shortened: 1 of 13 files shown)


komer3 (dismissed):

moshevayner (approved):
LGTM!

PR #1092, opened 2026-07-01
[deps] bump linodego to v2

AshleyDumaine:
linodego v1 is not receiving backports for features going into the new v2 (which has been out for a month).
See https://github.com/linode/linodego/blob/main/docs/linode_v2_migration_guide.md

I'm running the full e2e suite here given the scope of this change: https://github.com/linode/cluster-api-provider-linode/actions/runs/28807367677

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


AshleyDumaine:
Concerningly, I don't actually see any `POST`s being made for instances, VPCs, PGs, or FWs when the respective `client.Create` calls are being invoked 😬

api/v1alpha2/linodevpc_types.go
```diff
@@ -85,16 +85,11 @@ type LinodeVPCSpec struct {
...
+	Range *string `json:"range,omitzero"`
 
 	// allocationClass is the IPv6 inventory from which the VPC prefix should be allocated.
 	// +optional
-	AllocationClass *string `json:"allocationClass,omitempty"`
-
-	// allocation_class is the IPv6 inventory from which the VPC prefix should be allocated.
-	// +optional
-	//nolint:kubeapilinter // to maintain backwards compatibility
-	AllocationClassLegacy *string `json:"allocation_class,omitempty"`
```
komer3, line 97: Can we keep this legacy field? We will need to think a bit more about how we wanna migrate fields users don't have compatibility issues

internal/webhook/v1alpha2/linodevpc_webhook.go
```diff
@@ -126,7 +129,7 @@ func (r *linodeVPCValidator) validateLinodeVPCSpec(ctx context.Context, linodecl
 	var errs field.ErrorList
 
 	if !skipAPIValidation {
-		if err := validateRegion(ctx, linodeclient, spec.Region, field.NewPath("spec").Child("region"), linodego.CapabilityVPCs); err != nil {
+		if err := validateRegion(ctx, linodeclient, spec.Region, field.NewPath("spec").Child("region"), string(linodego.CapabilityVPCs)); err != nil {
```
komer3: So I see that we are doing a lot of string casting throughout all the changes. Do we want to take another look at it and just use the custom string types instead of doing these conversions?

api/v1alpha2/linodevpc_types.go
AshleyDumaine: updated, this is no longer a breaking change if we're leaving the CRDs alone 👍

komer3 (approved):
LGTM!

PR #1096, opened 2026-07-15
[improvement][fix] Check control plane component health of control plane machine to decide on readiness

AshleyDumaine:
This checks the (control plane) machine's control plane component conditions (apiserver, scheduler, controller-manager, etcd) for readiness similar to KCP via CAPI's `HasUnhealthyControlPlaneComponents` filter instead of relying on the Machine Ready condition or MachineHealthChecks:
https://github.com/kubernetes-sigs/cluster-api/blob/v1.13.4/controlplane/kubeadm/internal/control_plane.go#L333-L338

Fixes #

**TODOs**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/cloud/services/domains.go
+++ b/cloud/services/domains.go
@@ -15,9 +15,9 @@ import (
 	"github.com/akamai/AkamaiOPEN-edgegrid-golang/v12/pkg/dns"
 	"github.com/go-logr/logr"
 	"github.com/linode/linodego/v2"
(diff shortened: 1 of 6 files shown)


amold1 (dismissed):
lgtm - but i think we should test deadlocks more. may extending the new failure test such that we add a step that makes the etcd healthy and then check if our checks detect that and mark the cluster/node ready. just thinking about transitional errors and their possible routes to success

amold1 (approved):

PR #1127, opened 2026-09-02
[improvement][fix] Update firewall for Linode Interfaces on LinodeMachine firewall ID changes, only update instance firewall for instances using legacy interfaces

AshleyDumaine:
- If a `LinodeMachine` is configured to use Linode Interfaces and the `LinodeMachine.Spec.FirewallID` / ID of the `FirewallRef` changes, we need to be able to update the cloud firewall to allow the LinodeMachine's Linode Interfaces. 
Firewall ID updates on a `LinodeMachine` were already made possible in #813.

- There was also a bug in the instance firewall update logic where it would try to update the instance firewall regardless of interface generation resulting in this error:
```
2026-09-08T15:27:05Z	DEBUG	events	[400] Linode is using Linode Interfaces. Please apply Cloud Firewall to the interfaces of the Linode instead of directly to the Linode.	{"type": "Warning", "object": "nil", "action": "Reconcile", "reason": "UpdateError"}
```

- Fixes the events RBAC for the controller which uses a different API group:
```
E0908 15:26:16.368794       1 event_broadcaster.go:270] "Server rejected event (will not retry!)" err="events.events.k8s.io \"test-cluster-cp-92k2c.18d361f3893374d5\" is forbidden: User \"system:serviceaccount:capl-system:capl-controller-manager\" cannot patch resource \"events\" in API group \"events.k8s.io\" in the namespace \"default\"" event="&Event{ObjectMeta:{test-cluster-cp-92k2c.18d361f3893374d5  default    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] [] []},EventTime:2026-09-08 15:26:15.6030135 +0000 UTC m=+3117.259302542,Series:&EventSeries{Count:2,LastObservedTime:2026-09-08 15:26:16.367112459 +0000 UTC m=+3118.023401459,},ReportingController:LinodeMachineReconciler,ReportingInstance:LinodeMachineReconciler-capl-controller-manager-65f5cbc854-ndhcn,Action:Reconcile,Reason:UpdateError,Regarding:{LinodeMachine default test-cluster-cp-92k2c e89042ff-0c35-4856-86d5-b29077ef5e5c infrastructure.cluster.x-k8s.io/v1alpha2 20184 },Related:nil,Note:[400] Linode is using Linode Interfaces. Please apply Cloud Firewall to the interfaces of the Linode instead of directly to the Linode.,Type:Warning,DeprecatedSource:{ },DeprecatedFirstTimestamp:0001-01-01 00:00:00 +0000 UTC,DeprecatedLastTimestamp:0001-01-01 00:00:00 +0000 UTC,DeprecatedCount:0,}"
```

- This also cleans up a couple unused client methods on the firewalls client.

Fixes #

**TODOs**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/internal/controller/linodemachine_controller.go
+++ b/internal/controller/linodemachine_controller.go
@@ -839,40 +839,48 @@ func (r *LinodeMachineReconciler) reconcilePlacementGroup(ctx context.Context, l
 	return err
 }
 
-func (r *LinodeMachineReconciler) reconcileFirewallID(ctx context.Context, logger logr.Logger, machineScope *scope.MachineScope, instanceID int) (ctrl.Result, error) {
-	var (
-		firewalls  []linodego.Firewall
-		ifaceFWIDs []int
-		err        error
-	)
+// listAttachedFirewallIDs returns the IDs of the firewalls currently attached to the Linode instance,
+// using the appropriate API depending on whether the instance uses Linode Interfaces or not.
+func (r *LinodeMachineReconciler) listAttachedFirewallIDs(ctx context.Context, logger logr.Logger, machineScope *scope.MachineScope, instanceID int) ([]int, error) {
+	var firewalls []linodego.Firewall
+
 	// Get the instance's firewalls normally if this is not using the new linode interfaces,
 	// otherwise we have to get firewalls per linode interface
 	if machineScope.LinodeMachine.Spec.InterfaceGeneration == linodego.GenerationLinode {
 		interfaces, err := machineScope.LinodeClient.ListInterfaces(ctx, instanceID, nil)
 		if err != nil {
 			logger.Error(err, "Failed to list interfaces for Linode instance")
-			return ctrl.Result{RequeueAfter: reconciler.WithJitter(reconciler.DefaultMachineControllerWaitForRunningDelay)}, nil
+			return nil, err
 		}
 		for _, iface := range interfaces {
 			ifaceFWs, err := machineScope.LinodeClient.ListInterfaceFirewalls(ctx, instanceID, iface.ID, nil)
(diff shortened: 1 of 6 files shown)


internal/controller/linodemachine_controller.go
```diff
@@ -839,32 +838,41 @@ func (r *LinodeMachineReconciler) reconcilePlacementGroup(ctx context.Context, l
...
+	var firewalls []linodego.Firewall
+
 	// Get the instance's firewalls normally if this is not using the new linode interfaces,
 	// otherwise we have to get firewalls per linode interface
-	if machineScope.LinodeMachine.Spec.InterfaceGeneration == linodego.GenerationLinode {
-		interfaces, err := machineScope.LinodeClient.ListInterfaces(ctx, instanceID, nil)
+	switch machineScope.LinodeMachine.Spec.InterfaceGeneration {
+	case linodego.GenerationLinode:
+		linodeInterfaces, err := machineScope.LinodeClient.ListInterfaces(ctx, instanceID, nil)
 		if err != nil {
```
AshleyDumaine, line 851: I don't know why, but `ListInterfaces` has a zero-length result even when I can see in the `LINODE_DEBUG` that there should be at least a VPC interface for instances if using the new interface generation.

moshevayner (approved):
LGTM!

internal/controller/linodemachine_controller.go
AshleyDumaine: Turns out this was an issue with linodego 2.4.1, just pulling in main fixed it (2.6.0)

